PR #126, opened 2023-07-13
DPLT-1044 Expose context method for writing custom Grafana metrics

morgsmccauley:
This PR exposes a new context method (`context.putMetric()`) to allow writing of arbitrary metrics to CloudWatch Metrics, and eventually Grafana.

The main driver for this change is to expose the 'latest post block height lag' of the Feed Indexer. This requires comparing the latest block height in Postgres/GraphQL, to the latest block height from `api.social.near`. To do this we could use a scheduled Lambda, but it would be extremely tailored to this one Indexer. This PR provides the necessary tools to do this within the Feed Indexer itself.

After merging and adding the metrics writing to the Feed Indexer, I'll remove the existing scheduled Lambda which writes an incorrect block height lag.

### DIFF
--- a/indexer-js-queue-handler/__snapshots__/metrics.test.js.snap
+++ b/indexer-js-queue-handler/__snapshots__/metrics.test.js.snap
@@ -18,6 +18,10 @@ exports[`Metrics writes the block height for an indexer function 1`] = `
             "Name": "STAGE",
             "Value": "dev",
           },
+          {
+            "Name": "HISTORICAL",
+            "Value": false,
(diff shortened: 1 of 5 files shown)


roshaans (approved):
Really neat work! The current implementation ties us to Cloud Watch logs but that is probably fine since we are still using Lambda for now, but later we most likely want to abstract it out with an interface for a more general metrics collection library like Prometheus as you mentioned.

darunrs (approved):

PR #127, opened 2023-07-13
DPLT-1034 feat: allow to edit indexer when forking

roshaans:
While forking an indexer, you'll be given the chance to change the schema + indexerLogic before publishing. 

When forking,  we also update the indexer_name and account_name of the mutations and queries the new forked indexer targets.

-- What we do in this PR
Prior to this PR, the concept of forking an indexer(your own vs someone elses) or 
publishing an indexer(creating a new indexer vs publishing code to an existing indexer) were very confusing. 

With this PR, we separate the responsibilities to make it more clear. With that change, there is no concept of having an indexerConfig while forking an indexer as the only information we ask the user in the modal is about the new `indexerName` they would like to set. 

In the background, we use the `indexerConfig` of the indexer the user forked.

### DIFF
--- a/frontend/src/components/Editor/Editor.js
+++ b/frontend/src/components/Editor/Editor.js
@@ -34,6 +34,7 @@ const Editor = ({
     debugMode,
     isCreateNewIndexer,
     indexerNameField,
+    setAccountId,
   } = useContext(IndexerDetailsContext);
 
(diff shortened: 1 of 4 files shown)


darunrs (changes requested):
Just a couple clarification questions and nitpicks but it LGTM otherwise.

darunrs (approved):
LGTM

PR #148, opened 2023-07-25
DPLT-1066 feat: add fallback to old feed

roshaans:
If the lag between QueryAPI feed and NearSocial feed is more than 2 block heights, we fallback to the old feed widget

You can view the changes here: https://www.near.org/dev-queryapi.dataplatform.near/widget/QueryApi.dev.Feed

You can click `fork` and then mess with the QueryAPI link or the lag parameter to test the logic.


morgsmccauley (approved):
Since `social_feed` has the `social.near` contract filter, we'll essentially always have lag when there is a new post, since the indexer has not been triggered since the last post creating the lag. What this means is if we create a post, then immediately reload the feed, I think we'll fallback.

I don't think this is a problem, but perhaps something to keep in mind. What we'd really want is to check 'how long' we've been out of sync and fallback based on some time threshold. But this probably introduces too much complexity.

PR #174, opened 2023-08-09
fix: Fetch cargo packages from registry in Docker build

morgsmccauley:
When we had a multistage build, the second cargo build could use `--offline` as all packages would exist from the first build. With the first build removed we need to fetch the packages from the registry.


roshaans (approved):
Works for me now!

PR #176, opened 2023-08-12
feat: add replacement map feature

roshaans:
It was very confusing to have two applications `QueryApi.dev-App` vs `QueryApi.App`. As a result, we have removed the `dev-App` component and use a feature from [`bos-loader`](https://github.com/near/bos-loader) which allows you to write json files with keys/values that are used as replacements in your widgets. 

To run local setup, run `yarn start:widgets:local`

-- 
## Replacement Files
The replacement files should have the following format:
We have three different versions. For local, dev, and mainnet. 

{
  "REPL_PLACEHOLDER1": "value1",
  "REPL_PLACEHOLDER2": "value2"
}
The placeholders in widgets are replaced with specified values. For example the code for the following widget:

return <>
    <div> This is ${REPL_PLACEHOLDER1} </div>
    <Widget src="${REPL_ACCOUNT}/widget/SomeWidget">
    <div>${REPL_PLACEHOLDER2}</div>
</>;
will be resolved to:

return <>
    <div> This is value1 </div>
    <Widget src="accountId/widget/SomeWidget">
    <div>value2</div>
</>;


morgsmccauley (approved):
Why are all variables prefixed with `REPL_`? Seems unrelated

gabehamilton (approved):

darunrs (approved):
Oddly enough, if I run yarn dev locally, it puts the server on localhost: 9180 but I imagine that's a simple change in the local.json to update. Optionally, you can drop a line in the README to mention that.

PR #181, opened 2023-08-16
DPLT-1077 Process historical messages via Redis Streams

morgsmccauley:
This PR write historical messages to a dedicated Redis Stream per indexer. Redis keys have been slightly restructured to make it easier to distinguish which streams need to be monitored, mainly:
- Rather than storing the indexer name, i.e. `morgs.near/test`, we store the stream name, as this indexer could potentially have two streams, historical and real-time.
- `storage`, which stores the indexer config, is now suffixed under the stream name, so we can have separate storage for both historical and real-time, preventing race conflicts between them

The last point is a bit of a weird one, but I am trying to avoid the scenario where we change the historical indexers code out from underneath it. This would be the case if the developer changes their indexer code while a historical process is still running. Keeping the indexer code consistent throughout a given historical backfill seems most correct.

It would be possible for a developer to start another historical process, while an existing one is ongoing, causing the indexer code to change, as historical processes for a given indexer share the same storage key. But I don't think it's worth actively preventing this, as our ideal state is to stop the existing historical backfill when another is kicked off.

Once an historical process has been kicked off, Runner will _always_ start a process to listen to the stream, there's nothing telling it to stop listening. I may address this in a future PR, or just wait until we have control messages :).


roshaans (approved):
overall looks good to me.

PR #189, opened 2023-08-31
DPLT-1136: Implement update, upsert, and delete on context.db

darunrs:
~~NOTE: PR targeting my active PR https://github.com/near/queryapi/pull/188 for now to show accurate diff. Once that PR is merged, I'll switch this PR back to main (Or open a new one) and do a push to fix the commit history and reopen for review. This is just to be available to look at in the mean time.~~

**PR 188 has been merged and I have rebased and force pushed the history to this PR. There are no conflicts with main.**

Adding functionality for update, upsert, delete. Added relevant tests as well as performed simple integration tests using existing tests. No problems found there. Will do a full test in dev after push as well by updating a forked social_feed indexer to use context.db exclusively.

### DIFF
--- a/runner/src/dml-handler/dml-handler.test.ts
+++ b/runner/src/dml-handler/dml-handler.test.ts
@@ -35,7 +35,7 @@ describe('DML Handler tests', () => {
       accounts_liked: JSON.stringify(['cwpuzzles.near', 'devbose.near'])
     };
 
-    const dmlHandler = new DmlHandler(ACCOUNT, hasuraClient, PgClient);
+    const dmlHandler = await DmlHandler.create(ACCOUNT, hasuraClient, PgClient);
 
     await dmlHandler.insert(SCHEMA, TABLE_NAME, [inputObj]);
     expect(query.mock.calls).toEqual([
@@ -55,7 +55,7 @@ describe('DML Handler tests', () => {
       receipt_id: 'abc',
     }];
 
-    const dmlHandler = new DmlHandler(ACCOUNT, hasuraClient, PgClient);
+    const dmlHandler = await DmlHandler.create(ACCOUNT, hasuraClient, PgClient);
 
     await dmlHandler.insert(SCHEMA, TABLE_NAME, inputObj);
     expect(query.mock.calls).toEqual([
@@ -69,7 +69,7 @@ describe('DML Handler tests', () => {
       block_height: 999,
     };
 
-    const dmlHandler = new DmlHandler(ACCOUNT, hasuraClient, PgClient);
+    const dmlHandler = await DmlHandler.create(ACCOUNT, hasuraClient, PgClient);
 
     await dmlHandler.select(SCHEMA, TABLE_NAME, inputObj);
     expect(query.mock.calls).toEqual([
@@ -83,7 +83,7 @@ describe('DML Handler tests', () => {
       block_height: 999,
     };
 
-    const dmlHandler = new DmlHandler(ACCOUNT, hasuraClient, PgClient);
+    const dmlHandler = await DmlHandler.create(ACCOUNT, hasuraClient, PgClient);
 
     await dmlHandler.select(SCHEMA, TABLE_NAME, inputObj, 1);
     expect(query.mock.calls).toEqual([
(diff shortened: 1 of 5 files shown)


runner/src/dml-handler/dml-handler.ts
```diff
@@ -60,4 +60,56 @@ export default class DmlHandler {
...
+
+    const result = await wrapError(async () => await this.pgClient.query(this.pgClient.format(query), queryValues), `Failed to execute '${query}' on ${schemaName}.${tableName}.`);
+    if (!(result.rows && result.rows.length > 0)) {
+      console.log('No rows were selected.');
+    }
+    return result.rows;
+  }
+
+  async upsert (schemaName: string, tableName: string, objects: any[], conflictColumns: string[], updateColumns: string[]): Promise<any[]> {
+    await this.initialized; // Ensure constructor completed before proceeding
```
roshaans, line 83: Seems odd that we are awaiting a variable here. How about calling `await this.initialize()` and then inside of the `initialize()` method, you can change it to return early if it is already initialized.

You can also change the function signature of `initialize()` to return a Boolean rather than a void to make it more clearer.

runner/src/indexer/indexer.test.ts
```diff
@@ -507,38 +507,106 @@ CREATE TABLE
     expect(resultLimit.length).toEqual(1);
   });
```
roshaans, line 509: I think it also makes sense to cover the non-happy path scenarios such as testing to see if the `update, delete, upsert` methods fail as expected when garbage, malformed, or nonexistent table or field values are passed in to them.

roshaans (commented):
Good job, lgtm overall. Just left some comments on testing and conventions.

runner/src/indexer/indexer.test.ts
darunrs: So we actually do not support failure when invalid input is provided as we wouldn't know that unless we do one of two things: 
1. Run the query itself. 
2. Know what the input object structure should look like. 

1 is not helpful and 2, is a work in progress. We need to do the type generation on the schema to get table object types. We can use that to ensure good information is passed in. Unfortunately for now, we don't have that. The best I could do was use typescript strong typing to ensure the inputs sort of look normal.

runner/src/dml-handler/dml-handler.ts
darunrs: The reason I went with a variable is because I want the initialize function to be called in the constructor. However, I cannot await a function there. So, instead, I call it while setting the result to a variable. That way, I can call the function and not await in the constructor. Then, when a method is called, it just checks to see if the variable has been set. This was the most common pattern I could find for calling an async method in the constructor. It gives an error if I try to call the function as is. The return type of Promise<void> is because we don't actually care about the result of the function, just that it finished and set the variable to something.

morgsmccauley:
Can you please add the jira ticket to the PR title i.e. `DPLT-XXXX feat: Implement...`

runner/src/indexer/indexer.ts
```diff
@@ -258,6 +258,7 @@ export default class Indexer {
     try {
       const tables = this.getTableNames(schema);
       let dmlHandler: DmlHandler | null = null;
```
morgsmccauley, line 260: Why can this be `null`?

runner/src/dml-handler/dml-handler.ts
darunrs: That's neat! I'll give it a shot. Also updated the PR name.

runner/src/dml-handler/dml-handler.test.ts
```diff
@@ -90,4 +90,60 @@ describe('DML Handler tests', () => {
...
+
+    const dmlHandler = new DmlHandler(ACCOUNT, hasuraClient, PgClient);
+
+    await dmlHandler.update(SCHEMA, TABLE_NAME, whereObj, updateObj);
+    expect(query.mock.calls).toEqual([
+      ['UPDATE test_schema.test_table SET content=$1, receipt_id=$2 WHERE account_id=$3 AND block_height=$4 RETURNING *', [...Object.values(updateObj), ...Object.values(whereObj)]]
+    ]);
+  });
+
+  test('Test valid update on two fields', async () => {
```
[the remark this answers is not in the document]
darunrs: Fixed!

morgsmccauley (approved):
great job

PR #190, opened 2023-09-14
feat: Only write duration metric on success

morgsmccauley:
`EXECUTION_DURATION` will only be written when the indexer has been successfully executed, otherwise it is set to `-1`. This allows us to filter out these values in Grafana, and therefore only show accurate metrics.


morgsmccauley:
Because Grafana only scrapes every ~10 seconds or so, it's likely that the 'valid' values of these metrics will be missed, i.e. an indexer successfully executes, setting the metric duration, and then is almost instantly set back to `-1` if there is no other messages. Then when Grafana scrapes, it will only get `-1`, and not the actual duration prior.

PR #191, opened 2023-09-14
DPLT-1118 Parallelize stream processing with worker threads

morgsmccauley:
This PR parallelises indexer processing by moving them to their own threads. This means that CPU is no longer shared across all indexers on the single thread. I/O is still shared across all threads, but shouldn't be an issue as this isn't the limiting factor of Runner.


runner/src/metrics.ts
```diff
@@ -1,18 +1,23 @@
...
 });
 
-export const EXECUTION_DURATION = new promClient.Gauge({
+const EXECUTION_DURATION = new promClient.Gauge({
   name: 'queryapi_runner_execution_duration_milliseconds',
   help: 'Time taken to execute an indexer function',
   labelNames: ['indexer', 'type'],
 });
 
+export const METRICS = {
```
morgsmccauley, line 16: Exporting as a single group allows us to automatically create associated types, see: `Metric` in `StreamHandler`

runner/src/stream-handler/stream-handler.ts
```diff
@@ -0,0 +1,29 @@
+import path from 'path';
+import { Worker, isMainThread } from 'worker_threads';
+
+import { type Message } from './types';
+import { METRICS } from '../metrics';
+
+export default class StreamHandler {
```
morgsmccauley, line 7: This is a wrapper around `Worker` which will (eventually) expose a nicer API to the consumer, meaning they don't have to deal with sending/receiving messages.

runner/src/stream-handler/types.ts
```diff
@@ -0,0 +1,9 @@
+import { type METRICS } from '../metrics';
```
morgsmccauley, line 1: Shared types between the worker and api (`StreamHandler`)

runner/src/stream-handler/worker.ts
```diff
@@ -0,0 +1,75 @@
...
+      };
+      await indexer.runFunctions(Number(message.block_height), functions, false, {
+        provision: true,
+      });
+
+      await redisClient.deleteStreamMessage(streamKey, id);
+
+      const unprocessedMessages = await redisClient.getUnprocessedStreamMessages(streamKey);
+
+      parentPort?.postMessage({
```
morgsmccauley, line 57: Metrics written directly from another thread won't be picked up by the main thread, where the metrics server aggregates and exposes them. We therefore need to send the metric to the main thread and write them there.

PR #193, opened 2023-09-15
DPLT-1121: Generate and Support Strongly Typed Objects for DB Methods

darunrs:
This commit adds functionality for generating TypeScript interfaces and a context object which reflect rows in each table defined in the SQL table and respective functions for insert, select, upsert, update, and delete which all take in a strongly formatted object. This allows for autocomplete and type assertions for writing objects or calling functions related to each defined table. 

This is accomplished by using the library [node-sql-parser](https://www.npmjs.com/package/node-sql-parser) which contains functions for converting valid Postgres DDL into an AST or list of tables.

In the frontend, I have added a button which triggers code gen and attaches the types to Monaco. In addition, I set up listeners on monaco's mounting status as well as on the schemaTypes state variable which I introduced to store the generated types. These typesa re also saved to localStorage to be retrieved. The following actions trigger a code generation to take place: clicking the code gen button (Next to format button), switching to the indexingLogic.js tab, when monaco mounts, and when the schema and code are retrieved from local storage. The AST is parsed using the new PgSchemaTypeGen class. The end result is a context object formatted like so: context.db.TableName.methodName({}).

NOTE: The context object added to monaco is done so as a global variable. Currently our indexers pass in a context parameter. This causes the local parameter to shadow the global one, which prevents autocomplete or type checking. In order to use the global one, the context object must be removed from the function declaration for getBlock. 

In the backend, the context object received a similar refactor to the context.db.TableName.methodName format. It also uses the parser library but solely to get the table names to generate the context object.

### DIFF
--- a/frontend/src/components/Editor/Editor.js
+++ b/frontend/src/components/Editor/Editor.js
@@ -47,9 +47,6 @@ const Editor = ({
     }`;
   const CODE_STORAGE_KEY = `QueryAPI:Code:${indexerDetails.accountId}#${indexerDetails.indexerName || "new"
     }`;
-  
-  const pgSchemaTypeGen = useMemo(() => new PgSchemaTypeGen(), []);
-  const disposableRef = useRef(null);
 
   const [error, setError] = useState(undefined);
   const [blockHeightError, setBlockHeightError] = useState(undefined);
@@ -81,6 +78,8 @@ const Editor = ({
   };
 
   const indexerRunner = useMemo(() => new IndexerRunner(handleLog), []);
+  const pgSchemaTypeGen = useMemo(() => new PgSchemaTypeGen(), []);
+  const disposableRef = useRef(null);
 
   useEffect(() => {
     if (!indexerDetails.code || !indexerDetails.schema) return
@@ -88,7 +87,6 @@ const Editor = ({
     setOriginalSQLCode(formattedSchema)
     setOriginalIndexingCode(formattedCode)
     setIndexingCode(formattedCode)
-    console.log("Set original schema");
     setSchema(formattedSchema)
   }, [indexerDetails.code, indexerDetails.schema]);
 
@@ -97,7 +95,6 @@ const Editor = ({
     const savedSchema = localStorage.getItem(SCHEMA_STORAGE_KEY);
     const savedCode = localStorage.getItem(CODE_STORAGE_KEY);
 
-    console.log("Set cached schema", savedSchema);
     if (savedSchema) {
       setSchema(savedSchema);
       try {
@@ -117,7 +114,6 @@ const Editor = ({
   }, [schema, indexingCode]);
 
   useEffect(() => {
-    console.log("updated: ", schemaTypes);
     localStorage.setItem(SCHEMA_TYPES_STORAGE_KEY, schemaTypes);
     attachTypesToMonaco();
   }, [schemaTypes, monacoMount]);
@@ -129,7 +125,6 @@ const Editor = ({
 
   useEffect(() => {
     if (fileName === "indexingLogic.js") {
-      console.log("file is indexingLogic.js");
       try {
         setSchemaTypes(pgSchemaTypeGen.generateTypes(schema));
         setError(() => undefined);
@@ -144,7 +139,6 @@ const Editor = ({
   }, [heights]);
 
   const attachTypesToMonaco = () => {
-    console.log("prev disposable: ", disposableRef.current);
     // If types has been added already, dispose of them first
     if (disposableRef.current) {
       disposableRef.current.dispose();
@@ -155,7 +149,6 @@ const Editor = ({
       // Add generated types to monaco and store disposable to clear them later
       const newDisposable = monaco.languages.typescript.typescriptDefaults.addExtraLib(schemaTypes);
       disposableRef.current = newDisposable;
-      console.log(disposableRef.current);
     }
   }
 
@@ -222,9 +215,7 @@ const Editor = ({
     if (isCreateNewIndexer) {
       setShowResetCodeModel(false);
       setIndexingCode(originalIndexingCode);
-      console.log("New indexer schema");
       setSchema(originalSQLCode);
-      console.log("new indexer load");
       setSchemaTypes(defaultSchemaTypes);
       return;
     }
@@ -232,9 +223,7 @@ const Editor = ({
     const data = await queryIndexerFunctionDetails(indexerDetails.accountId, indexerDetails.indexerName);
     if (data == null) {
       setIndexingCode(defaultCode);
-      console.log("null data schema");
       setSchema(defaultSchema);
-      console.log("null data schema types");
       setSchemaTypes(defaultSchemaTypes);
       setError(() => onLoadErrorText);
     } else {
@@ -247,7 +236,6 @@ const Editor = ({
         }
         if (unformatted_schema !== null) {
           setOriginalSQLCode(unformatted_schema);
-          console.log("Unformatted schema");
           setSchema(unformatted_schema);
         }
         // if (data.start_block_height) {
@@ -257,7 +245,7 @@ const Editor = ({
         // if (data.filter) {
         //   setContractFilter(data.filter.matching_rule.affected_account_id)
         // }
-        await reformat(unformatted_wrapped_indexing_code, unformatted_schema) // Also sets schemaTypes
+        await reformat(unformatted_wrapped_indexing_code, unformatted_schema)
       } catch (error) {
         console.log(error);
       }
@@ -286,7 +274,6 @@ const Editor = ({
     setIndexingCode(formattedCode);
 
     const formattedSchema = formatSQL(schema);
-    console.log("Reformat all schema");
     setSchema(formattedSchema);
 
     return { formattedCode, formattedSchema }
@@ -301,7 +288,6 @@ const Editor = ({
           setIndexingCode(formattedCode);
         } else if (fileName === "schema.sql") {
           formattedCode = formatSQL(schema);
-          console.log("reformat schema");
           setSchema(formattedCode);
         }
         setError(() => undefined);
@@ -350,7 +336,6 @@ const Editor = ({
       `${primitives}}`,
       "file:///node_modules/@near-lake/primitives/index.d.ts"
     );
-    console.log("MOUNT");
     setMonacoMount(true);
   }
 
--- a/frontend/src/components/Editor/Editor.js
+++ b/frontend/src/components/Editor/Editor.js
@@ -78,7 +78,7 @@ const Editor = ({
   };
 
   const indexerRunner = useMemo(() => new IndexerRunner(handleLog), []);
-  const pgSchemaTypeGen = useMemo(() => new PgSchemaTypeGen(), []);
+  const pgSchemaTypeGen = new PgSchemaTypeGen();
   const disposableRef = useRef(null);
 
   useEffect(() => {
@@ -101,7 +101,7 @@ const Editor = ({
         setSchemaTypes(pgSchemaTypeGen.generateTypes(savedSchema));
         setError(() => undefined);
       } catch (error) {
-        handleCodeGenError();
+        handleCodeGenError(error);
       }
       
     } 
@@ -129,7 +129,7 @@ const Editor = ({
         setSchemaTypes(pgSchemaTypeGen.generateTypes(schema));
         setError(() => undefined);
       } catch (error) {
-        handleCodeGenError();
+        handleCodeGenError(error);
       }
     }
   }, [fileName]);
@@ -148,6 +148,9 @@ const Editor = ({
     if (window.monaco) { // Check if monaco is loaded
       // Add generated types to monaco and store disposable to clear them later
       const newDisposable = monaco.languages.typescript.typescriptDefaults.addExtraLib(schemaTypes);
+      if (newDisposable != null) {
+        console.log("Types successfully imported to Editor");
+      }
       disposableRef.current = newDisposable;
     }
   }
@@ -156,7 +159,13 @@ const Editor = ({
     try {
       let formatted_sql = formatSQL(schema);
       let formatted_schema = formatted_sql;
-      pgSchemaTypeGen.generateTypes(formatted_sql); // Sanity check
+      try {
+        pgSchemaTypeGen.generateTypes(formatted_sql); // Sanity check
+      } catch (error) {
+        handleCodeGenError(error);
+        return undefined;
+      }
+      
       return formatted_schema;
     } catch (error) {
       console.log("error", error);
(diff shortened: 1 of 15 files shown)


runner/src/indexer/indexer.ts
```diff
@@ -219,83 +222,87 @@ export default class Indexer {
...
+    if (/^[0-9]/.test(sanitizedName)) {
+      sanitizedName = '_' + sanitizedName;
+    } else {
+      sanitizedName = sanitizedName.charAt(0).toUpperCase() + sanitizedName.slice(1); // Convert to Pascal Case
+    }
+
+    // If table name already exists, append a number
+    const count = tableNameCount.get(sanitizedName) ?? 0;
+    if (count) {
+      console.warn(`Collision detected for table name ${tableName} while converting to Camel Case. Appending number.`);
```
morgsmccauley, line 253: Is this even possible? Wouldn't Postgres prevent duplicate table names?

runner/src/indexer/indexer.ts
```diff
@@ -219,83 +222,87 @@ export default class Indexer {
...
 
-  sanitizeTableName (tableName: string): string {
-    tableName = tableName.startsWith('"') && tableName.endsWith('"') ? tableName.substring(1, tableName.length - 1) : tableName;
-    return tableName.replace(/[^a-zA-Z0-9_]/g, '_');
+  sanitizeTableName (tableName: string, tableNameCount: Map<string, number>): string {
+    // Replace special characters with underscores
+    let sanitizedName: string = tableName.replace(/[^a-zA-Z0-9_]/g, '_');
+    // Convert to CamelCase
+    sanitizedName = sanitizedName
+      .replace(/_([a-zA-Z0-9])/g, (match: string) => match.toUpperCase())
```
morgsmccauley, line 241:
```suggestion
      .replace(/_([a-zA-Z])/g, (match: string) => match.toUpperCase())
```
We shouldn't need to match numbers here?

runner/src/indexer/indexer.ts
```diff
@@ -219,83 +222,87 @@ export default class Indexer {
...
   }
 
-  sanitizeTableName (tableName: string): string {
-    tableName = tableName.startsWith('"') && tableName.endsWith('"') ? tableName.substring(1, tableName.length - 1) : tableName;
-    return tableName.replace(/[^a-zA-Z0-9_]/g, '_');
+  sanitizeTableName (tableName: string, tableNameCount: Map<string, number>): string {
+    // Replace special characters with underscores
+    let sanitizedName: string = tableName.replace(/[^a-zA-Z0-9_]/g, '_');
+    // Convert to CamelCase
+    sanitizedName = sanitizedName
```
morgsmccauley, line 240:
```suggestion
    camelCaseTableName = sanitizedName
```
Nit: Rather than leaving comments, descriptive variable names can describe the intent

morgsmccauley: Same for below

runner/src/indexer/indexer.test.ts
```diff
@@ -616,16 +616,21 @@ CREATE TABLE
...
-      'insert_yet_another_table', 'select_yet_another_table', 'update_yet_another_table', 'upsert_yet_another_table', 'delete_yet_another_table']);
+      'CreatorQuest',
+      'ComposerQuest',
+      'ContractorQuest',
+      'Posts',
+      'Comments',
+      'PostLikes',
+      'AnotherTable',
+      'ThirdTable',
+      'YetAnotherTable']);
```
morgsmccauley, line 627: Feels like there is a lot of functionality in `sanitizeTableName()` that is not covered in these tests, e.g. numbers, duplicates, etc.

Can we please add tests for those?

runner/src/indexer/indexer.ts
```diff
@@ -219,83 +222,87 @@ export default class Indexer {
...
+      .replace(/_([a-zA-Z0-9])/g, (match: string) => match.toUpperCase())
+      .replace(/_/g, '');
+    // If starting with number, include a starting underscore. Otherwise, capitalize first character.
+    if (/^[0-9]/.test(sanitizedName)) {
+      sanitizedName = '_' + sanitizedName;
```
morgsmccauley, line 245: This seems like niche edge case, I'm not even sure it's worth covering?

I believe the rest of the code will still work with out this, and the property will still be accessible via:
```js
context.db['9myTable'].insert()
```

runner/src/indexer/indexer.ts
```diff
@@ -219,83 +222,87 @@ export default class Indexer {
...
-            dmlHandler = dmlHandler ?? await this.deps.DmlHandler.create(account);
-            return await dmlHandler.delete(schemaName, tableName, object);
+          [`${sanitizedTableName}`]: {
+            insert: async (objects: any) => {
+              await this.writeLog(`context.db.${sanitizedTableName}.insert`, blockHeight,
+                `Calling context.db.${sanitizedTableName}.insert.`,
+                `Inserting object ${JSON.stringify(objects)} into table ${tableName} on schema ${schemaName}`);
+              dmlHandler = dmlHandler ?? await this.deps.DmlHandler.create(account);
+              return await dmlHandler.insert(schemaName, tableName, Array.isArray(objects) ? objects : [objects]);
+            },
```
morgsmccauley, line 277:
```suggestion
            insert: async (objects: any) => {
              await this.writeLog(
                `context.db.${sanitizedTableName}.insert`,
                blockHeight,
                `Calling context.db.${sanitizedTableName}.insert.`,
                `Inserting object ${JSON.stringify(objects)} into table ${tableName} on schema ${schemaName}`
              );

              dmlHandler = dmlHandler ?? await this.deps.DmlHandler.create(account);
              return await dmlHandler.insert(schemaName, tableName, Array.isArray(objects) ? objects : [objects]);
            },
```
Nit: Can we please add some formatting here and below, it's quite dense and hard to read 😅

runner/src/dml-handler/dml-handler.ts
```diff
@@ -24,6 +26,10 @@ export default class DmlHandler {
     return new DmlHandler(pgClient);
   }
 
+  private quoteTableName (tableName: string): string {
+    return !this.validTableNameRegex.test(tableName) ? `"${tableName}"` : tableName;
```
morgsmccauley, line 30: Why don't you just quote every table and skip this check? seems simpler?

morgsmccauley (changes requested):

frontend/src/components/Editor/Editor.js
```diff
@@ -72,6 +78,9 @@ const Editor = ({
   };
 
   const indexerRunner = useMemo(() => new IndexerRunner(handleLog), []);
+  const pgSchemaTypeGen = useMemo(() => new PgSchemaTypeGen(), []);
```
morgsmccauley, line 81: creating this value isn't expensive, we shouldn't need `useMemo` here

frontend/src/components/Editor/Editor.js
```diff
@@ -94,25 +113,50 @@ const Editor = ({
...
+  }, [fileName]);
 
   useEffect(() => {
     localStorage.setItem(DEBUG_LIST_STORAGE_KEY, heights);
   }, [heights]);
 
+  const attachTypesToMonaco = () => {
+    // If types has been added already, dispose of them first
+    if (disposableRef.current) {
+      disposableRef.current.dispose();
```
morgsmccauley, line 144: Is `.dispose()` defined on `monaco` or `useRef()`? If the former, I think this would be better named `schemaTypesMonacoRef` rather than `disposableRef`?

runner/src/indexer/indexer.ts
darunrs: Well there is in fact one possible scenario where a duplicate could occur that we can't handle. The library I use omits quotes when getting the table names. As it turns out, in postgres, putting quotes is only necessary if the name contains special characters. What it also does is state to postgres that this table name is case sensitive. So, if a customer, for example, had a schema with "CREATE TABLE tableName" as well as "CREATE TABLE "tableName", that would actually be valid. That's because the table name without quotes can be referenced with tAbLenAmE just the same as TABLENAME since its case insensiitive. Whereas the one with quotes can ONLY be referenced with "tableName". Its a pretty fringe case so realistically I could just the same skip if there's a duplicate. I actually realize now that my current handling wouldn't work anyway since when I implement it in runner, it wouldn't add quotes (I add quotes only if special characters are present) to the SQL statement. So, it's not possible with current implementation to target it. I would need to determine if the tableName I am currently targeting had quotes originally. I'll have to go in and fix that.

runner/src/indexer/indexer.ts
darunrs: True. We don't need to. I'll remove it!

runner/src/indexer/indexer.ts
darunrs: Makes sense. I was trying to reuse the variable, but creating more but being more declarative is worth it.

runner/src/dml-handler/dml-handler.ts
darunrs: I think I got caught in the weeds of trying to chase down bugs. This is indeed an elegant way to fix the majority of the problems. What it wouldn't be able to cover is the edge case I mentioned above with tableName vs "tableName". I think what I'll do is just enforce that. If there's a table name collision, instead of allow it, I'll error out and not build DB methods unless that's fixed.

runner/src/indexer/indexer.ts
darunrs: I'll actually reject upon table name collision and go with quoting all table names, as you suggested below.

runner/src/indexer/indexer.test.ts
darunrs: Will do!

runner/src/indexer/indexer.ts
darunrs: Ah my impression was that it isn't possible to access it at all (leads to an error). I just tested and it seems to work just fine. I'll validate a bit more and remove that edge case handling if it all works out.

runner/src/indexer/indexer.ts
darunrs: Haha sorry. Will do!

frontend/src/components/Editor/Editor.js
darunrs: It's the former. I can change the name, sure.

runner/src/indexer/indexer.ts
darunrs: So I am able to access the method. However, the types generated for the table (like the input or item object) have invalid names since they start with a number. So, either I can keep that edge case coverage or simply error out. Either way, I would need to handle it since the 'error' is invisible otherwise.

runner/src/indexer/indexer.ts
darunrs: Oh I also forgot. Due to sanitizing, we increase the chance of collisions since we are removing information. All special characters are converted to underscore. So "test table" and "test!table" both become TestTable, which is a problem. I think in general, no one would encounter such an edge case and if they do, they're just trying to mess with us honestly. I can't imagine anyone *asking* for supporting that. Anyway, I'll update the code a bit to catch collisions in sanitized table names too.

runner/src/indexer/indexer.ts
pkudinov: Let's enforce it in eslint? Not in this PR of course.

frontend/src/utils/indexerRunner.js
```diff
@@ -153,57 +154,35 @@ export default class IndexerRunner {
...
-          [`select_${sanitizedTableName}`]: async (object, limit = 0) => await this.dbOperationLog(blockHeight,
-            `Selecting objects with values ${JSON.stringify(object)} from table ${tableName} on schema ${schemaName} with ${limit === 0 ? 'no' : roundedLimit.toString()} limit`),
-          [`update_${sanitizedTableName}`]: async (whereObj, updateObj) => await this.dbOperationLog(blockHeight,
-            `Updating objects that match ${JSON.stringify(whereObj)} with values ${JSON.stringify(updateObj)} in table ${tableName} on schema ${schemaName}`),
-          [`upsert_${sanitizedTableName}`]: async (objects, conflictColumns, updateColumns) => await this.dbOperationLog(blockHeight,
-            `Inserting objects with values ${JSON.stringify(objects)} in table ${tableName} on schema ${schemaName}. Conflict on columns ${conflictColumns.join(', ')} will update values in columns ${updateColumns.join(', ')}`),
-          [`delete_${sanitizedTableName}`]: async (object) => await this.dbOperationLog(blockHeight,
-            `Deleting objects with values ${JSON.stringify(object)} in table ${tableName} on schema ${schemaName}`)
+          [`${sanitizedTableName}`]: {
```
roshaans, line 174: At the moment it is hard to tell which logs were generated by the system and by the user. 

you can use console.group() to highlight the system logs and clean things up a bit like this:
image

I changed `this.dbOperationLog()` to be the following:
```
dbOperationLog(blockHeight, logData, objects) {
    let operationToLogMap = {
      insert: `Inserting objects with values on ${logData.tableName} on Schema ${logData.schema}`,
      // rest of the operations
    };

    this.handleLog(blockHeight, "", () => {
      console.group(
        `${operationToLogMap[logData.operation]}`
      );
      console.log(objects);
      console.groupEnd();
    });
    return {};
  }
```

and this is how I am calling it from `buildDatabaseContext()`

```
  buildDatabaseContext(blockHeight, schemaName, schema) {
    
       //  Beginning of function

        let logData = {
          tableName: sanitizedTableName,
          schemaName,
        };

        // Generate context.db methods for table
        const funcForTable = {
          [`${sanitizedTableName}`]: {
            insert: async (objects) =>
              await this.dbOperationLog(
                blockHeight,
                { operation: "insert", ...logData },
                objects
              ),
// Rest of the function
}
```

roshaans (changes requested):
Great Job @darunrs! This will be definitely make the UX much better!

Just some changes on how the logs are viewed from the console window.

frontend/src/utils/indexerRunner.js
darunrs: Oh that's awesome! Thanks for the tip! I'll implement it.

morgsmccauley (approved):

darunrs:
https://github.com/near/queryapi/pull/193

frontend/src/utils/indexerRunner.js
```diff
@@ -217,12 +204,15 @@ export default class IndexerRunner {
     }
   }
 
-  dbOperationLog(blockHeight, logMessage) {
+  dbOperationLog(blockHeight, logMessage, data) {
+    console.log(logMessage);
```
roshaans, line 208: shoudl this be here twice?

roshaans (approved):
LGTM. Just a small comment on logs that I think are redundant

frontend/src/utils/indexerRunner.js
darunrs: What's there twice?

roshaans: if you see, you do console.log on line 208 and then you do it again later in the function.

darunrs: Ahhh I got you. Yeah that doesn't need to be there twice. Good catch!

PR #195, opened 2023-09-20
fix: Skip missing blocks in manual filtering

### LINKED ISSUE #194: [bug] Missing block numbers cause historical processing to fail
When historical processing does manual filtering of the past hour or two since the last index run, missing block numbers cause it to fail.

gabehamilton:
Proceed with historical filtering when missing block numbers are encountered.

### DIFF
--- a/indexer/queryapi_coordinator/src/historical_block_processing.rs
+++ b/indexer/queryapi_coordinator/src/historical_block_processing.rs
@@ -330,18 +330,32 @@ async fn filter_matching_unindexed_blocks_from_lake(
     for current_block in (last_indexed_block + 1)..ending_block_height {
         // fetch block file from S3
         let key = format!("{}/block.json", normalize_block_height(current_block));
-        let block = s3::fetch_text_file_from_s3(&lake_bucket, key, s3_client.clone()).await;
-        if block.is_err() {
-            tracing::error!(
-                target: crate::INDEXER,
-                "Skipping block in manual filtering, Error fetching block {} from S3 for function {:?} {:?}",
-                current_block,
-                indexer_function.account_id,
-                indexer_function.function_name,
-            );
-            continue;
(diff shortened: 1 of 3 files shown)


indexer/queryapi_coordinator/src/historical_block_processing.rs
```diff
@@ -330,7 +330,18 @@ async fn filter_matching_unindexed_blocks_from_lake(
     for current_block in (last_indexed_block + 1)..ending_block_height {
         // fetch block file from S3
         let key = format!("{}/block.json", normalize_block_height(current_block));
-        let block = s3::fetch_text_file_from_s3(&lake_bucket, key, s3_client.clone()).await?;
+        let block = s3::fetch_text_file_from_s3(&lake_bucket, key, s3_client.clone()).await;
+        if block.is_err() {
```
morgsmccauley, line 334: What if it fails for reasons other than 'missing'? should we inspect the error message/status?

indexer/queryapi_coordinator/src/historical_block_processing.rs
```diff
@@ -330,7 +330,18 @@ async fn filter_matching_unindexed_blocks_from_lake(
...
+            tracing::error!(
```
darunrs, line 335: We may want to include the error itself in the trace or at least log it to the machine itself.

darunrs (approved):
LGTM. One comment on error logging.

indexer/queryapi_coordinator/src/historical_block_processing.rs
gabehamilton: Fair enough. New commit added with more specific error handling.

indexer/queryapi_coordinator/src/historical_block_processing.rs
gabehamilton: Just changed it to only catch the 404

morgsmccauley (approved):
Nice!

PR #237, opened 2023-09-26
feat: put toggle + disclaimer

### LINKED ISSUE #234: Create a V2/V1 toggle in QueryAPI.App

roshaans:
Adds a toggle to update which graphql endpoint the logs and state are pulled from. 

 
gipp

### DIFF
--- a/frontend/replacement.dev.json
+++ b/frontend/replacement.dev.json
@@ -1,6 +1,7 @@
 {
     "REPL_ACCOUNT_ID": "dev-queryapi.dataplatform.near",
     "REPL_GRAPHQL_ENDPOINT": "https://near-queryapi.dev.api.pagoda.co",
+    "REPL_GRAPHQL_ENDPOINT_V2": "https://queryapi-hasura-graphql-mainnet-vcqilefdcq-ew.a.run.app",
     "REPL_EXTERNAL_APP_URL": "https://queryapi-frontend-vcqilefdcq-ew.a.run.app",
     "REPL_REGISTRY_CONTRACT_ID": "dev-queryapi.dataplatform.near"
 }
--- a/frontend/widgets/src/QueryApi.IndexerStatus.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerStatus.jsx
@@ -97,7 +97,6 @@ if (!indexer_name) return "missing indexer_name";
 
 let v1_endpoint = `${REPL_GRAPHQL_ENDPOINT}`;
 let v2_endpoint = `${REPL_GRAPHQL_ENDPOINT_V2}`;
-let graphQLEndpoint = state.v2Toggle ? v2_endpoint : v1_endpoint;
 
 State.init({
   logs: [],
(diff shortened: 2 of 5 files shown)


frontend/widgets/src/QueryApi.IndexerStatus.jsx
```diff
@@ -244,6 +291,39 @@ return (
...
+            justifyContent: "center",
+          }}
+        >
+          <DisclaimerContainer>
+            <div className="flex">
+              <Notice>V2 Testing Notice</Notice>
+              <div style={{ display: "flex" }}>
+                <DisclaimerText>
+                  QueryAPI is still in beta. We are working on a OueryAPI V2
+                  with faster historical processing, easier access to DB and and
```
darunrs, line 308: nit: two "and"s next to each other.

frontend/replacement.dev.json
```diff
@@ -1,6 +1,7 @@
 {
     "REPL_ACCOUNT_ID": "dev-queryapi.dataplatform.near",
     "REPL_GRAPHQL_ENDPOINT": "https://near-queryapi.dev.api.pagoda.co",
+    "REPL_GRAPHQL_ENDPOINT_V2": "https://queryapi-hasura-graphql-mainnet-vcqilefdcq-ew.a.run.app",
```
darunrs, line 4: I see the V1 and V2 endpoint don't look similar. Is the V1 endpoint a domain we registered or something?

frontend/widgets/src/QueryApi.IndexerStatus.jsx
```diff
@@ -105,11 +109,15 @@ State.init({
   indexer_resPage: 0,
   logsPage: 0,
   statePage: 0,
+  v2Toggle: false,
```
darunrs, line 112: Could we pull the v2Toggle from storage to retain the user's previous selection, and default to false if none? If someone has to refresh their page to view logs for example, and it switches to off each time, that could be frustrating. 

Also, does the v2Toggle remain when switching indexers or is it switched back to off when you switch indexers?

frontend/widgets/src/components/toggle.jsx
```diff
@@ -0,0 +1,78 @@
...
+    {...rest}
+  >
+    <ToggleLabel htmlFor={`toggle-${key}`}>{label}</ToggleLabel>
+
+    <ToggleSwitchRoot
+      checked={active}
+      className="shadow-none"
+      id={`toggle-${key}`}
+      onCheckedChange={disabled ? null : onSwitch}
+      title={disabled ? `Permanently ${active ? "enabled" : "disabled"}` : null}
```
darunrs, line 70: What does this line do? What is supposed to be "permanently enabled/disabled"?

roshaans: Not quite sure actually. I copied this component from devgogs.near

frontend/widgets/src/QueryApi.IndexerStatus.jsx
roshaans: Good point. I tried to implement it but it was a little finicky.

Issue #204, opened 2023-09-26
Optimize Runner Streamer Message Acquisition


Ishatt:
Description

Currently the Indexer Runner fetches and constructs the Streamer Message from S3. Instead, we can store real-time messages in Redis to avoid the; amount of duplicate requests, and the latency cost. The message can have a short TTL to avoid consuming too much memory, and we can fallback on fetching the message if it doesn’t exist. Processing of the streamer message such as the call to renameUnderscoreFieldsToCamelCase should be done before caching to further reduce latency. 

Publish cache hits/misses to Prometheus so we can fine tune TTL.

Upgrade AWS SDK from v2, which is to be deprecated.

Pre-Fetch Historical Blocks from S3 during historical processing. 
```[tasklist]
### Tasks
- [ ] https://github.com/near/queryapi/issues/262
- [ ] https://github.com/near/queryapi/issues/264
```

darunrs:
PR: https://github.com/near/queryapi/pull/241

PR #241, opened 2023-09-27
Store Real Time Streamer Messages in Redis

### LINKED ISSUE #262: Cache Real Time Streamer Message in Redis
Migrate AWS SDK from v2 to v3.
Cache streamer message in Redis in Coordinator. 
Pull cached message from Redis in Runner if not historical. 
Migrate camel case processing to coordinator. 
Log metrics for cache hits and misses.

darunrs:
The streamer message is used by both the coordinator and runner. However, both currently poll the message from S3. There is a huge latency impact for pulling the message from S3. In order to improve this, the streamer message will now be cached in Redis with a TTL and pulled by runner from Redis. Only if there is a cache miss will runner pull from S3 again. 

Pulling from S3 currently takes up 200-500ms, which is roughly 80-85% of the overall execution time of a function in runner. By caching the message, a cache hit leads to loading the data in 1-3ms in local testing, which corresponds to about 3-5% of the execution time, or a 1100% improvement in latency. The reduction of network related activity to such a low percentage of execution time also reduces the variability of a function's execution time greatly. Cache hits and misses will be logged for further tuning of TTL to reduce cache misses. In addition, processing the block takes around 1-3ms. This processing can be moved to be done before caching, saving 1-3ms each time that block is read from cache, which does stack up over time. The improvement there will be more important for historical backfill, which is planned to be cached soon.

 Tracking Issue: https://github.com/near/queryapi/issues/262
Parent Issue: https://github.com/near/queryapi/issues/204


darunrs:
https://github.com/near/queryapi/issues/204

runner/src/indexer/indexer.ts
```diff
@@ -479,7 +492,7 @@ export default class Indexer {
   }
 
   renameUnderscoreFieldsToCamelCase (value: Record<string, any>): Record<string, any> {
-    if (typeof value === 'object' && !Array.isArray(value)) {
+    if (value !== null && typeof value === 'object' && !Array.isArray(value)) {
```
darunrs, line 495: This if loop was allowing null values through, which was leading to them being replaced by {}. Not sure if this is problematic or a non-issue. In any case, I updated the if statement to retain null values and was able to confirm my implementation does match what S3 returns.

gabehamilton: Nice. This code was copied from https://github.com/near/near-lake-framework-js/blob/eb35cf836c03c9cd925d4b7c6dcb29b52aa757fa/src/utils.ts#L35 
Looks like it got the same update at some point.

gabehamilton (commented):
👍

runner/src/metrics.ts
```diff
@@ -13,9 +13,23 @@ const EXECUTION_DURATION = new promClient.Gauge({
...
+  name: 'redis_cache_hit_for_streamer_message',
+  help: 'The number of times the streamer message cache was hit',
+  labelNames: ['type']
+});
+
+const CACHE_MISS_STREAMER_MESSAGE = new promClient.Counter({
+  name: 'redis_cache_miss_for_streamer_message',
+  help: 'The number of times the streamer message cache was missed',
+  labelNames: ['type']
+});
```
morgsmccauley, line 26:
```suggestion
const CACHE_HIT = new promClient.Counter({
  name: 'queryapi_runner_cache_hit',
  help: 'Number of times cache hit was successful',
  labelNames: ['type', 'key']
});
```
It might make sense to make this more generic, and add a new dimension/label for the key which was hit/missed. That way we use this more globally.

Same for above

runner/src/indexer/indexer.ts
```diff
@@ -132,7 +136,17 @@ export default class Indexer {
     return blockHeight.toString().padStart(12, '0');
   }
 
-  async fetchStreamerMessage (blockHeight: number): Promise<{ block: any, shards: any[] }> {
+  async fetchStreamerMessage (blockHeight: number, isHistorical: boolean): Promise<{ block: any, shards: any[] }> {
+    if (!isHistorical) {
+      const cachedMessage = await this.deps.redisClient.getStreamerMessageFromCache(`near-lake-data-${this.network}`, blockHeight);
+      if (cachedMessage) { // Cache hit on streamer message
+        METRICS.CACHE_HIT_STREAMER_MESSAGE.labels(isHistorical ? 'historical' : 'realtime').inc(); // increment the cache hit counter
```
morgsmccauley, line 143:
```suggestion
        METRICS.CACHE_HIT_STREAMER_MESSAGE.labels(isHistorical ? 'historical' : 'real-time').inc(); // increment the cache hit counter
```
We use this for other metrics, let's stay consistent with that.

runner/package.json
```diff
@@ -42,8 +42,8 @@
     "typescript": "^5.1.6"
   },
   "dependencies": {
+    "@aws-sdk/client-s3": "^3.414.0",
```
morgsmccauley, line 45: nice

runner/src/indexer/indexer.ts
```diff
@@ -132,7 +136,17 @@ export default class Indexer {
     return blockHeight.toString().padStart(12, '0');
   }
 
-  async fetchStreamerMessage (blockHeight: number): Promise<{ block: any, shards: any[] }> {
+  async fetchStreamerMessage (blockHeight: number, isHistorical: boolean): Promise<{ block: any, shards: any[] }> {
+    if (!isHistorical) {
+      const cachedMessage = await this.deps.redisClient.getStreamerMessageFromCache(`near-lake-data-${this.network}`, blockHeight);
+      if (cachedMessage) { // Cache hit on streamer message
```
morgsmccauley, line 142: Nit: It's best to add comments when the intent isn't super obvious, here and below I think the code speaks for itself, I think the comments are unnecessary :)

morgsmccauley (changes requested):
Great first start :) I've left a few comments. Can you please also link the issue?

darunrs:
Tracking Issue: https://github.com/near/queryapi/issues/262
Parent Issue: https://github.com/near/queryapi/issues/204

indexer/storage/src/lib.rs
```diff
@@ -57,6 +59,22 @@ pub async fn set(
     Ok(())
 }
 
+pub async fn setEx(
+    redis_connection_manager: &ConnectionManager,
+    key: impl ToRedisArgs + std::fmt::Debug,
+    expiration: usize,
+    value: impl ToRedisArgs + std::fmt::Debug,
+) -> anyhow::Result<()> {
+    redis::cmd("SETEX")
```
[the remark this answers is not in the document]
darunrs: Yep that works. I've implemented it under set(). I went with using an optional.

indexer/queryapi_coordinator/src/main.rs
```diff
@@ -146,6 +147,21 @@ async fn handle_streamer_message(
...
     let block_height: BlockHeight = context.streamer_message.block.header.height;
 
+    // Cache streamer message block and shards for use in real time processing
+    storage::setEx(
+        context.redis_connection_manager,
+        format!(
+            "{}{}{}:{}",
+            storage::STREAMER_MESSAGE_HASH_KEY_BASE,
+            storage::LAKE_BUCKET_PREFIX,
+            context.chain_id,
```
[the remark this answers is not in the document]
darunrs: Ah gotcha. In that case, I'll remove it.

indexer/queryapi_coordinator/src/utils.rs
```diff
@@ -49,3 +51,51 @@ pub(crate) async fn stats(redis_connection_manager: storage::ConnectionManager)
...
 }
+
+pub(crate) fn process_streamer_message(
+    streamer_message: &near_lake_framework::near_indexer_primitives::StreamerMessage,
+) -> Value {
+    // Serialize the Message object to a JSON string
+    let json_str = serde_json::to_string(&streamer_message).unwrap();
+
+    // Deserialize the JSON string to a Value Object
+    let mut message_value: Value = serde_json::from_str(&json_str).unwrap();
```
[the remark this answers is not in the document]
darunrs: Thanks for the info! I've updated it to not use unwrap.

indexer/queryapi_coordinator/src/utils.rs
```diff
@@ -49,3 +51,51 @@ pub(crate) async fn stats(redis_connection_manager: storage::ConnectionManager)
         tokio::time::sleep(std::time::Duration::from_secs(interval_secs)).await;
     }
 }
+
+pub(crate) fn process_streamer_message(
+    streamer_message: &near_lake_framework::near_indexer_primitives::StreamerMessage,
+) -> Value {
```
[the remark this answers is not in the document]
darunrs: That's indeed a better name. And gotcha, I'll have it return serialized to string.

indexer/queryapi_coordinator/src/main.rs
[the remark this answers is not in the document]
darunrs: I'll have to remove it from Runner too. I'm guessing we'll have a different runner instance which is handling requests?

runner/src/metrics.ts
darunrs: That's a good idea. I'll implement it.

runner/src/indexer/indexer.ts
darunrs: True! I'll remove it.

indexer/queryapi_coordinator/src/main.rs
```diff
@@ -146,6 +147,19 @@ async fn handle_streamer_message(
...
     let block_height: BlockHeight = context.streamer_message.block.header.height;
 
+    // Cache streamer message block and shards for use in real time processing
+    storage::set(
+        context.redis_connection_manager,
+        format!(
+            "{}{}",
+            storage::STREAMER_MESSAGE_HASH_KEY_BASE,
+            block_height
+        ),
```
morgsmccauley, line 157: Nit: Might be good to extract this to a method like we have for the others in `storage/lib.rs`
```rust
pub fn generate_historical_stream_key(prefix: &str) -> String {
    format!("{}:historical:stream", prefix)
}
```

morgsmccauley (commented):
Couple small comments but otherwise looks good!

indexer/storage/src/lib.rs
```diff
@@ -47,13 +49,23 @@ pub async fn set(
...
+        cmd.arg("EX").arg(expiration);
+        exp_to_print = format!("EX {}", expiration);
+    } else {
+        exp_to_print = "".to_string();
+    }
+
+    cmd.query_async(&mut redis_connection_manager.clone())
         .await?;
-    tracing::debug!(target: STORAGE, "SET: {:?}: {:?}", key, value,);
+    tracing::debug!(target: STORAGE, "SET: {:?}: {:?} {:?}", key, value, exp_to_print);
```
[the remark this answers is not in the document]
darunrs: Good idea! Will do.

morgsmccauley (approved):
LGTM!

PR #269, opened 2023-10-05
feat: Pre-Fetch Streamer Messages

### LINKED ISSUE #264: Pre-Fetch Streamer Messages
Historical streamer messages are not fetched in coordinator prior to the IndexerRunner call. As a result, we cannot apply the latency saving benefits of having coordinator cache the streamer message for use by runner. Instead, we want to pre fetch from S3 so that runner invocations don't have to wait for the streamer message to be retrieved from S3. Instead, these messages will be pre fetched and awaited in an array to ensure in order processing of block heights. 

Below is the current flow for historical processing: 
1. Coordinator retrieves timestamp for the block height that historical processing is starting from as well as current block height. It uses this to see which days to look for index files. 
2. Index files are fils generated for each day an indexer is active. They contain information such as block heights which the particular indexer function was applied for. Coordinator fetches each index file available for the indexer starting from the day the starting block height falls into. 
3. Coordinator parses the block heights from the file and puts them in the historical redis stream. **This is where the divergence between real time and historical lies. Real time does not have any index files so it reads the streamer message from S3 to get data including block height. This block height is put into the real time stream.**
4. Runner reads the block height from the historical stream. It pulls the streamer message from S3, parses it, and uses it for execution. This leads to each invocation taking at least 200ms if not more. I've seen as high as 700ms in a sample size of 20 invocations. 99th percentile might be much higher. 

Below is the new workflow:
1. Coordinator functionality remains the same. 
2. In runner, fetch X blocks from S3 as a promise. 
3. Load the promises into an array, which is used as a queue. 
4. Delete the block height from the stream, for each block height successfully placed on queue. 
5. Await the first block in the queue. Upon completion of the promise, trigger the function call and pass in the loaded data. 

I've also made it so that real-time also uses prefetch mechanism on top of the existing caching.

While an indexer function is running, several other blocks are being loaded simultaneously. For each loop, we ensure the array is as full as possible. This ensures few functions are waiting for the block instead of all of them. 
```[tasklist]
### Tasks
- [x] Migrate Shared S3 Code to New Class
- [x] Add console statements to breakdown execution time
- [x] Tune Queue size parameter
- [x] Get accurate estimates of overhead latency
- [x] Create metrics on overhead, database, code execution, and so on
- [x] Reduce latency as much as possible, targeting 100+ BPS
- [x] Update BPS to also include skipped blocks
```

darunrs:
Historical streamer messages are not fetched in coordinator prior to the IndexerRunner call. As a result, we cannot apply the latency saving benefits of having coordinator cache the streamer message for use by runner. Instead, we want to pre fetch from S3 so that runner invocations don't have to wait for the streamer message to be retrieved from S3. 

In addition, it's possible for real-time messages to backup temporarily preventing the cached message from being used. So, we also want to prefetch any messages which aren't found in the cache.

The new workflow works by having two loops for each worker thread: a producer and a consumer. The producer loads a promise for fetching the block (either from cache or S3) into an array. The consumer then removes the first element from the array and processes it, deleting the streamer message upon success. While one block is being processed, the other blocks are being fetched. This ensures that wait time is minimal. The producer loop attempts to keep the array as close to full as possible.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,93 +1,93 @@
 version: "3.9"  # optional since v1.27.0
 services:
 
-  # coordinator:
-  #   build:
-  #     context: ./indexer
-  #     args:
-  #       - CARGO_BUILD_MODE=debug
-  #   depends_on:
-  #     - redis
-  #   environment:
-  #     REDIS_CONNECTION_STRING: redis://redis
-  #     LAKE_AWS_ACCESS_KEY:
-  #     LAKE_AWS_SECRET_ACCESS_KEY:
-  #     QUEUE_AWS_ACCESS_KEY:
-  #     QUEUE_AWS_SECRET_ACCESS_KEY:
-  #     QUEUE_URL: MOCK
-  #     START_FROM_BLOCK_QUEUE_URL: MOCK
-  #     PORT: 9180
-  #     REGISTRY_CONTRACT_ID: dev-queryapi.dataplatform.near
-  #     AWS_QUEUE_REGION: eu-central-1
-  #   command:
-  #     - mainnet
-  #     - from-interruption
+  coordinator:
+    build:
+      context: ./indexer
+      args:
+        - CARGO_BUILD_MODE=debug
+    depends_on:
+      - redis
+    environment:
+      REDIS_CONNECTION_STRING: redis://redis
+      LAKE_AWS_ACCESS_KEY:
+      LAKE_AWS_SECRET_ACCESS_KEY:
+      QUEUE_AWS_ACCESS_KEY:
+      QUEUE_AWS_SECRET_ACCESS_KEY:
+      QUEUE_URL: MOCK
+      START_FROM_BLOCK_QUEUE_URL: MOCK
+      PORT: 9180
+      REGISTRY_CONTRACT_ID: dev-queryapi.dataplatform.near
+      AWS_QUEUE_REGION: eu-central-1
+    command:
+      - mainnet
+      - from-interruption
 
-  # runner:
-  #   build:
-  #     context: ./runner
-  #   depends_on:
-  #   - "hasura-graphql"
-  #   - "redis"
-  #   environment:
-  #     REGION: eu-central-1
-  #     HASURA_ENDPOINT: http://hasura-graphql:8080
-  #     HASURA_ADMIN_SECRET: myadminsecretkey
-  #     REDIS_CONNECTION_STRING: redis://redis
-  #     PGHOST: postgres
-  #     PGPORT: 5432
-  #     PGUSER: postgres
-  #     PGPASSWORD: postgrespassword
-  #     PGDATABASE: postgres
-  #     PORT: 9180
-  #     AWS_ACCESS_KEY_ID:
-  #     AWS_SECRET_ACCESS_KEY:
+  runner:
+    build:
+      context: ./runner
+    depends_on:
+    - "hasura-graphql"
+    - "redis"
+    environment:
+      REGION: eu-central-1
+      HASURA_ENDPOINT: http://hasura-graphql:8080
+      HASURA_ADMIN_SECRET: myadminsecretkey
+      REDIS_CONNECTION_STRING: redis://redis
+      PGHOST: postgres
+      PGPORT: 5432
+      PGUSER: postgres
+      PGPASSWORD: postgrespassword
+      PGDATABASE: postgres
+      PORT: 9180
+      AWS_ACCESS_KEY_ID:
+      AWS_SECRET_ACCESS_KEY:
 
-  # redis:
-  #   image: redis
-  #   command:
-  #     - redis-server
-  #     - "--save 60 1"
-  #     - "--loglevel warning"
-  #   volumes:
-  #     - redis:/data
-  #   ports:
-  #     - "6379:6379"
+  redis:
+    image: redis
+    command:
+      - redis-server
+      - "--save 60 1"
+      - "--loglevel warning"
+    volumes:
+      - redis:/data
+    ports:
+      - "6379:6379"
 
-  # postgres:
-  #   image: postgres:12
-  #   restart: always
-  #   volumes:
-  #   - postgres:/var/lib/postgresql/data
-  #   environment:
-  #     POSTGRES_PASSWORD: postgrespassword
-  #   ports:
-  #     - "5432:5432"
+  postgres:
+    image: postgres:12
+    restart: always
+    volumes:
+    - postgres:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: postgrespassword
+    ports:
+      - "5432:5432"
 
-  # hasura-auth:
-  #   build:
-  #     context: ./hasura-authentication-service
-  #   ports:
-  #   - "4000:4000"
-  #   environment:
-  #     PORT: 4000
-  #     DEFAULT_HASURA_ROLE: append
+  hasura-auth:
+    build:
+      context: ./hasura-authentication-service
+    ports:
+    - "4000:4000"
+    environment:
+      PORT: 4000
+      DEFAULT_HASURA_ROLE: append
 
-  # hasura-graphql:
-  #   image: hasura/graphql-engine:latest
-  #   ports:
-  #   - "8080:8080"
-  #   depends_on:
-  #   - "postgres"
-  #   - "hasura-auth"
-  #   restart: always
-  #   environment:
-  #     HASURA_GRAPHQL_DATABASE_URL: postgres://postgres:postgrespassword@postgres:5432/postgres
-  #     HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
-  #     HASURA_GRAPHQL_DEV_MODE: "true"
-  #     HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
-  #     HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
-  #     HASURA_GRAPHQL_AUTH_HOOK: http://hasura-auth:4000/auth
+  hasura-graphql:
+    image: hasura/graphql-engine:latest
+    ports:
+    - "8080:8080"
+    depends_on:
+    - "postgres"
+    - "hasura-auth"
+    restart: always
+    environment:
+      HASURA_GRAPHQL_DATABASE_URL: postgres://postgres:postgrespassword@postgres:5432/postgres
+      HASURA_GRAPHQL_ENABLE_CONSOLE: "true"
+      HASURA_GRAPHQL_DEV_MODE: "true"
+      HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
+      HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
+      HASURA_GRAPHQL_AUTH_HOOK: http://hasura-auth:4000/auth
   grafana:
     image: grafana/grafana
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -88,6 +88,21 @@ services:
       HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
       HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
       HASURA_GRAPHQL_AUTH_HOOK: http://hasura-auth:4000/auth
+  grafana:
+    image: grafana/grafana
+    volumes:
+      - grafana:/var/lib/grafana
+    ports:
+      - "3000:3000"
+    environment:
+      - GF_SECURITY_ADMIN_PASSWORD=secret
+    
+  prometheus:
+    image: prom/prometheus
+    volumes:
+      - ./prometheus.yml:/etc/prometheus/prometheus.yml
+    ports:
+      - "9090:9090"
       
 volumes:
   postgres:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,93 +1,93 @@
 version: "3.9"  # optional since v1.27.0
 services:
 
-  coordinator:
-    build:
-      context: ./indexer
-      args:
-        - CARGO_BUILD_MODE=debug
-    depends_on:
-      - redis
-    environment:
-      REDIS_CONNECTION_STRING: redis://redis
-      LAKE_AWS_ACCESS_KEY:
-      LAKE_AWS_SECRET_ACCESS_KEY:
-      QUEUE_AWS_ACCESS_KEY:
-      QUEUE_AWS_SECRET_ACCESS_KEY:
-      QUEUE_URL: MOCK
-      START_FROM_BLOCK_QUEUE_URL: MOCK
-      PORT: 9180
-      REGISTRY_CONTRACT_ID: dev-queryapi.dataplatform.near
-      AWS_QUEUE_REGION: eu-central-1
-    command:
-      - mainnet
-      - from-interruption
+  # coordinator:
+  #   build:
+  #     context: ./indexer
+  #     args:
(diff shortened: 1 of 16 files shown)


darunrs:
https://github.com/near/queryapi/issues/264

runner/src/stream-handler/worker.ts
```diff
@@ -19,40 +30,73 @@ void (async function main () {
...
+async function streamerMessageQueueProducer (queue: Array<Promise<QueueMessage>>, streamKey: string): Promise<void> {
   while (true) {
-    try {
-      const startTime = performance.now();
+    const preFetchCount = HISTORICAL_BATCH_SIZE - queue.length;
+    if (preFetchCount <= 0) {
+      await sleep(300); // Wait for more messages in array to process
+      continue;
+    }
+    const messages = await redisClient.getNextStreamMessage(streamKey, preFetchCount);
```
morgsmccauley, line 51: `getNextStreamMessage` will, by default, return messages starting from the beginning of the stream. Instead, we should maintain a 'current position/id' and feed that in. Without it, we'll be re-reading the same messages and putting them on to the queue.

runner/src/stream-handler/worker.ts
```diff
@@ -19,40 +30,73 @@ void (async function main () {
...
+  };
 
-      if (messages == null) {
-        await sleep(1000);
-        continue;
-      }
+  while (true) {
+    const startTime = performance.now();
+    const blockStartTime = startTime;
+    const queueMessage = await queue.shift();
```
morgsmccauley, line 82: We might not want to `shift()` right away - removing _before_ we successfully execute means that the message will be lost if the executing fails. And since it wouldn't be deleted, I guess it would then be put back on to the queue but out of order.

Instead we should just read the first element, and then `.shift()` after success.

runner/src/stream-handler/worker.ts
morgsmccauley, line 51: Maybe we should [`BLOCK`](https://redis.io/commands/xread/) here rather `sleep`ing, that would mean we are reacting as quick as possible to new messages - could you experiment with that please?

runner/src/stream-handler/worker.ts
```diff
@@ -19,40 +30,73 @@ void (async function main () {
...
+  void streamerMessageQueueConsumer(queue, streamKey);
+}
 
+async function streamerMessageQueueProducer (queue: Array<Promise<QueueMessage>>, streamKey: string): Promise<void> {
   while (true) {
-    try {
-      const startTime = performance.now();
+    const preFetchCount = HISTORICAL_BATCH_SIZE - queue.length;
+    if (preFetchCount <= 0) {
+      await sleep(300); // Wait for more messages in array to process
```
morgsmccauley, line 48: Nit: lets remove these comments and the others, the code is descriptive enough as is :)

morgsmccauley: Also let's experiment with smaller numbers to increase reaction speed.

runner/src/stream-handler/worker.ts
```diff
@@ -4,13 +4,24 @@ import promClient from 'prom-client';
...
 if (isMainThread) {
   throw new Error('Worker should not be run on main thread');
 }
 
-const indexer = new Indexer('mainnet');
+const HISTORICAL_BATCH_SIZE = 100;
+const indexer = new Indexer();
 const redisClient = new RedisClient();
+const lakeClient = new LakeClient();
+let isHistorical = false;
```
morgsmccauley, line 18: Nit: Global variables like this are convenient, but generally discouraged because it opens up the possibility for this to be changed by many different things, which can become difficult to track.

Let's scope this to `main()` or ideally, closer to where it is used.

runner/src/stream-handler/worker.ts
```diff
@@ -4,13 +4,24 @@ import promClient from 'prom-client';
...
+
+interface QueueMessage {
+  streamerMessage: StreamerMessage
+  streamId: string
+}
+const queue: Array<Promise<QueueMessage>> = [];
```
morgsmccauley, line 24: Same comment here around global variables - let's put this in `main`

runner/src/stream-handler/worker.ts
```diff
@@ -19,40 +30,73 @@ void (async function main () {
...
-          function_name: indexerConfig.function_name,
-          code: indexerConfig.code,
-          schema: indexerConfig.schema,
-          provisioned: false,
-        },
-      };
-      await indexer.runFunctions(Number(message.block_height), functions, isHistorical, {
-        provision: true,
-      });
+    try {
```
morgsmccauley, line 95: Let's wrap as much as possible with this `try` block to avoid errors escaping this context, ideally anything that is `await`ed, should be wrapped.

runner/src/stream-handler/worker.ts
```diff
@@ -19,40 +30,73 @@ void (async function main () {
...
+      await sleep(1000); // Wait for new messages to appear in stream
+      continue;
+    }
+    console.log(`Fetched ${messages?.length} messages from stream ${streamKey}`);
 
-      const messages = await redisClient.getNextStreamMessage(streamKey);
-      const indexerConfig = await redisClient.getStreamStorage(streamKey);
+    for (const streamMessage of messages) {
+      const { id, message } = streamMessage;
+      fetchAndQueue(queue, Number(message.block_height), id);
```
morgsmccauley, line 60:
```suggestion
      queue.push(transformStreamerMessageToQueueMessage(blockHeight, id));
```
Nit: I think this code is simple/descriptive enough to just be inlined?

runner/src/stream-handler/worker.ts
```diff
@@ -61,10 +105,22 @@ void (async function main () {
...
     }
   }
-})();
+}
+
+function fetchAndQueue (queue: Array<Promise<QueueMessage>>, blockHeight: number, id: string): void {
+  queue.push(transformStreamerMessageToQueueMessage(blockHeight, id));
+}
+
+async function transformStreamerMessageToQueueMessage (blockHeight: number, streamId: string): Promise<QueueMessage> {
```
morgsmccauley, line 120: Nit: I feel like `transform` doesn't fully capture the intent here, how about `generateQueueMessage`, as it also has some logic to fetch the streamer message. Probably being pedantic here 😅

morgsmccauley (changes requested):
Really great start on this - left a couple comments :)

runner/src/lake-client/lake-client.ts
```diff
@@ -0,0 +1,97 @@
...
+          })
+          .join('');
+        newValue[newKey] = value[key];
+      }
+      return newValue;
+    }
+    return value;
+  }
+
+  async fetchStreamerMessage (blockHeight: number, isHistorical: boolean): Promise<StreamerMessage> {
```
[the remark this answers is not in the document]
darunrs: You're correct. I forgot to polish up this new class after migrating all the code to it. I'll fix it up!

runner/src/metrics.ts
```diff
@@ -1,35 +1,55 @@
...
+  name: 'queryapi_runner_cache_hit',
+  help: 'The number of times cache was hit successfully'
+});
+
+const CACHE_MISS = new Counter({
+  name: 'queryapi_runner_cache_miss',
+  help: 'The number of times cache was missed'
+});
+
+const USER_CODE_EXECUTION_DURATION = new Histogram({
```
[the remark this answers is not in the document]
darunrs: I suppose that's fair. I'll remove this and make a new issue around instrumentation, if it doesn't already exist.

runner/src/metrics.ts
```diff
@@ -1,35 +1,55 @@
...
+
+const CACHE_HIT = new Counter({
+  name: 'queryapi_runner_cache_hit',
+  help: 'The number of times cache was hit successfully'
+});
+
+const CACHE_MISS = new Counter({
+  name: 'queryapi_runner_cache_miss',
+  help: 'The number of times cache was missed'
+});
```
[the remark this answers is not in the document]
darunrs: I've rebased. It shows up as new because I moved the metrics around to be in order of when they are recorded. This was to make it easier to understand all the metrics at a glance.

docker-compose.yml
```diff
@@ -88,7 +88,23 @@ services:
       HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
       HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
       HASURA_GRAPHQL_AUTH_HOOK: http://hasura-auth:4000/auth
+  grafana:
```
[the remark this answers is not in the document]
darunrs: I guess on one hand I do agree and on another I don't. From a indexer developer perspective, definitely agree. I know we have a goal of eventually supporting local development of indexers and this docker file was a step in that direction. If our goal is to keep this file oriented for that goal, we probably don't need metrics things. On the other hand, I think having metrics would be very helpful to anyone intending to contribute to the repo itself, whether in our team or not. I'm personally leaning towards keeping it in as someone can simply comment out this bit if they don't want it. And if it's running, the impact should be minimal, I think. Let me know if your thoughts align or not!

runner/src/lake-client/lake-client.ts
```diff
@@ -0,0 +1,97 @@
...
+    if (!isHistorical) {
```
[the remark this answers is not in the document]
darunrs: I figured this was the case but I didn't want historical to pollute the cache miss metric since every historical block would fail to hit cache. I guess I can do something like only post the cache miss if it's not historical, but since I already have an if statement, I might as well put it all in there.

runner/src/lake-client/lake-client.ts
```diff
@@ -0,0 +1,97 @@
...
+      const cachedMessage = await this.redisClient.getStreamerMessage(blockHeight);
+      if (cachedMessage) {
+        METRICS.CACHE_HIT.labels().inc();
```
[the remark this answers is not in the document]
darunrs: So, I actually removed the label since the label was if it's real-time or historical. Except, we aren't using the cache if it's historical so I figured there's no label to set and did not set any. Do you think I should still set that information anyway? Should I add both indexer name as well as type? My graphs in grafana got messy when I did so but perhaps that can be fixed by adjusting the query. Let me know what you think. In the meantime, I'll play around with it and if I can get it to look right while adding the labels, I'll do so, and adjust dev's graph when the PR goes through if we end up setting them.

runner/src/stream-handler/worker.ts
darunrs: Yep you're right. I'll look over the process and make sure it's consistent. I think retaining the latest stream ID read and not shifting the queue should be sufficient.

darunrs: Will do!

runner/src/stream-handler/worker.ts
darunrs: A few of them I was able to move easily into one of the functions. But, some of the others like redis client, lake client, and isHistorical are used all over. I moved them into a single object called WorkerContext and will pass that into function calls instead. I couldn't find a way to make private global variables so I don't think I can do anything to make them accessible like a class variable through this. Let me know if there's a cleaner way of doing this!

runner/src/stream-handler/worker.ts
darunrs: Pedantic is fine! Might as well get a good name going the first time. I also like that name. Changed!

darunrs:
I'm still experimenting with the blocking values. In the meantime, I released the changes I've made so far. Let me know what you think!

darunrs:
I ran roughly 5000 messages of sweat blockheight before and after the blocking commit and I can visually see an improvement. Left is with the commit and right is without it. Block wait time is lower and less spread out. Although, it was already low enough to not impact much anymore anyway. Execution duration and overhead latency spread is tighter too. The BPS curve is more linear and the peak was higher too. In any case, it seems the commit helps a fair bit. I will try again but this time run all indexers simultaneously. And finally, I'll check running coordinator and runner together (Although I wont be able to see metrics for that since coordinator's metrics server would block runners'. I plan to look into that later. 

image

runner/src/lake-client/lake-client.ts
```diff
@@ -0,0 +1,94 @@
...
+
+export default class LakeClient {
+  constructor (
+    private readonly network: string = 'mainnet',
+    private readonly s3Client: S3Client = new S3Client(),
+    private readonly redisClient: RedisClient = new RedisClient()
+  ) {
+    this.network = network;
+    this.s3Client = s3Client;
+    this.redisClient = redisClient;
```
morgsmccauley, line 14: This are assigned automatically when you declare the `private readonly` above - you can remove them

runner/src/lake-client/lake-client.ts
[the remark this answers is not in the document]
morgsmccauley: Nah seems like it's not needed, in which case you can remove the `.label()` call

runner/src/redis-client/redis-client.ts
```diff
@@ -44,12 +44,15 @@ export default class RedisClient {
     await this.client.disconnect();
   }
 
-  async getNextStreamMessage (
+  async getStreamMessages (
     streamKey: string,
+    count = 1,
+    streamId = this.SMALLEST_STREAM_ID,
```
morgsmccauley, line 50:
```suggestion
    streamId = this.SMALLEST_STREAM_ID,
    count = 1,
```
Can we swap these, `count` seems less likely to change that `streamId`, so putting it after means we don't need to define `count` every time

runner/src/stream-handler/worker.ts
```diff
@@ -2,69 +2,152 @@ import { isMainThread, parentPort, workerData } from 'worker_threads';
...
 
-      const messages = await redisClient.getNextStreamMessage(streamKey);
-      const indexerConfig = await redisClient.getStreamStorage(streamKey);
+    streamMessageStartId = incrementId(messages[messages.length - 1].id);
+  }
+}
 
-      indexerName = `${indexerConfig.account_id}/${indexerConfig.function_name}`;
+async function blockQueueConsumer (workerContext: WorkerContext, streamKey: string): Promise<void> {
+  const streamType = workerContext.redisClient.getStreamType(streamKey);
```
morgsmccauley, line 89: This is already in the `workerContext` right?

runner/src/stream-handler/worker.ts
```diff
@@ -2,69 +2,152 @@ import { isMainThread, parentPort, workerData } from 'worker_threads';
...
+  });
+}
+
+async function blockQueueProducer (workerContext: WorkerContext, streamKey: string): Promise<void> {
+  let streamMessageStartId = '0';
 
   while (true) {
-    try {
-      const startTime = performance.now();
+    const preFetchCount = await waitForQueueSpace(workerContext);
```
morgsmccauley, line 72:
```suggestion
    const preFetchCount = HISTORICAL_BATCH_SIZE - workerContext.queue.length;
    if (preFetchCount <= 0) {
      await sleep(100);
      continue;
    }
```

Seems simpler just to add this here rather than creating an interval?

runner/src/redis-client/redis-client.ts
darunrs: Ah good point. Switched!

runner/src/stream-handler/worker.ts
darunrs: Ah oops. it is! Fixed!

darunrs:
I did some more experimentation and found that blocking does in fact block both loops, not jsut the producer. As a result, once no stream messages are present, execution of the functions becomes slow as the xread blocks the whole thread for the blocking duration. Removing it returned the speed improvements. 

In addition, I ran a test against sweat_blockheight where I appended the block-height of each stream message to one file and then appended the block height to a second file after runFunction is called. I diff'd the two files and found no differences. This was enough to verify that every message in the stream is seen and processed exactly once.

morgsmccauley (approved):
LGTM - great work :)

morgsmccauley:
Just need to fix failing tests

PR #355, opened 2023-11-01
refactor: Aggregate worker metrics in background

morgsmccauley:
Previously, we would need to manually send the metric data to the main thread, then write the metric there. This PR updates worker metrics so it is done seamlessly in the background, metrics can be written as normal regardless of whether in the main thread or not. This is achieved by sending all metrics collected within a worker to the main thread, and then aggregating them there.

### DIFF
--- a/runner/src/stream-handler/stream-handler.ts
+++ b/runner/src/stream-handler/stream-handler.ts
@@ -4,7 +4,7 @@ import { Worker, isMainThread } from 'worker_threads';
 import { registerWorkerMetrics } from '../metrics';
 
 export default class StreamHandler {
-  private readonly worker?: Worker;
+  private readonly worker: Worker;
 
   constructor (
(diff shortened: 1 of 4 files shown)


runner/src/stream-handler/stream-handler.ts
```diff
@@ -1,12 +1,10 @@
...
 import { Worker, isMainThread } from 'worker_threads';
 
-import { type Message } from './types';
-import { METRICS } from '../metrics';
-import { Gauge, Histogram } from 'prom-client';
+import { registerWorkerMetrics } from '../metrics';
 
 export default class StreamHandler {
-  private readonly worker?: Worker;
+  private readonly worker: Worker;
```
morgsmccauley, line 7: This will always be defined, otherwise the constructor will throw

darunrs (approved):

Issue #216, opened 2023-09-26
Handoff between real-time and historical processing


Ishatt:
Description

Problem:

Currently we do not coordinate between real-time and historical processing, indexers can run on blocks out of order. 

See Confluence pages

See [Real-time to Historical communication](https://pagodaplatform.atlassian.net/wiki/spaces/DAT/pages/333545513/Real-time+to+Historical+communication) for architectural concerns.

Proposed Solutions:

Append real-time blocks to be processed to the processing queue of the historical thread.

After the move to Redis we could store a redis flag for each function for historical_backfill_in_progress . When this is set at the start of historical backfill, real-time processing can skip processing until both the flag is cleared and the historical queue is empty. Once the historical queue has been filled the flag can be cleared.
```[tasklist]
### Linked Issues
- [ ] https://github.com/near/queryapi/issues/198
```

```[tasklist]
### Relates to
```

roshaans:
@ailisp would like to create a "sql join query" in query-api, but is unable to do so without Hasura foriegn-key relationships which he is unable to do as QueryAPI currently processes historical backfill in parallel with real-time blocks and in order to create forign-key relationships, you have to impose an order of insertions

object X references object Y, but X can be inserted before Y that result in foreign key constraint violation.

morgsmccauley:
One way to solve this is for each indexer to have their own 'real time' block process, rather than the shared one we have now. This would make the handover much simpler, as it could just 'start' after historical has completed.

Issue #346, opened 2023-10-31
Delete all resources when deleting indexer


morgsmccauley:
Deleting the indexer from the UI will only remove it from the registry contract, but all the infrastructure resources will remain. The following should be deleted along with the registry entry:
- Related redis keys
- Hasura configuration
- Database schema

```[tasklist]
### Tasks
- [ ] https://github.com/near/queryapi/issues/759
- [ ] https://github.com/near/queryapi/issues/760
- [ ] https://github.com/near/queryapi/issues/761
- [ ] https://github.com/near/queryapi/issues/792
```

morgsmccauley:
This may be tricky with Coordinator viewing the snapshots of the Registry at single points in time. Coordinator compares the registry to the system directly, so will only know of a deletion if the system is actually running that deleted indexer, but if we restart _during_ a deletion, the event will be missed.

We may need to introduce some persistent state which Coordinator can refer to as to gain more awareness of these events.

morgsmccauley:
Another perhaps more straight forward approach would be to query Hasura/Postgres to the current provisioning state, which could be compared to the registry. This removes the need for capturing specific point-in-time events, and allows us to simply synchronise actual/desired states exactly like we are for Block streams/Excutors.

PR #363, opened 2023-11-06
feat: Cancel historical backfill process before starting anew

### LINKED ISSUE #200: Stop historical processing on indexer redeployment
Description

When saving the new code, user has an option to “start over” – wipe the database, replace the code, and restart historical processing from the selected block

Coordinator should detect this option in registry indexing

Coordinator should purge the SQS queue using the group id of this indexer, or the runner should ignore all messages. This might need to be refactored to use Redis Streams instead (@Eduardo Ohe will investigate how Runners can be hooked to Redis Streams https://redis.io/docs/data-types/streams-tutorial/#streams-observability )

Coordinator should send a re-provision message to the runner so that it deletes the database schema for this indexer and provisions it again. Runner should update the state of the indexer and keep user updated in the logs

Coordinator should re-start the historical thread from the provided block height

morgsmccauley:
This PR updates Coordinator such that **only one** Historical Backfill process can exist for a given Indexer Function. Currently, these processes do not have awareness of each other, meaning a developer can trigger multiple simultaneous backfills, with them all writing to the same Redis Stream. The major changes are detailed below.

## Streamer/Historical backfill tasks
In this PR, I've introduced the `historical_block_processing::Streamer` struct. This is used to encapsulate the async task which is spawned for handling the Historical Backfill. I went with "Streamer", as I envision this becoming the single process which handles both historical/real-time messages for a given Indexer, as described in https://github.com/near/queryapi/issues/216#issuecomment-1777938653.

`Streamer` is responsible for starting the `async` task, and provides a basic API for interacting with it. Right now, that only includes cancelling it. Cancelling a `Streamer` will drop the existing future, preventing any new messages from being added, and then delete the Redis Stream removing all existing messages.

## Indexer Registry Mutex/Async
`indexer_registry` is the in-memory representation of the given Indexer Registry. Currently, we pass around the `MutexGuard` to mutate this data, resulting in the lock being held for much longer than is needed, blocking all other consumers from taking the lock. There isn't much contention for this lock, so it isn't a problem right now, but could end up being one in future.

Adding in the `Streamer` `Mutex` made mitigating this problem trivial, so I went ahead and made some changes. I've updated the code so that we pass around the `Mutex` itself, rather than the guard, allowing us to the hold the lock for the least amount of time possible. This required updating most methods within `indexer_registry` to `async`, so that we could take the `async` `lock`.

With the `Mutex` being used rather than the `MutexGuard`, it seemed sensible to add `indexer_registry` to `QueryAPIContext`, rather than passing it round individually.

### DIFF
--- a/indexer/queryapi_coordinator/src/historical_block_processing.rs
+++ b/indexer/queryapi_coordinator/src/historical_block_processing.rs
@@ -26,45 +26,43 @@ pub struct Streamer {
 
 pub fn spawn_historical_message_thread(
     current_block_height: BlockHeight,
-    new_indexer_function: &IndexerFunction,
+    indexer_function: &IndexerFunction,
     redis_connection_manager: &storage::ConnectionManager,
     s3_client: &S3Client,
     chain_id: &ChainId,
     json_rpc_client: &JsonRpcClient,
-) -> Option<Streamer> {
-    new_indexer_function.start_block_height.map(|_| {
-        let new_indexer_function_copy = new_indexer_function.clone();
-        let redis_connection_manager = redis_connection_manager.clone();
-        let s3_client = s3_client.clone();
-        let chain_id = chain_id.clone();
-        let json_rpc_client = json_rpc_client.clone();
+) -> Streamer {
+    let indexer = indexer_function.clone();
+    let redis_connection_manager = redis_connection_manager.clone();
+    let s3_client = s3_client.clone();
+    let chain_id = chain_id.clone();
+    let json_rpc_client = json_rpc_client.clone();
 
-        let cancellation_token = tokio_util::sync::CancellationToken::new();
-        let cancellation_token_clone = cancellation_token.clone();
+    let cancellation_token = tokio_util::sync::CancellationToken::new();
+    let cancellation_token_clone = cancellation_token.clone();
 
-        let handle = tokio::spawn(async move {
-            tokio::select! {
-                _ = cancellation_token_clone.cancelled() => {
-                    0
-                },
-                processed_block_count = process_historical_messages_or_handle_error(
-                    current_block_height,
-                    new_indexer_function_copy,
-                    &redis_connection_manager,
-                    &s3_client,
-                    &chain_id,
-                    &json_rpc_client,
-                ) => {
-                    processed_block_count
-                }
+    let handle = tokio::spawn(async move {
+        tokio::select! {
+            _ = cancellation_token_clone.cancelled() => {
+                0
+            },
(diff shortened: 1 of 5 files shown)


indexer/queryapi_coordinator/src/historical_block_processing.rs
```diff
@@ -17,45 +17,91 @@ pub const INDEXED_ACTIONS_FILES_FOLDER: &str = "silver/accounts/action_receipt_a
...
+
+        Err(anyhow::anyhow!(
+            "Attempted to cancel already cancelled, or not started, Streamer"
+        ))
+    }
 }
 
 pub(crate) async fn process_historical_messages_or_handle_error(
-    block_height: BlockHeight,
+    current_block_height: BlockHeight,
```
morgsmccauley, line 96: I got confused about what `block_height` this actually represents, so ended up renaming it

indexer/queryapi_coordinator/src/indexer_registry.rs
```diff
@@ -8,25 +8,13 @@ use near_lake_framework::near_indexer_primitives::views::QueryRequest;
...
-use tokio::sync::MutexGuard;
-use tokio::task::JoinHandle;
 use unescape::unescape;
 
 use crate::indexer_reducer;
 use crate::indexer_reducer::FunctionCallInfo;
 use crate::indexer_types::{IndexerFunction, IndexerRegistry};
 use indexer_rule_type::indexer_rule::{IndexerRule, IndexerRuleKind, MatchingRule, Status};
 
-pub(crate) fn registry_as_vec_of_indexer_functions(
```
morgsmccauley, line 20: Now done inline

indexer/queryapi_coordinator/src/indexer_registry.rs
```diff
@@ -220,7 +212,12 @@ fn index_and_process_remove_calls(
                         function_invocation.account_id.clone(),
                         function_invocation.function_name.clone(),
                     );
-                    match registry.entry(function_invocation.account_id.clone()) {
+                    match context
+                        .indexer_registry
+                        .lock()
+                        .await
+                        .entry(function_invocation.account_id.clone())
+                    {
```
gabehamilton, line 220: Should we also clean up streams here?

gabehamilton (approved):

indexer/queryapi_coordinator/src/indexer_registry.rs
morgsmccauley: Yes, good call. I'll do that in this issue https://github.com/near/queryapi/issues/346

PR #365, opened 2023-11-06
refactor: Remove SQS references from Coordinator

### LINKED ISSUE #201: Clean up SQS code in Coordinator
Once we’re finished with the Redis Migration, we should remove all the SQS specific code.

morgsmccauley:
Doing some clean-up to make the development of #200 a bit easier. Now that we have moved completely to Redis, this PR removes all SQS code/references from Coordinator. 

Additionally, I did some small refactoring to `QueryApiContext`, adding both `s3_client` and `json_rpc_client`, allowing a single instance of each to be passed around and used throughout the code. Previously, many `s3_client`s were being instantiated in various parts of the code. This removed the dependancy on `Opts` from Historical Backfill, so that has also been removed.

Finally, I fixed all Clippy errors :)


indexer/queryapi_coordinator/src/historical_block_processing.rs
```diff
@@ -26,16 +21,25 @@ pub fn spawn_historical_message_thread(
     block_height: BlockHeight,
     new_indexer_function: &IndexerFunction,
     redis_connection_manager: &storage::ConnectionManager,
+    s3_client: &S3Client,
+    chain_id: &ChainId,
+    json_rpc_client: &JsonRpcClient,
 ) -> Option<JoinHandle<i64>> {
     let redis_connection_manager = redis_connection_manager.clone();
+    let s3_client = s3_client.clone();
```
darunrs, line 29: Why do we clone these instead of using the referenced client directly?

darunrs (commented):
We should also update the docker image to remove the aws queue references there. Also, has the changes been deployed to the AWS account to remove those resources already?

indexer/queryapi_coordinator/src/historical_block_processing.rs
morgsmccauley: `tokio::spawn` requires, and takes, ownership of the variables referenced, this is signified by the `move` keyword. The compiler does not know how long the async task (`tokio::spawn`) will last, and therefore cannot determine whether the reference will live long enough. 

Cloning allows us to take ownership of the variable, which then means we can move it to the async task.

indexer/queryapi_coordinator/src/historical_block_processing.rs
```diff
@@ -331,11 +313,14 @@ async fn filter_matching_unindexed_blocks_from_lake(
...
-        let s3_result = s3::fetch_text_file_from_s3(&lake_bucket, key, s3_client.clone()).await;
+        let s3_result = s3::fetch_text_file_from_s3(&lake_bucket, key, s3_client).await;
 
         if s3_result.is_err() {
             let error = s3_result.err().unwrap();
-            if let Some(_) = error.downcast_ref::<aws_sdk_s3::error::NoSuchKey>() {
+            if error
+                .downcast_ref::<aws_sdk_s3::error::NoSuchKey>()
+                .is_some()
+            {
```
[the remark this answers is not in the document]
morgsmccauley: Nah it's just a clippy suggestion

morgsmccauley:
> We should also update the docker image to remove the aws queue references there. Also, has the changes been deployed to the AWS account to remove those resources already?

The docker image doesn't reference AWS directly, but I assume you are referring to the configuration of the the docker container in GCP? I'll eventually raise another PR to remove that, as well as the terraform config for both GCP and AWS. For now, they'll just be silently ignored.

darunrs (approved):

darunrs:
Sweet! Thanks for taking care of all the clippy suggestions!

gabehamilton (approved):

PR #381, opened 2023-11-09
Helper functions to simplify writing social indexers.

### LINKED ISSUE #366: Social operations context helper functions
## The problem
Indexer functions would be easier to read if there were helper functions to extract common data.

## Proposed Solution
Add the following context methods:
 * `base64decode` Every indexer currently defines it.
 *  `getActionsForFunctionCall(contract, method)`
 *  `getFunctionCalls(contract, method)` 
 *  `getSocialOperations(operation)`

Decide whether to pass `block` in method signature.

## Proposed code blocks
`base64decode(encodedValue)`
```
    let buff = Buffer.from(encodedValue, "base64");
    return JSON.parse(buff.toString("utf-8"));
```

`getActionsForFunctionCall(contract, method)` 
```
block
    .actions()
    .filter((action) => action.receiverId === contract)
    .flatMap((action) =>
      action.operations
        .map((operation) => operation["FunctionCall"])
        .filter((operation) => operation?.methodName === method)
```

`getFunctionCalls(contract, method)`
```
 getActionsForFunctionCall(contract, method)
 .map((functionCallOperation) => ({
          ...functionCallOperation,
          args: base64decode(functionCallOperation.args),
          receiptId: action.receiptId,
        }))
```

`getSocialOperations(operation)`
```
const contract = "social.near";
const method = "set";
getFunctionCalls(contract, method)
.filter((functionCall) => {
        const accountId = Object.keys(functionCall.args.data)[0];
        return (
          functionCall.args.data[accountId][operation]
```

gabehamilton:
- See linked [issue](https://github.com/near/queryapi/issues/366) for the proposal that this implements.
 - See [Test Indexer](https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=flatirons.near/test_helper_methods&view=editor-window)

### DIFF
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -19,6 +19,13 @@ interface Context {
   log: (...log: any[]) => Promise<void>
   fetchFromSocialApi: (path: string, options?: any) => Promise<any>
   db: Record<string, Record<string, (...args: any[]) => any>>
+  base64Decode: (base64EncodedString: string) => string
+  getFunctionCalls: (
+    block: Block,
+    contract: string,
+    method: string,
+  ) => Array<{ Operation, args: string }>
+  getSocialOperations: (block: Block, operation: string) => Array<Record<string, unknown>>
 }
 
 interface IndexerFunction {
@@ -140,6 +147,25 @@ export default class Indexer {
     const functionNameWithoutAccount = functionName.split('/')[1].replace(/[.-]/g, '_');
     const schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
 
+    const base64DecodeFunction = (base64EncodedString: string): string => {
+      const buff = Buffer.from(base64EncodedString, 'base64');
+      return JSON.parse(buff.toString('utf-8'));
+    };
+    const getFunctionCallsFunction = (block: Block, contract: string, method: string): any[] => {
+      return block
+        .actions()
+        .filter((action) => action.receiverId === contract)
+        .flatMap((action) =>
+          action.operations
+            .map(({ FunctionCall }) => FunctionCall)
+            .filter((operation) => operation?.methodName === method)
+            .map((functionCallOperation) => ({
+              ...functionCallOperation,
+              args: base64DecodeFunction(functionCallOperation.args),
+            })),
+        );
+    };
+
     return {
       graphql: async (operation, variables) => {
         console.log(`${functionName}: Running context graphql`, operation);
@@ -164,7 +190,25 @@ export default class Indexer {
       fetchFromSocialApi: async (path, options) => {
         return await this.deps.fetch(`https://api.near.social${path}`, options);
       },
-      db: this.buildDatabaseContext(account, schemaName, schema, blockHeight)
+      db: this.buildDatabaseContext(account, schemaName, schema, blockHeight),
+      base64Decode: base64DecodeFunction,
+      getFunctionCalls: getFunctionCallsFunction,
+      getSocialOperations: (block, operation) => {
+        const contract = 'social.near';
+        const method = 'set';
+        return getFunctionCallsFunction(block, contract, method)
+          .filter((functionCall) => {
+            const accountId = Object.keys(functionCall.args.data)[0];
+            return functionCall.args.data[accountId][operation];
+          })
+          .map((functionCall) => {
+            const accountId = Object.keys(functionCall.args.data)[0];
+            return {
+              accountId,
+              data: functionCall.args.data[accountId][operation],
+            };
+          });
+      },
     };
   }
 
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -13,6 +13,15 @@ interface Dependencies {
   parser: Parser
 };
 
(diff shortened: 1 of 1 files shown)


gabehamilton:
Please weigh in on naming.
Currently the helper functions are named:
```
base64Decode
getFunctionCalls
getSocialOperations
```
One alternative would be to add the extraction methods to the `block` instead of the context and call them
```
functionCalls
socialOperations

```

runner/src/indexer/indexer.ts
```diff
@@ -140,6 +156,26 @@ export default class Indexer {
...
+        .actions()
+        .filter((action) => action.receiverId === contract)
+        .flatMap((action) =>
+          action.operations
+          // @ts-expect-error block data is an object keyed by type
+            .map(({ FunctionCall }) => FunctionCall)
+            .filter((operation) => operation?.methodName === method)
+            .map((functionCallOperation) => ({
+              ...functionCallOperation,
+              args: base64DecodeFunction(functionCallOperation.args),
```
darunrs, line 174: It might be possible for the decode to fail depending on the structure of args. I vaguely recall a block which failed there. I would either wrap this in a try catch and return an empty result. The user is not able to control the contents of a block or this function code. So, if it throws errors, they need to be aware that some blocks *will* crash their code and stop their indexer if they dont handle the error. Or we can just handle it, which I feel is the better option.

runner/src/indexer/indexer.ts
```diff
@@ -164,7 +200,25 @@ export default class Indexer {
       fetchFromSocialApi: async (path, options) => {
         return await this.deps.fetch(`https://api.near.social${path}`, options);
       },
-      db: this.buildDatabaseContext(account, schemaName, schema, blockHeight)
+      db: this.buildDatabaseContext(account, schemaName, schema, blockHeight),
+      base64Decode: base64DecodeFunction,
+      getFunctionCalls: getFunctionCallsFunction,
+      getSocialOperations: (block, operation) => {
```
darunrs, line 206: Fortunately, I do actually have a test block for this piece of code. The block `100505912` has social posts with malformed data. As a result, the below filter and map functions will actually fail, throwing an uncaught error that will stop your indexer. You can run that block agaisnt the indexer you made and see that it does in fact fail. The fix I did was to simply wrap the calls in a try catch. You can see the fix here: https://github.com/near/near-discovery-components/blob/develop/indexers/social_feed/social_feed.js#L271-L283

runner/src/indexer/indexer.ts
```diff
@@ -164,7 +200,25 @@ export default class Indexer {
...
+        const contract = 'social.near';
+        const method = 'set';
+        return getFunctionCallsFunction(block, contract, method)
```
darunrs, line 209: Can't we simply put the contract and method in the function call, since they're constants anyway? Otherwise, if we want to keep the variables to declare the constants, we should place them as a class level constant with the names `SOCIAL_CONTRACT` and `SET_METHOD`.

darunrs (commented):
The functions look very helpful! I'd love to rewrite social feed with them, to get that indexer even shorter. Left some comments around error handling. I'd love to see some unit tests too, to ensure the new functions are error tolerant.

darunrs:
Regarding the naming, they look fine to me. For the socialOperations one, I have a question. The name implies there are different operations you can execute on the social.near contract. I have only seen the set method. What other methods are used commonly? 

I think I would like to keep socialOperations in this. But decode and get functions could possibly go in Block. I'd say we can keep the PR as is for now and migrate if we want to, when we look into supporting things like other Actions.

gabehamilton:
> Regarding the naming, they look fine to me. For the socialOperations one, I have a question. The name implies there are different operations you can execute on the social.near contract. I have only seen the set method. What other methods are used commonly?

Maybe this needs a naming discussion as well. It's not the method it's the top-level data element being stored. `post`, `graph`, `settings`, `widget`, `profile`, anythings else anyone decides to store. `index` is a special one. The next level below it are `like`, `notify`, `star`, anything someone decides to store. Those get indexed by [api-server-js](https://github.com/NearSocial/api-server-js/blob/main/src/server.js#L70).

Maybe this should be called path and arguments like  `/index/like` would match operations where `index: { like: ` data is being set?

darunrs:
Gotcha. My comment was mainly around if it should be socialOperations or socialPosts (as social post is what its called in the social feed indexer). But, from what you just explained, socialOperations is probably better. I think path would make it seem more like an API. I think going as far as paths is reaching a bit too far. We would need to keep that up to date. I would say its better to let the user handle parsing the data (such as seeing if a post contains a like, or a post) themselves. We just strip out parts of the block that, if you call the function, we know you don't want/need.

pkudinov:
Adding these methods to `context` means we have to support them perpetually until the last indexer upgrades to another solution. They all (besides the social methods) belong to near primitives, so we should prioritize adding them there instead.  

For socialOperations, I think this deserves to be a separate library altogether, because this is not generic enough and does not adhere to any NEP standard, and therefore subject to change. So maintaining it inside near-primitives will involve version management, migrations, etc.

Perhaps for now you can keep them as tidy snippet of code you keep in gist and docs, so that users can copy-paste it into their indexers?

morgsmccauley:
To be moved to `near-lake-primitives`

PR #387, opened 2023-11-13
refactor: Remove async `indexer_rules_engine`

morgsmccauley:
We have two versions of `indexer_rules_engine`; sync and async. This logic does not need to be `async` so I've removed this version and refactored so only the sync version is used.

Additionally, I've migrated the `async` tests across to use the `sync` version, and have added `cargo test` to Github Actions so that they are always run. This required fixing the `set_provisioning_flag` test, as well as ignoring the integration tests which rely on AWS access keys.


indexer/indexer_rules_engine/blocks/93085141.json
```diff
@@ -0,0 +1,3551 @@
+{
```
morgsmccauley, line 1: Added as `indexer_rules_engine` tests depend on it

indexer/queryapi_coordinator/src/s3.rs
```diff
@@ -204,6 +204,7 @@ mod tests {
     /// Parses env vars from .env, Run with
     /// cargo test s3::tests::list_delta_bucket -- mainnet from-latest;
     #[tokio::test]
+    #[ignore]
```
darunrs, line 207: Did you intend to leave all of these ignores around?

darunrs (approved):
LGTM!

indexer/queryapi_coordinator/src/s3.rs
morgsmccauley: They do, but they rely on external sources and AWS access keys so I didn't want to run them in CI. They're mostly run locally so in that case the `#[ignore]` can be removed then.

PR #413, opened 2023-11-18
Fix Connection Unavailable Error

### LINKED ISSUE #340: Resolve DB Connection Issues and Support Transactions
see https://github.com/near/queryapi-mvp/blob/0c9fa36142538e5d009b633209272fa412c47560/runner/src/pg-client.ts#L33

darunrs:
So after a lot of testing, I found the problem. It turns out that the lazy loading implementation I used for context.db was set up so that any calls to a db function would first check if the dmlHandler exists (Load on access). If not, then instantiate it. This is helpful for situations where the functions aren't used. However, this had a side effect: If a bunch of db function calls come in simultaneously, they will all poll dmlHandler and see it is uninitialized and all try and create a new one. Each initialization will try to create a pool with 10 available clients. And then all of them will create at least one client each. This will exhaust the clients very quick if there's enough of them. Changing it so that the dmlHandler is lazy loaded immediately instead. And on access, it will await its completion before calling any functions. This ensures there is no more than 1 client active for that invocation. I wrote a test as well which tries 100 simultaneous calls and verified 1 pool is created.

Transaction Support can come next, now that the pooling works correctly now.

### DIFF
--- a/runner/src/dml-handler/dml-handler.ts
+++ b/runner/src/dml-handler/dml-handler.ts
@@ -14,6 +14,7 @@ export default class DmlHandler {
     hasuraClient: HasuraClient = new HasuraClient(),
     PgClient = PgClientModule
   ): Promise<DmlHandler> {
+    console.log('CALL MADE');
     const connectionParameters = await hasuraClient.getDbConnectionParameters(account);
     const pgClient = new PgClient({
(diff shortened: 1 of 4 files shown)


morgsmccauley (approved):
Can you please link the relevant issue

PR #424, opened 2023-11-21
Replace block wait metric with Histogram

### LINKED ISSUE #412: Capture Peaks in Block Wait Time

darunrs:
Current block wait metric uses a gauge, which is causing spikes in block wait to be lost. So, I have switched it to a histogram instead, which does a much better job at capturing peaks and having a variety of more useful expressions, some of which I list below. These expressions are also customizable. 

A more accurate average: 
`rate(queryapi_runner_block_wait_duration_milliseconds_sum[$__rate_interval]) / rate(queryapi_runner_block_wait_duration_milliseconds_count[$__rate_interval])`

The duration under which 95% of block requests were fulfilled
`histogram_quantile(0.95, sum(rate(queryapi_runner_block_wait_duration_milliseconds_bucket[$__rate_interval])) by (le))`

The % of requests for each indexer where the block wait was under 100ms
```
  sum(rate(queryapi_runner_block_wait_duration_milliseconds_bucket{le="100"}[$__rate_interval])) by (indexer)
/
  sum(rate(queryapi_runner_block_wait_duration_milliseconds_count[$__rate_interval])) by (indexer)
```

### DIFF
--- a/runner/src/metrics.ts
+++ b/runner/src/metrics.ts
@@ -1,10 +1,11 @@
 import express from 'express';
 import { Gauge, Histogram, Counter, AggregatorRegistry } from 'prom-client';
 
-const BLOCK_WAIT_DURATION = new Gauge({
+const BLOCK_WAIT_DURATION = new Histogram({
   name: 'queryapi_runner_block_wait_duration_milliseconds',
   help: 'Time an indexer function waited for a block before processing',
   labelNames: ['indexer', 'type'],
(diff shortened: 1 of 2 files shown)


morgsmccauley (approved):

morgsmccauley:
Oh, is this still draft?

darunrs:
I started it as a draft and then switched it to ready haha.

PR #428, opened 2023-11-23
feat: Create initial Block Streamer service

### LINKED ISSUE #418: Created dedicated block stream per Indexer
Rather than managing separate Historical and Real-time streams. We should remove the shared real-time process, and establish a single continuous process per indexer. With this approach, no hand-off is needed as the dedicated real-time stream and simply continue from where Historical left off.

This functionality associates closely with the Data Plane, and therefore should exist as a new service rather than being implemented within Coordinator. Coordinator, once it is Control Plane focussed, will then communicate with this directly to start the necessary Streamers.

We'll need to create all the relevant infrastructure, which makes this a good opportunity to add the `mainnet` separation.

```[tasklist]
#### Relates to
- [ ] https://github.com/near/queryapi/issues/216
```

morgsmccauley:
This PR introduces a new Rust service called Block Streamer. This is essentially a refactored version of Historical Backfill, which does not stop the 'manual filtering' portion of the backfill, continuing indefinitely. Eventually, this service will be able to serve requests, allowing control over these Block Streams. I've outlined the major changes below.

## `trait S3Client`
Low level S3 requests have been abstracted behind a `trait`. This allows us to use the `trait` as an argument, allowing the injection of mock implementations in tests. We no longer need to use real access keys, and can make more concrete assertions.

## `DeltaLakeClient`
The majority of the S3 related logic, `queryapi_coordinator/src/s3.rs` included, has been refactored in to `DeltaLakeClient`. This client encapsulates all logic relating the interaction/consumption of `near-delta-lake` S3. Now, only a single method needs to be called from `BlockStream` in order to get the relevant block heights for a given contract_id/pattern. There's been a fair amount of refactor across all the methods themselves, I've added tests to ensure the still behave as expected.

## `BlockStream`
The refactored version of Historical Backfill, not much has changed here in this version, the main change is that it now used `DeltaLakeClient`. This will eventually be expanded to provide more control over the Stream.

## `indexer_rules_*` & `storage`
Currently these exist as separate crates within the `indexer/` workspace. Rather than creating a workspace for `block-streamer` I've added the respective files for each to the single crate. This is probably fine for `storage`, now called `redis`, but given `indexer_rules_*` is also used in the Registry Contract, I'll probably extract this in a later PR, to avoid it being duplicated.


block-streamer/src/block_stream.rs
```diff
@@ -0,0 +1,418 @@
...
+) -> anyhow::Result<()> {
+    tracing::info!(
+        "Starting block stream at {start_block_height} for indexer: {}",
+        indexer.get_full_name(),
+    );
+
+    let delta_lake_client = crate::delta_lake_client::DeltaLakeClient::new(s3_client.clone());
+
+    // TODO move to DeltaLakeClient
+    let start_date = get_nearest_block_date(&s3_client, start_block_height, chain_id).await?;
```
morgsmccauley, line 110: Previously, this used JSON RPC, but I refactored it to use S3 since we already have access to that and use it elsewhere.

This logic is pretty specific to Delta Lake so I'll probably put this function in there.

block-streamer/src/block_stream.rs
```diff
@@ -0,0 +1,418 @@
...
+        MatchingRule::ActionAny {
+            affected_account_id,
+            ..
+        } => {
+            tracing::debug!(
+                "Fetching block heights starting from {} from delta lake for indexer: {}",
+                start_date.date_naive(),
+                indexer.get_full_name()
+            );
+            // TODO Remove all block heights after start_block_height
```
morgsmccauley, line 125: This is probably a bug - we fetch _all_ block heights within the given day that the specified block height occurs. We should probably remove all block heights which are lower than `start_block_height`.

This plays in well with the refactoring mentioned above, since it can be encapsulated in `DeltaLakeClient`.

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,528 @@
...
+        }
+
+        let objects = self
+            .s3_client
+            .list_all_objects(DELTA_LAKE_BUCKET, prefix)
+            .await?;
+
+        let mut results = vec![];
+        // TODO do in parallel?
+        // TODO only list objects without .json extension
```
morgsmccauley, line 80: We can speed this up by only listing folders/`common_prefixes`. Currently, we list all matching objects, and if that object is a file, it just returns itself.

block-streamer/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+mod indexer_config;
+mod redis;
+mod rules;
+mod s3_client;
+
+pub(crate) const LOG_TARGET: &str = "block_streamer";
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::registry()
```
morgsmccauley, line 18: Starting a random Block Stream just to serve as an integration test. Eventually this will start an endpoint which provides control over the Block Streams.

block-streamer/src/rules/matcher.rs
```diff
@@ -0,0 +1,141 @@
+use near_lake_framework::near_indexer_primitives::{
```
morgsmccauley, line 1: All `rules/` code is the same as what we have currently, just copied over to this new crate.

block-streamer/src/s3_client.rs
```diff
@@ -0,0 +1,134 @@
+const MAX_S3_LIST_REQUESTS: usize = 1000;
+
+#[mockall::automock]
+#[async_trait::async_trait]
+pub trait S3ClientTrait {
```
morgsmccauley, line 5: Using a `trait` allows us to inject mocks in to other code

block-streamer/src/rules/mod.rs
```diff
@@ -0,0 +1,79 @@
+pub mod matcher;
```
morgsmccauley, line 1: Ah, expect this file which differs from the original. `indexer_rules_type` has been merged in here. This will eventually be extracted.

block-streamer/src/block_stream.rs
morgsmccauley: This has been moved to `DeltaLakeClient`

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,720 @@
...
+            block_heights.dedup();
+        }
+
+        tracing::debug!(
+            "Found {} matching block heights matching {}",
+            block_heights.len(),
+            contract_pattern,
+        );
+
+        // TODO Remove all block heights after start_block_height
```
morgsmccauley, line 262: This is probably a bug - we fetch all block heights within the given day that the specified block height occurs. If the block height is finalised in the middle of the day, we unnecessarily include all block heights prior.

We should probably remove all block heights which are lower than `start_block_height`.

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,720 @@
...
+    }
+
+    fn get_lake_bucket(&self) -> String {
+        match self.chain_id {
+            ChainId::Mainnet => "near-lake-data-mainnet".to_string(),
+            ChainId::Testnet => "near-lake-data-testnet".to_string(),
+        }
+    }
+
+    pub async fn get_nearest_block_date(
```
morgsmccauley, line 71: Previously used JSON RPC but has been refactored to use S3 to avoid maintaining multiple clients.

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,720 @@
...
+                        retry_count += 1;
+                        if retry_count > MAX_S3_RETRY_COUNT {
+                            anyhow::bail!("Exceeded maximum retries to fetch block from S3");
+                        }
+
+                        tracing::debug!(
+                            "Block {} not found on S3, attempting to fetch next block",
+                            current_block_height
+                        );
+                        current_block_height += 1;
```
darunrs, line 104: So the retry count here isn't that we're retrying to get the same block but instead retrying to find the closest matching block by trying to get the number 1 over? What scenario would we need to try finding the next block over, up to 20 times?

block-streamer/src/delta_lake_client.rs
darunrs: I'm a tiny bit confused about the application of this function. What kinds of objects are we looking for? By "common_prefixes", do you mean "block." or "shard."? I believe ListObject allows matching the entire prefix against some pattern, using *, ., and ? symbols.

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,720 @@
...
+                if content.is_empty() {
+                    None
+                } else {
+                    serde_json::from_str::<IndexFile>(&content).ok()
+                }
+            })
+            .flat_map(|index_file| index_file.heights)
+            .collect();
+
+        let pattern_has_multiple_contracts = contract_pattern.chars().any(|c| c == ',' || c == '*');
```
darunrs, line 250: I had a feeling we supported multiple contracts since I've seen code related to it, but I never tried it. I don't think the UI let's you put multiple. Is this a functionality we just have some partial support for?

block-streamer/src/delta_lake_client.rs
darunrs: The code comment says all block heights after start. I assume you meant before? Also, is the list of block heights in order? Maybe we can get a slice of the list starting from start_date? Or when we flush to redis, we can filter there.

darunrs (approved):
Lot of code to look through. I believe I understand/see the changes outside of the rule engine, which I need to pour over some more. Some questions, but otherwise looks great! I salute you for your efforts in refactoring coordinator, sir.

block-streamer/src/redis.rs
```diff
@@ -0,0 +1,142 @@
+pub use redis::{self, aio::ConnectionManager, FromRedisValue, ToRedisArgs};
+
+pub const LAKE_BUCKET_PREFIX: &str = "near-lake-data-";
+pub const STREAMS_SET_KEY: &str = "streams";
+
+pub async fn get_redis_client(redis_connection_str: &str) -> redis::Client {
+    redis::Client::open(redis_connection_str).expect("can create redis client")
+}
+
+pub fn generate_real_time_stream_key(prefix: &str) -> String {
```
darunrs, line 10: Are these still used in block-streamer?

block-streamer/src/block_stream.rs
```diff
@@ -0,0 +1,201 @@
...
+        MatchingRule::ActionFunctionCall { .. } => {
+            bail!("ActionFunctionCall matching rule not yet supported for historical processing, function: {:?} {:?}", indexer.account_id, indexer.function_name);
+        }
+        MatchingRule::Event { .. } => {
+            bail!("Event matching rule not yet supported for historical processing, function {:?} {:?}", indexer.account_id, indexer.function_name);
+        }
+    }?;
+
+    tracing::debug!(
+        "Flushing {} block heights from index files to historical Stream for indexer: {}",
```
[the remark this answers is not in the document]
morgsmccauley: Ah, no. We only have a single stream so don't need to distinguish between historical/real-time, I'll update these logs in a future PR

block-streamer/src/delta_lake_client.rs
```diff
@@ -0,0 +1,720 @@
...
+}
+
+impl<T> DeltaLakeClient<T>
+where
+    T: crate::s3_client::S3ClientTrait,
+{
+    pub fn new(s3_client: T) -> Self {
+        DeltaLakeClient {
+            s3_client,
+            // hardcode to mainnet for
```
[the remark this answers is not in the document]
morgsmccauley: "for now" lol - I fixed it in this PR https://github.com/near/queryapi/pull/430/files#diff-bd02e799d38dbea2e9e4c46adfbcc107693ec11fb4bffbed84d442a7191b88bbR50

block-streamer/src/delta_lake_client.rs
morgsmccauley: `common_prefixes` is an S3 concept, It is essentially the 'sub-folders'. `list_all_objects` will return both 'files' and 'folders', and in the original implementation (which has mostly been carried over to here), we would then call list on each file/folder. Listing a file is pointless as it just returns itself, we only need to list folders - hence the comment.

block-streamer/src/delta_lake_client.rs
morgsmccauley: Ah yes, I mean before.

The list of block heights is in order. Yeah, we can slice it from the `start_block_height`, I just wanted to avoid changing the code too much, we can address in a future PR :).

block-streamer/src/delta_lake_client.rs
morgsmccauley: The UI does/should let you input multiple :)

PR #430, opened 2023-11-26
feat: Expose endpoint to control streams

### LINKED ISSUE #420: Expose endpoint to control streams
Coordinator requires a direct control point to the Block Streamer service so that it can control the state of the existing Streams.

An endpoint with the following capabilities should be exposed:
- Start an Block Stream
- Stop an Block Stream
- List existing Block Streams

As this is part of the new Block Streamer service, we don't need to implement a toggle like we are in Runner. The toggle exists at the system level by starting either this service, or Coordinator V1.

morgsmccauley:
This PR exposes a gRPC endpoint for the Block Streamer service to provide control over individual Block Streams. There are currently 3 methods: start/stop/list. The functionality across these methods is quite basic, I didn't want to spend too much time fleshing them out as I'm still not certain on how exactly they will be used. I expect them to Change once the Control Plane starts using them.


block-streamer/build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
morgsmccauley, line 1: Generates the Rust types from the `proto` file on `cargo build`

block-streamer/examples/list_streams.rs
```diff
@@ -0,0 +1,17 @@
+use tonic::Request;
```
morgsmccauley, line 1: These examples use the exposed Rust types to interact with the gRPC service - these not only serve as examples, but also make it easy to quickly debug by running them

block-streamer/src/indexer_config.rs
```diff
@@ -12,15 +15,21 @@ use near_lake_framework::near_indexer_primitives::types::AccountId;
...
     pub account_id: AccountId,
     pub function_name: String,
-    pub code: String,
-    pub start_block_height: Option<u64>,
-    pub schema: Option<String>,
-    pub provisioned: bool,
+    // pub code: String,
+    // pub start_block_height: Option<u64>,
+    // pub schema: Option<String>,
+    // pub provisioned: bool,
```
morgsmccauley, line 21: No longer used/needed - this is more for the Runner side of things which will be passed from the Control Plane directly

block-streamer/src/server/block_streamer_service.rs
```diff
@@ -0,0 +1,182 @@
...
+            })
+            .collect();
+
+        let response = blockstreamer::ListStreamsResponse {
+            streams: block_streams,
+        };
+
+        Ok(Response::new(response))
+    }
+}
```
morgsmccauley, line 182: I should probably add tests here

block-streamer/examples/list_streams.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use block_streamer::block_streamer_client::BlockStreamerClient;
+use block_streamer::{start_stream_request::Rule, ActionAnyRule, ListStreamsRequest, Status};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut client = BlockStreamerClient::connect("http://[::1]:10000").await?;
```
darunrs, line 8: Do we want to create a client for each request? Can we just use one client instead?

block-streamer/src/block_stream.rs
```diff
@@ -0,0 +1,208 @@
...
+
+    let mut last_indexed_block =
+        blocks_from_index
+            .last()
+            .map_or(last_indexed_block, |&last_block_in_index| {
+                // Check for the case where index files are written right after we fetch the last_indexed_block metadata
+                std::cmp::max(last_block_in_index, last_indexed_block)
+            });
+
+    tracing::debug!(
```
darunrs, line 164: I see. So, here is where we'd end up with duplicated reads since each stream has its own lake stream? I can see why we would need to rethink the caching as you had briefly mentioned in the weekly meeting.

block-streamer/src/redis.rs
```diff
@@ -0,0 +1,142 @@
+pub use redis::{self, aio::ConnectionManager, FromRedisValue, ToRedisArgs};
```
darunrs, line 1: I feel like I saw these changes already. I don't know if a rebase is needed or something?

darunrs (approved):

block-streamer/src/redis.rs
morgsmccauley: Ah sorry @darunrs, this PR was pointing to #428 so included the commits from that branch. I thought the commits would be removed after merging but that only works for merge commits, not squash merge.

I've removed the commits from the base PR.

block-streamer/examples/list_streams.rs
morgsmccauley: This is a cargo example, a standalone binary which can be executed via `cargo --example list_streams`. I created it so we can easily send requests to the server, rather than creating verbose cli commands.

block-streamer/src/block_stream.rs
morgsmccauley: Yep exactly - we'd need to create some shared client so it can cache internally, but that's a future problem :)

block-streamer/src/server/block_streamer_service.rs
morgsmccauley: https://github.com/near/queryapi/pull/442

PR #438, opened 2023-11-30
fix: Fix bugs in stream message handling and minor logging improvements

darunrs:
It was observed that there are three related bugs in stream message processing: 

1. In some cases, the stream will increment the sequence and not the main number, causing incrementId to skip messages. 
2. The current stream ID is used when calling xRange, giving an inaccurate count of messages if any skips occur. 
3. Unprocessed messages are no longer reported when the queue of blocks ie empty. This occasionally led to a non-zero message count being scraped and then left that way even when the stream was in fact zero. 

In addition, xRange takes a significant amount of time to run depending on the size of the queue. This impacts all other operations since redis is single-threaded. xLen takes much less time to call while also accurately returning the count of messages in the stream. 

To resolve the bugs, I now increment the sequence instead of the main ID. In addition, if no more stream messages are being fetched, the stream message ID is reset to '0' to ensure new messages are collected regardless of their ID. Finally, I replaced xRange with xLen. I also changed a blocking while loop to an if statement in consumer so that if the queue is empty, it continues, which triggers the finally. 

I made some small additions to logging statements to include the indexer type and block number for success/failure to help diagnose problems with indexers in the future.

### DIFF
--- a/runner/src/stream-handler/worker.ts
+++ b/runner/src/stream-handler/worker.ts
@@ -132,8 +132,8 @@ async function blockQueueConsumer (workerContext: WorkerContext, streamKey: stri
       await sleep(10000);
       console.log(`Failed: ${indexerName} ${workerContext.streamType} on block ${currBlockHeight}`, err);
     } finally {
-      const unprocessedMessages = await workerContext.redisClient.getUnprocessedStreamMessages(streamKey);
-      METRICS.UNPROCESSED_STREAM_MESSAGES.labels({ indexer: indexerName, type: workerContext.streamType }).set(unprocessedMessages);
+      const unprocessedMessageCount = await workerContext.redisClient.getUnprocessedStreamMessageCount(streamKey);
+      METRICS.UNPROCESSED_STREAM_MESSAGES.labels({ indexer: indexerName, type: workerContext.streamType }).set(unprocessedMessageCount);
 
       parentPort?.postMessage(await promClient.register.getMetricsAsJSON());
     }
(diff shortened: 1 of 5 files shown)


runner/src/stream-handler/worker.ts
```diff
@@ -121,15 +125,15 @@ async function blockQueueConsumer (workerContext: WorkerContext, streamKey: stri
...
+      console.log(`Success: ${indexerName} ${workerContext.streamType} on block ${currBlockHeight}}`);
     } catch (err) {
       await sleep(10000);
-      console.log(`Failed: ${indexerName}`, err);
+      console.log(`Failed: ${indexerName} ${workerContext.streamType} on block ${currBlockHeight}`, err);
     } finally {
-      const unprocessedMessages = await workerContext.redisClient.getUnprocessedStreamMessages(streamKey, streamMessageId);
-      METRICS.UNPROCESSED_STREAM_MESSAGES.labels({ indexer: indexerName, type: workerContext.streamType }).set(unprocessedMessages?.length ?? 0);
+      const unprocessedMessages = await workerContext.redisClient.getUnprocessedStreamMessages(streamKey);
+      METRICS.UNPROCESSED_STREAM_MESSAGES.labels({ indexer: indexerName, type: workerContext.streamType }).set(unprocessedMessages);
```
gabehamilton, line 136: Is this supposed to be the messages and not the length?

darunrs: Yes, its the count of messages now. Previously, we were actually fetching every single message and then just checking the length of the array. I'll rename the variable to unprocessedMessageCount.

gabehamilton (approved):

PR #442, opened 2023-12-11
test: Make `block-streamer` unit-testable and add tests

morgsmccauley:
This PR is a combination of refactoring + testing, the former enabling the latter

## Mock `impl` instead of `trait`
Originally, `trait`s such as `S3ClientTrait` were used to abstract the behaviour of I/O related `struct`s, allowing for the injection of mock implementations via `automock`. This was fine, but lead to very verbose syntax which needed to be propagated throughout the codebase where ever it was used.

Now, the `impl` itself is mocked, and either the mock or real implementation is returned dependant on the `cfg(test)` flag. This means we no longer need to worry about generics and can instead use the `struct` itself, as we would with non-mocked code.

The trade-off here is that rust now thinks the 'real' code is dead/unused as `test` is enabled by default. 

## Mocking `near-lake-framework`
`aws` exposes an HTTP client which can be configured with mock request/response values. `aws-sdk-s3` can be configured to use this client, creating a mock client. Injecting this config into `near-lake-framework` allows us to create deterministic outputs, enabling unit-testing.

`aws_config` is now taken as a parameter where relevant so that we can mock `near-lake-framework` in out tests. The functions for achieving this have been placed in `test_utils` so they can be reused.

## Test `BlockStreamerService` and `BlockStream`
With `RedisClient`, `S3Client`, and `Lake` now mockable - tests have been added for these `struct`s. I've also added a Github action which runs on every PR. All raw block data has been added to `data/`, which makes up the majority of the diff in this PR.


block-streamer/Cargo.toml
```diff
@@ -21,13 +18,18 @@ serde = { version = "1", features = ["derive"] }
...
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
-tokio = { version = "1.28.0", features = ["rt-multi-thread"]}
+tokio = { version = "1.28.0", features = ["full"]}
 tokio-util = "0.7.10"
 tokio-stream = "0.1.14"
 tonic = "0.10.2"
 wildmatch = "2.1.1"
 
-near-lake-framework = "0.7.1"
+near-lake-framework = { git = "https://github.com/near/near-lake-framework-rs", branch="chore/upgrade-aws-0.7.x"}
```
morgsmccauley, line 27: Will pin to the specific version once https://github.com/near/near-lake-framework-rs/pull/94 is merged

darunrs (approved):
Really cool stuff!

PR #446, opened 2023-12-13
feat: Runner gRPC endpoint

### LINKED ISSUE #416: Expose endpoint to control executors
Coordinator requires a direct control point to Runner so that it can control the state of the existing Executors.

An endpoint with the following capabilities should be exposed:
- Start an Executor
- Stop an Executor
- List existing Executors

To enable a seamless transition to the Control Plane, we should be able to toggle between endpoint based control, and the existing solution in which Runner starts Executors on its own.

darunrs:
As part of implementing a new control plane to oversee QueryApi resources, Runner needed endpoints with which control plane can send commands. This PR adds code to create a new gRPC server with endpoints to start, stop, or list Runner executors. This provides the control plane the ability to fully control Runner, and removes the need for bilateral decision making, which was a problem of the previous design. 

The changes will be changed further as needs become more concrete, closer to integration/release.


darunrs:
Need to figure out that snyk error. I don't currently have access to that run.

runner/protos/runner.proto
```diff
@@ -0,0 +1,66 @@
...
+
+    // Stops an existing Runner stream worker
+    rpc StopStream (StopStreamRequest) returns (StopStreamResponse);
+
+    // Lists all Runner stream workers
+    rpc ListStreams (ListStreamsRequest) returns (ListStreamsResponse);
+}
+
+// Start Stream Request 
+message StartStreamRequest {
```
morgsmccauley, line 19: We're working with "runners" here rather than "streams", I guess this was something you forgot to update after copy/paste?

I've historically used "executors" as the term for the thing that executes indexer code. Maybe we could use that instead? what do you think?

runner/src/generated/runner.ts
```diff
@@ -0,0 +1,24 @@
+import type * as grpc from '@grpc/grpc-js';
```
morgsmccauley, line 1: These types bloat the repo/PR, should we just remove/ignore them given that they can be deterministically generated locally, similar to `node_modules`.

runner/src/service/runner-service.ts
```diff
@@ -0,0 +1,241 @@
...
+import type StreamHandler from '../stream-handler';
+import * as grpc from '@grpc/grpc-js';
+import assert from 'assert';
+
+type StreamHandlers = Map<string, StreamHandler>;
+
+function getRunnerService (StreamHandlerType: typeof StreamHandler): RunnerHandlers {
+  const streamHandlers: StreamHandlers = new Map();
+
+  const RunnerService: RunnerHandlers = {
```
morgsmccauley, line 21: I think we can use a class here, we could use that to encapsulate all related methods, that might be cleaner?:
```js
class RunnerService implements RunnerHandlers {
  ListStreams () {
  }

  StartStream () {
  }

  StopStream () {
  }
}
```

runner/protos/runner.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
+package spec;
+
+service Runner {
+    // Starts a new Runner stream worker
+    rpc StartStream (StartStreamRequest) returns (StartStreamResponse);
+
+    // Updates an existing Runner stream worker
+    rpc UpdateStream (UpdateStreamRequest) returns (UpdateStreamResponse);
```
morgsmccauley, line 9: If it's not a big change - let's remove this for now? It may just provide more complexity on the Coordinator side.

morgsmccauley (commented):
Have only skimmed the PR and have noted a few high level things. 

Couple more questions:
- How do we switch between the new/old version
- Can you add tests?

runner/protos/runner.proto
darunrs: Ahhh haha. I think I might have recognized that the naming felt wonky internally but just pushed forward. I didn't like startRunner since we call the overall service Runner, but maybe we can directly call it Indexer? Like startIndexer, stopIndexer. Feels like it's quite direct and I can't foresee any scenario where a particular worker thread is doing more things alongside running an indexer.

runner/src/generated/runner.ts
darunrs: You're absolutely correct. I'll remove these from the repo. Is my approach good regarding building of the types? I basically created a new target called "codegen" which can be run the same way we do build or test. And I modified the Dockerfile to make a codegen call before running build.

runner/protos/runner.proto
morgsmccauley: IMO "indexer" would be both the "block stream" and the "executor". The execution of the code alone doesn't make up the whole picture.

runner/protos/runner.proto
darunrs: As in remove the update API? I felt like it was necessary since one of the things I like about the current system after our improvements was that if your indexer is broken, and you submit a new indexer version, runner catches it really quickly. You can almost immediately see your change reflected in the logs tab. I wanted to make sure we retain any positive customer experience behavior. If we don't have an update API, then we need another way to quickly update Runner. I would say stopping and restarting the Indexer thread would take much longer and possibly prone to issues instead of just telling worker to update its config. 

Those were my thoughts surrounding including it anyway. Let me know what you think.

darunrs:
I added tests but its under the "Load Diff". I think its hidden by default since its a bit large.

runner/protos/runner.proto
darunrs: Yeah that's fair. I agree with that. I think Executor works in that case.

darunrs:
Also, regarding transitioning between old and new, I'm still working on ideas for that. Do you think I should include enabling the server in this PR or not? I felt like this was already a lot and maybe I can just release the implementation and tests, and separately enable the server with a good old to new transition logic in it. What do you think?

morgsmccauley:
> Also, regarding transitioning between old and new, I'm still working on ideas for that. Do you think I should include enabling the server in this PR or not? I felt like this was already a lot and maybe I can just release the implementation and tests, and separately enable the server with a good old to new transition logic in it. What do you think?

Let's do it in a follow up PR to keep the work scoped. We could just create two entrypoints, and then have the main `index.js` file use either entrypoint based on an env flag?

runner/src/service/runner-service.ts
darunrs: I am reasonably certain that a class implementation would require a lot of annoying reworking. My main problems come from the fact that the RunnerHandlers interface extends an interface which has an index signature in it, forcing all elements (Including class variables) to adhere to the format "[name: string]: UntypedHandleCall". I think returning an object that implements the interface is a better approach. Or, I can implement a function which returns a class containing the implementation, but at that point, I think they are indistinguishable?

runner/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM node:18.18.2 AS builder
 WORKDIR /usr/src/app
 COPY . .
 RUN npm install
+RUN npm run codegen
```
morgsmccauley, line 5: No longer needed as its called in `npm build` right?

runner/package.json
```diff
@@ -7,7 +7,8 @@
     "node": "18.17"
   },
   "scripts": {
-    "build": "rm -rf ./dist && tsc",
+    "build": "rm -rf ./dist && npm run codegen && tsc",
+    "codegen": "npx proto-loader-gen-types --longs=String --enums=String --defaults --oneofs --grpcLib=@grpc/grpc-js --outDir=src/generated/ protos/runner.proto",
```
morgsmccauley, line 11:
```suggestion
    "codegen": "proto-loader-gen-types --longs=String --enums=String --defaults --oneofs --grpcLib=@grpc/grpc-js --outDir=src/generated/ protos/runner.proto",
```
`npx` is only needed if you call this from a standalone cli. `npm` scripts will automatically have access to `node_modules` `bin`

runner/protos/runner.proto
```diff
@@ -0,0 +1,66 @@
...
+
+    // Lists all Runner stream workers
+    rpc ListExecutors (ListExecutorsRequest) returns (ListExecutorsResponse);
+}
+
+// Start Executor Request 
+message StartExecutorRequest {
+    string stream_id = 1;
+    string redis_stream = 2;
+    string indexer_config = 3; // JSON containing code, schema, etc.
```
morgsmccauley, line 22: We should create a specific type for this, like i have with `ActionAnyRule` etc. That way specific types can be generated/enforced.

morgsmccauley: Actually, we only need `code`, `schema`, `function_name`, `account_id`? Maybe we should just flatten it and include those instead?

runner/protos/runner.proto
```diff
@@ -0,0 +1,66 @@
...
+    string stream_id = 1;
+}
+
+// Update Executor Request 
+message UpdateExecutorRequest {
+    string stream_id = 1;
+    string indexer_config = 2; // JSON containing code, schema, etc.
+}
+
+// Update Executor Response
```
morgsmccauley, line 36: Still a lot of "stream" references here 😅

runner/src/service/runner-service.ts
```diff
@@ -0,0 +1,241 @@
...
+      return grpcError;
+    }
+  } catch (error) {
+    grpcError.message = 'Invalid indexer config. It must be a valid JSON string.';
+    return grpcError;
+  }
+  return null;
+}
+
+function validateStartExecutorRequest (request: StartExecutorRequest__Output, streamHandlers: StreamHandlers): any | null {
```
morgsmccauley, line 164: Why do we need this manual validation? Shouldn't this be done by grpc and/or typescript?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -3,16 +3,27 @@ import { Worker, isMainThread } from 'worker_threads';
...
+}
+
 export default class StreamHandler {
   private readonly worker: Worker;
+  readonly indexerName: string;
 
   constructor (
-    public readonly streamKey: string
+    public readonly streamKey: string,
+    public readonly indexerConfig: IndexerConfig | undefined = undefined
```
morgsmccauley, line 19: Feels like we should always pass in config, why make this `undefined`?

morgsmccauley: Ah I see, because the original version defines it itself

runner/src/service/runner-client.ts
```diff
@@ -0,0 +1,12 @@
+import * as grpc from '@grpc/grpc-js';
```
morgsmccauley, line 1: I'm not sure this is needed - the client will be Rust

runner/src/service/index.ts
```diff
@@ -0,0 +1 @@
+export { default } from './runner-server';
```
morgsmccauley, line 1: Nit: `service/` seems like the wrong abstraction here, feels like it should be `server/` instead. A GRPC "server" is composed of "services", so we'd expose the `server` and house the `service`s within it.

runner/protos/runner.proto
morgsmccauley: Do you know how long it takes to stop/start an executor? I can't imagine it would be very long.

Stop/start feels cleaner tbh as we're starting from a clean slate, wiping anything which is not needed and/or may cause conflicts. Also from the Coordinator side the control flow is simpler, we don't need to distinguish between update and stop/start - under what conditions should we do one over the other? We always just wipe it and restart.

May be over thinking things here 😓

darunrs: Gotcha, I think that makes sense. I'll quickly check stop/start time and if it's not too problematic, I'll remove the update API.

runner/Dockerfile
darunrs: Yep! Will remove.

runner/protos/runner.proto
darunrs: Oh blast it. haha. Well let's see if I can knock em all out in this second go around

runner/protos/runner.proto
darunrs: Well that's true. For now, we only need those four. I can flatten this. This also gives us typing.

runner/src/service/index.ts
darunrs: True! Will change name.

runner/src/service/runner-client.ts
darunrs: It was a useful thing for integration testing locally. I can remove it from the PR for now and have it as an ignored file locally.

morgsmccauley: By integration testing do you mean running it locally from CLI, or within tests? The Rust client can still help with the former, but if you need the later we can keep this round?

runner/protos/runner.proto
darunrs: It takes around 800-1000ms to start an indexer executor (from making the server call to there being a message in the prefetch queue). Realistically, it's longer since the S3 fetch is unbuffered, so could take another 100-2000ms on average to get the actual first block. But that's really variable. I think in the end the impact is probably invisible right now. We can consider update over start/stop in the future as a further optimization approach, but yeah I agree that for the launch, start/stop is good enough and simple.

morgsmccauley: Thanks 👍🏼

runner/src/service/runner-client.ts
darunrs: I would certainly like to keep it around for now. I'll drop a TODO in it to remove if we no longer need it (e.g. we get a rust client going).

runner/src/service/runner-service.ts
darunrs: I thought the same but when I made the client and tested just calling one of the endpoints with missing fields, it went through. The callback I got was the error I had coded myself.

runner/package.json
```diff
@@ -7,11 +7,12 @@
...
   },
   "scripts": {
-    "build": "rm -rf ./dist && tsc",
+    "build": "rm -rf ./dist && npm run codegen && tsc",
+    "codegen": "rm -rf ./src/generated && proto-loader-gen-types --longs=String --enums=String --defaults --oneofs --grpcLib=@grpc/grpc-js --outDir=src/generated/ protos/runner.proto",
     "start": "npm run build && node dist/index.js",
     "start:dev": "ts-node ./src/index.ts",
     "start:docker": "node dist/index.js",
-    "test": "node --experimental-vm-modules ./node_modules/.bin/jest",
+    "test": "npm run codegen && node --experimental-vm-modules ./node_modules/.bin/jest",
```
morgsmccauley, line 15:
```suggestion
    "test": "npm run build && node --experimental-vm-modules ./node_modules/.bin/jest",
```
Lets do `build` instead as there will most likely be other things within build which would be required?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -23,6 +34,16 @@ export default class StreamHandler {
     }
   }
 
+  async stop (): Promise<void> {
+    await this.worker.terminate();
+  }
+
+  updateIndexerConfig (indexerConfig: IndexerConfig): void {
```
morgsmccauley, line 41: Not needed right?

runner/src/stream-handler/worker.ts
```diff
@@ -26,8 +27,18 @@ interface WorkerContext {
 
 const sleep = async (ms: number): Promise<void> => { await new Promise((resolve) => setTimeout(resolve, ms)); };
 
+let config: IndexerConfig | undefined;
+
+parentPort?.on('message', (message) => {
```
morgsmccauley, line 32: Not needed?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -3,16 +3,27 @@ import { Worker, isMainThread } from 'worker_threads';
...
 export default class StreamHandler {
   private readonly worker: Worker;
+  readonly indexerName: string;
 
   constructor (
-    public readonly streamKey: string
+    public readonly streamKey: string,
+    public readonly indexerConfig: IndexerConfig | undefined = undefined
   ) {
+    this.indexerName = (indexerConfig?.account_id ?? 'undefined_account') + '/' + (indexerConfig?.function_name ?? 'undefined_function');
```
morgsmccauley, line 21: Seems more intuitive to just expose `accountId` and `function_name`, as the client will probably want to compare against those directly. Exposing this concatenated string would require parsing on the client side.

morgsmccauley (approved):
Great job on this 👏🏼 

Mostly looks good, left a few more minor comments. The validation still doesn't feel right, I feel like that should be implemented out of the box, can you spend some more time investigating please 🙏🏼 if we end up needing it feel free to merge 😄

runner/package.json
darunrs: Makes sense. Sure!

runner/src/stream-handler/stream-handler.ts
darunrs: Ah yep. Not needed.

runner/src/stream-handler/worker.ts
darunrs: Also not needed. Will remove.

runner/src/stream-handler/stream-handler.ts
darunrs: That makes sense. I'll expose the two separately then!

darunrs:
I'll investigate validation some more! I agree it looks awkward.

darunrs:
There are no required fields in Protocol Buffers. Their docs state you can have an optional, repeated, or map type. I wasn't able to find any resources that indicate it's available in proto3. Apparently proto2 had it though. I'll simplify my code on the idea that the fields will have a default value, at least though.

darunrs:
I think running build before test in that way causes some problems in github. Most likely its the deleting of the dist folder, since that's where the snapshots are saved. Not really sure since tsc should regenerate them. But in any case, I have to keep it as codegen && test instead of build (codegen inside) && test.

PR #448, opened 2023-12-14
fix: Case Sensitive Quoted Column Names Support for Context.DB

darunrs:

INSERT INTO nearpavel_near_staking_v3."receipts" (blockHeight, blockHash, blockTimestamp, blockDate, depositIn, receiptId, predecessorId, signerId, validatorId) VALUES ('106896621', '8o26GZfSt4z7TrbHdHhDCXttXkG3hUFNkLCL33y52c1d', '1701388833845.7158', '2023-12-01', '900000000000000000000000', 'AZhuyEwThxEHUyUcudiNJ49pBhMekgjgokj9tmfHEgkg', '2ee9129228b532b25181ec6cbd1e09bf058e5169c9a9606fec69b5f635423d97', '2ee9129228b532b25181ec6cbd1e09bf058e5169c9a9606fec69b5f635423d97', 'astro-stakers.poolv1.near') ON CONFLICT (receiptId) DO UPDATE SET depositIn = excluded.depositIn RETURNING *

column "blockheight" of relation "receipts" does not exist

The fix is to quote the column names in the query, if they were quoted in the schema. 

In addition, the logs for context.db were being written under the wrong function name. I fixed that too.


darunrs:
Needs some more integration testing with postgres to ensure there aren't any exisitng cases broken. So far I ran the staking v3 indexer with fixed code and verified that worked and inserted into the database. 

Need to: 

- Test social feed
- Test all methods
- Test with case sensitive quoted columns in schema

Open to more test ideas.

darunrs:
Turns out that the reserve of this problem is also problematic. If the schema uses blockHeight, and the query contains "blockHeight", that will also not work. So, I actually need to only quote it if the schema itself quoted the table name. This is a problem because the current way I get tableNames strips the quotes, if they were present.

darunrs:
Will create new PR as approach was entirely reworked.

PR #449, opened 2023-12-14
Fix type gen bug

### LINKED ISSUE #452: Error Handling is inconsistent for code + schema + type generation flows
**Describe the bug**
often times, when the UI is loaded with incorrectly checked in code, we throw an error instantly. Not giving the developer the opportunity to fix their mistake and check in the correct code. 

Also, the type generation error is using the same state variable to keep track of its errors as is the code.js and schema.sql files. We should differentiate these as its getting harder to keep track of what is currently in error state.

**To Reproduce**
1. Go to `https://near.org/dev-queryapi.dataplatform.near/widget/QueryApi.App?selectedIndexerPath=dataplatform.near/social_feed`
2. The indexer can editor window is not thrown. No error in the console window either.

roshaans:
We have an issue where if the formatting of SQL is not current or we encounter a case in which we do not support the type generation library for the context method, we fail very abruptly. 

This PR handles these errors better. 

- If sql/code encounters syntax error and formatting fails, we simply return the unformatted code to the editor so the user can fix it. 
- If code + sql both have syntax errors, the user will now see two different alerts. One for each one of them. 
- We throw the error, but we also log everything to console for debugging purposes. 
- We catch and throw type generation errors in a new variable.

### DIFF
--- a/frontend/src/components/Editor/Editor.js
+++ b/frontend/src/components/Editor/Editor.js
@@ -121,10 +121,10 @@ const Editor = ({
     if (fileName === "indexingLogic.js") {
       try {
         setSchemaTypes(pgSchemaTypeGen.generateTypes(schema));
-        setTypesGenerationError(() => undefined);
+        setTypesGenerationError(undefined);
       } catch (typesError) {
         console.log("typesGenerationError generating types for saved schema.\n", typesError);
-        setTypesGenerationError(() => "There was an error generating types from your schema. Please fix your schema or file a support ticket.");
+        setTypesGenerationError("There was an error generating types from your schema. Please fix your schema or file a support ticket.");
       }
     }
   }, [fileName]);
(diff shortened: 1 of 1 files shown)


frontend/src/components/Editor/Editor.js
```diff
@@ -127,9 +121,10 @@ const Editor = ({
     if (fileName === "indexingLogic.js") {
       try {
         setSchemaTypes(pgSchemaTypeGen.generateTypes(schema));
-        setError(() => undefined);
-      } catch (error) {
-        handleCodeGenError(error);
+        setTypesGenerationError(() => undefined);
```
morgsmccauley, line 124:
```suggestion
        setTypesGenerationError(undefined);
```
Using a function is only necessary when you intend to use the existing state, i.e. `(currentState) => currentState + 1`

frontend/src/components/Editor/Editor.js
```diff
@@ -268,64 +249,40 @@ const Editor = ({
...
 
-  const handleCodeGenError = (error) => {
-    console.error("Error generating types for saved schema.\n", error);
-    const errorMessage = "Oh snap! We could not generate types for your SQL schema. Make sure it is proper SQL DDL."
-    setError(() => errorMessage);
-  };
-
   async function handleFormating() {
-    await reformat(indexingCode, schema);
+    await reformatAll(indexingCode, schema);
```
morgsmccauley, line 285:
```suggestion
    reformatAll(indexingCode, schema);
```
This function is not `async`

morgsmccauley (approved):

frontend/src/components/Editor/Editor.js
```diff
@@ -127,9 +121,10 @@ const Editor = ({
     if (fileName === "indexingLogic.js") {
       try {
         setSchemaTypes(pgSchemaTypeGen.generateTypes(schema));
-        setError(() => undefined);
-      } catch (error) {
-        handleCodeGenError(error);
+        setTypesGenerationError(() => undefined);
+      } catch (typesError) {
+        console.log("typesGenerationError generating types for saved schema.\n", typesError);
```
darunrs, line 126: Should this be a console.error? Same for the others.

darunrs (approved):
FYI, I tried to run the frontend locally again, since it's been a while. But, it wasn't loading for some reason. I did: 
1. yarn serve:widgets:local
2. Set flag to http://127.0.0.1:3030
3. Ran yarn dev in a separate terminal
4. Visited localhost:3000. 

I was stuck at the loading screen.

PR #453, opened 2023-12-14
refactor: Extract registry types in to own crate

### LINKED ISSUE #421: Create Control service

  2. Fetch Executor statuses
  3. Fetch Streamer statuses
  4. For each indexer
      - Provisioning
      - Streams
          - else: do nothing
      - Executors
          - else: do nothing

morgsmccauley:
This PR extract the types exposed by the registry contract in to its own reusable crate: `registry_types`. This was done to avoid the duplication across our different service which; fetch the JSON, and then construct the Rust types.

## `registry_types` crate 
The registry types were already partially shared between Coordinator and the registry, but I've expanded on this to include all types rather than just a subset. I've created a new crate, rather than using the existing `./indexer/indexer_rule_type` crate as some packages needed to be upgraded, which caused conflicts across the `indexer/` cargo workspace - which seemed unnecessary to fix given that it is being deprecated.

## Registry contract updates
`near-sdk` version `4.1.1` is no longer compilable because it depends on a non-existent crate, hence the upgrade to `5.0.0-alpha.1`. Since I was already making changes here, I decided to remove the migration related code as it is no longer needed.

## Consuming `registry_types`
With the shared types in place, I've updated `block-streamer/` to use them, removing the existing types which it duplicates. I also went ahead and removed all types which were unused (left over from Alertexer).

`indexer_rule_type` depends on `near-sdk` version `4.1.1` (the broken one), and since it's no longer shared, I removed that dependency so that it can compile again.

Relates to https://github.com/near/queryapi/issues/421


registry/contract/Cargo.toml
```diff
@@ -8,9 +8,10 @@ edition = "2021"
 crate-type = ["cdylib"]
 
 [dependencies]
-near-sdk = "4.1.1"
+borsh = "1.0.0"
+near-sdk = "5.0.0-alpha.1"
 uint = { version = "0.9.3", default-features = false }
-indexer_rule_type = { features = ["nearsdk"], path = "../../indexer/indexer_rule_type" }
+registry-types = { path = "../types", features = ["nearsdk"] }
```
roshaans, line 14: are we using the `nearsdk` feature?

roshaans:
Can we add some details in the main readme about this `readme`?

morgsmccauley:
> Can we add some details in the main readme about this `readme`?

The types are used internally, I'm not sure adding something would provide much benefit? Did you have something specific in mind?

registry/contract/Cargo.toml
morgsmccauley: Yes - it forces the crate to use `near-sdk` types so that it is compatible with the registry contract

darunrs (approved):

roshaans (approved):

PR #458, opened 2023-12-19
feat: Add created/updated at block heights to registry

### LINKED ISSUE #415: Add created/updated at timestamps
An Executor, or Block Streamer, will store the version it is currently running. This will be compared with the latest version in the Registry contract, and when it differs, appropriate commands will be issued.

To enable the above, we should add the `created_at_block_height` and `updated_at_block_height` fields to the Registry contract entries.

morgsmccauley:
This PR adds the `created_at_block_height` and `updated_at_block_height` fields to `IndexerConfig` within the registry contract. The motive behind this is to provide Coordinator V2 with a way for comparing the actual and desired states of the system, i.e. if there is a mismatch between the registry and the system, action should be taken. Without versions, there is no way of making this comparison.

## Compilation Errors
~~I ran in to several issues trying to compile the `wasm32` binary, and have outlined all these issues in https://github.com/near/near-sdk-rs/issues/1125, as well as in the `README.md` so that the fixes are documented. These fixes are a bit janky, but I've tested the deployed contract and all seems to be ok.~~

These have been resolved, see: https://github.com/near/queryapi/pull/458#issuecomment-1874977342

## Account Roles Migration
I've also included `account_roles` in this migration as we have some incorrect accounts as `Owner`s (`pavelnear.near`). All owners will be wiped and re-written from the contract default state. All `User`s will remain.

## Coordinator V1
Coordinator V1 has been tested to ensure that it can still parse the registry after these new fields have been applied.


morgsmccauley:
@roshaans will these new fields create any problems on the frontend side? I'd expect not because it is an expanding change, rather than removing/updating existing fields?

gabehamilton (approved):

morgsmccauley:
The compilation errors originally mentioned in the PR description have now been resolved.

These errors were related to the compilation of `near-primitives` to `wasm32`, the smart contract target. This dependency was included as a `default` feature for the shared `registry-types` crate, meaning it would be included in the compilation. It has since been removed and now must be explicitly listed as a feature, hence the change to `block-streamer`.

roshaans (approved):
Able to build properly! These changes also do not seem like they would affect the frontend in anyway so LGTM 💯

morgsmccauley:
Deployment status:
- Dev ✅ 
- Prod 🚧

PR #479, opened 2024-01-03
feat: Create Rust gRPC client for Runner

darunrs:
A Rust gRPC Client for Runner is necessary to allow Coordinator V2 to control executors. Coordinator V2 will be responsible for keeping executors up to date and handle failures in executors (e.g. worker crashes). 

Coordinator V2 will go in another folder. Therefore, most of the code here needs to be migrated once Coordinator V2 is ready.


morgsmccauley (approved):

darunrs:
Had offline discussion with Morgan about the below: 

Hey Morgan. Something I noticed just now while thinking over the rust client. I remembered that the docker file structure is a bit different than local so I tried to docker compose build and it actually broke on two fronts: protoc not installed and proto not found. The protoc part is an easy fix: update the Dockerfile. The proto not found I think is a bit tougher. I'd need to copy it into the docker, but that's not doable since the build context is indexer/, which doesn't include runner. So, I have these options for proceeding, I believe:
Don't copy anything, add a .ok() to forcibly suppress the tonic_build errors (if any), and figure out the file later. (What is currently in [PR](https://github.com/near/queryapi/pull/479) now)
Expand build context for coordinator to include runner, copy the file in, and adjust Dockerfile accordingly.
Copy the file directly into coordinator for now and implement a better solution later.

As a result of the discussions, we agreed that the amoutn of code that depends on his https://github.com/near/queryapi/pull/444 PR is too high. So, he can integrate this work into his PR instead and resolve any integration problems that might exist. I will test out approach 2 today since we know 1+3 theoretically works already.

morgsmccauley:
Duplicates https://github.com/near/queryapi/pull/491

PR #480, opened 2024-01-03
Feat/editor error logging

jl-santana:
#### Changes Made:
- Implemented a function to validate types and format from SQL schema, enhancing the schema editor's error logging capabilities.
- Reduced error states and removed unnecessary Alerts for improved code clarity.
- Added debounce functionality to schema validation for enhanced performance.
- Created a dedicated __tests__ folder to improve project folder structure.

https://github.com/near/queryapi/assets/15988846/0030c4c9-71af-4c2c-9736-9f07360581bd


roshaans (commented):
Thanks for submitting this PR @jl-santana 

Nice addition on the debounce logic! 

We recently had one state variable to keep track of our Editor error states and that caused a few issues: 
1. If there are issues in multiple files (indexerFunction.js, schema, or type generation from schema), the user would only see one of these errors. So we made them explicit in order to show the full exhaustive list of errors. You can check out his [PR here](https://github.com/near/queryapi/pull/449) for more details or this [issue](https://github.com/near/queryapi/issues/452)

I think we should leave these as exhaustive even if they seem redundant as without it, its really difficult to track down which of the following is causing the issue (code, schema, typegeneration).

Open to hear about other suggestions on how to improve this!

jl-santana:
@roshaans and I synced offline, tested the new flow, and found that using only one error state suffices for user communication. Additionally, the new flow prevents indexer registration in case of a schema error. I'll further test it on dev before deploying to prod.

gabehamilton:
> Additionally, the new flow prevents indexer registration in case of a schema error. 

That sounds worrisome since our client side schema validation often hasn't supported valid SQL features that we need in the schema (for example, `DEFAULT ''`)

jl-santana:
@gabehamilton valid point, can we meet about this?

jl-santana:
@gabehamilton, @darunrs  and I synced offline to discuss this issue. 
- We decided to block the user from register a function if there is a formatting error on the schema, in the case of a type generation error, we will let the user know that a contextDB object wont be generated but he can choose whether to proceed

PR #488, opened 2024-01-04
feat: Toggle Runner Version

### LINKED ISSUE #478: Enable Toggle between Endpoint and Redis Control of Executors
Runner needs to be capable of switching between using endpoint based control of runners, and using the existing redis polling method. 

Potential options below.

darunrs:
Runner will need the ability to start using V1 (Polling Redis to start streams) or V2 (Starting streams upon receiving calls to gRPC server) depending on an environment variable we set before deployment. This way, we can make the migration from V1 to V2 and roll that back without too much difficulty. 

In order to do this, the main thread needs to have a separate map of stream handlers owned by the grpc server. The server needs to be spun up only in V2 mode. And Redis polling should only happen in V1 mode. In addition, the worker thread needs to only poll Redis stream storage for indexer config in V1 mode. To do so, only V2 will pass in the config into the worker upon executor init. This provides version control over indexer config as well. 

Finally, I addressed some remaining TODOs leftover. Mainly returning more information, tracking status of executors, and returning all. information available when listing executors. 

For now, I've defaulted the version to V1, if the env variable is not set.

### DIFF
--- a/runner/package-lock.json
+++ b/runner/package-lock.json
@@ -23,7 +23,10 @@
         "prom-client": "^14.2.0",
         "redis": "^4.6.7",
         "verror": "^1.10.1",
-        "vm2": "^3.9.19"
+        "vm2": "^3.9.19",
+        "zipkin": "^0.22.0",
+        "zipkin-context-cls": "^0.22.0",
+        "zipkin-transport-http": "^0.22.0"
       },
       "devDependencies": {
         "@types/express": "^4.17.17",
@@ -34,6 +37,7 @@
         "@types/pg-format": "^1.0.2",
         "@types/pluralize": "^0.0.29",
         "@types/verror": "^1.10.6",
+        "@types/zipkin-context-cls": "^0.11.7",
         "@typescript-eslint/eslint-plugin": "^5.62.0",
         "@typescript-eslint/parser": "^5.62.0",
         "eslint": "^8.46.0",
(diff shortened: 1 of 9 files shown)


darunrs:
Some tests I ran manually: 

Unset RUNNER_VERSION and verify Runner starts in V1. 
Run `export RUNNER_VERSION=V1` and verify V1 started. Tried calling server and received connection refused. 
Run `export RUNNER_VERSION=V2` and verify V2 started. Then, make the following calls:
- list executors and verify empty list returned
- start one executorA which runs
- list executors and verify it shows up with running status
- stop the executorA
- list again and verify no executors show up
- start executorA again
- start another executorB with a hard coded flag which makes worker crash if the account matches this indexer
- list executors immediately after executorB and verify both executors show up with running status
- verify executorA continues to run while executorB stops running
- list executors again and verify A has running status while B has stopped status
- stop executorB
- list executors and verify executorA is only one in list

runner/src/index.ts
```diff
@@ -16,24 +18,38 @@ void (async function main () {
...
+      console.log('Starting Runner in V2 mode.');
+      runnerServer = startRunnerServer();
+    } else if (version === 'V1') {
+      console.log('Starting Runner in V1 mode.');
+    } else {
+      console.error('Unknown version', version);
+      process.exit(1);
+    }
+
     while (true) {
```
morgsmccauley, line 32: I'm not sure it's a good idea to just loop indefinitely, it will probably take up more CPU than necessary. Maybe we can set a timeout with an hour (?) delay instead. This will keep the event loop busy but reduce the amount of work done.

darunrs: Yeah that makes sense. Will do!

morgsmccauley: Looks like `setInterval()` isn't even needed, the application stays live without it, probably because there are bound ports. We can remove entirely :)

darunrs: Oh. Huh I don't know why I didn't attempt to try that out. Let me give that a go and I will remove it!

darunrs:
So we have decided to pivot away from a full release to a rolling release of V2. As a result, this PR will change to instead be focused on an approach where V and V2 share the same stream handlers, allowing Coordinator to affect the Redis streams set on its own, and turn off any executor, V1 or not. This allows to later introduce a allow list to transition indexers to V2 while only requiring allow list visibility to coordinator.

darunrs:
New use cases to test manually:
Redis set executor stopped by Server
Stopped executor restarted by server
Stopped executor restarted by redis
List shows both redis and non-redis streams

I believe that by returning a version of -1, we can make the change in Coordinator V2 trivial (I might be wrong about this):
If we put an if statement for the synchronizing executors code for the allow list, then whenever something is added, the code will see an existing executor, see a mismatched version, and stop and restart it. We only need to remove the stream from Redis before restarting. This could be a check for if version is -1 or something. Just an idea. 

Regardless, this code is now "live" as it does not care what version Coordinator is active. It will immediately execute anything its told to do. So, we should be careful when deploying Coordinator V2 to not enable that infra until we are ready. 

I've also added a env variable lock on the start and stop endpoints so we can deploy this to whatever env without explicitly enabling V2 functionality until we are ready.

runner/package-lock.json
```diff
@@ -23,7 +23,10 @@
         "prom-client": "^14.2.0",
```
morgsmccauley, line 23: Seems strange for this file to change when the dependencies in `package.json` have not changed - is this needed?

runner/src/server/runner-server.ts
```diff
@@ -2,28 +2,28 @@ import * as grpc from '@grpc/grpc-js';
...
   ) as unknown) as ProtoGrpcType;
 
   const server = new grpc.Server();
-  server.addService(runnerProto.runner.Runner.service, getRunnerService(StreamHandler));
+  server.addService(runnerProto.runner.Runner.service, getRunnerService(executors));
   const credentials = grpc.ServerCredentials;
 
   server.bindAsync(
-    '0.0.0.0:50007', // TODO: Read port from ENV
+    `${process.env.RUNNER_HOST ?? 'undefined'}:${process.env.RUNNER_PORT ?? 'undefined'}`,
```
morgsmccauley, line 20: We should probably just `assert()` these exist rather than setting to `undefined`?

runner/src/server/runner-service.ts
```diff
@@ -33,10 +29,20 @@ function getRunnerService (StreamHandlerType: typeof StreamHandler): RunnerHandl
         return;
       }
 
+      // TODO: Remove toggle after full V2 migration
+      if (process.env.RUNNER_VERSION !== 'V2') {
```
morgsmccauley, line 33: I'm not sure we even need a 'version' anymore. There's little harm in allowing this code to execute without this toggle, because it doesn't change the default behaviour. We need to actually call the RPC endpoints for something to happen.

runner/src/server/runner-service.ts
```diff
@@ -86,21 +112,27 @@ function getRunnerService (StreamHandlerType: typeof StreamHandler): RunnerHandl
...
-      console.log('ListExecutors called');
-      // TODO: Return more information than just executorId
       const response: ExecutorInfo__Output[] = [];
       try {
-        streamHandlers.forEach((handler, executorId) => {
-          if (handler.indexerConfig?.account_id === undefined || handler.indexerConfig?.function_name === undefined) {
-            throw new Error(`Stream handler ${executorId} has no/invalid indexer config.`);
+        executors.forEach((handler, executorId) => {
+          let config = handler.getIndexerConfig();
+          if (config === undefined) {
```
morgsmccauley, line 119: Under what conditions is this `undefined`?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -33,12 +38,19 @@ export default class StreamHandler {
     }
   }
 
+  getIndexerConfig (): IndexerConfig | undefined {
```
morgsmccauley, line 41: Why don't we just keep `indexerConfig` `public` and then just read it directly? Is there benefit to this getter?

runner/src/server/runner-service.ts
morgsmccauley: Oh because the config is read internally 🤔 So we'll get no information for all V1 indexers?

morgsmccauley (commented):
Great job! Just a few comments :)

runner/package-lock.json
darunrs: Ah no these changes are not necessary. These changes are from a different branch where I am instrumenting Runner with a tracer library. Not sure why the changes carried over between checkouts. Maybe package-lock is ignored by gitignore? Good catch though! I'll reset the lock file and submit that.

runner/src/server/runner-server.ts
darunrs: That would work! Will change.

runner/src/server/runner-service.ts
darunrs: I wasn't sure if we wanted an always active V2 form of Runner depending on how we activate Coordinators during the transition. But yeah the impact is probably minimal, so I'll remove the toggle.

runner/src/server/runner-service.ts
darunrs: Correct. The config is passed in for V2 executors and is used as opposed to pulling config from Redis. So, aside from my hardcoded obviously incorrect version number, the empty config should be another indicator its a V1 indexer.

runner/src/stream-handler/stream-handler.ts
darunrs: I guess it comes down to write protection. indexerConfig will be write-able after being passed in as public, which didn't fit the intention of the config which is passed in and is only internally modified. I initially had it as public since we would modify it in the updateExecutor API but we don't need that anymore.

runner/src/server/runner-service.ts
morgsmccauley: We'll need the account ID/function name so we know which executor to stop, we can probably extract that from the redis stream key, which in this case would be the `executorId`?

runner/src/stream-handler/stream-handler.ts
morgsmccauley: I think we can have it a `public readonly` so that it is visible externally, but can't be modified.

darunrs: OH. I don't know why I forgot. The underlying fields for the object can be modified, but not the config object's reference. So yeah that would work. I need field modification to update the status of the executor if it fails.

morgsmccauley: Ahh, yeah you're right, only the reference itself is `readonly`.

It seems strange to have `IndexerConfig.status`, it isn't _really_ part of the configuration of the Indexer. I think we should have `status` as a separate property of the `StreamHandler`/`Executor`, as it's related to the execution of that configuration.

Then we can we can keep `indexerConfig` immutable, as it should be?

darunrs: Yeah I thought the same thing just now looking back on it. I think that's probably a fair separation to make, and we might as well make it early on. 

I don't think I can even make indexerConfig immutable easily. readonly is the strongest immutability characteristic I get in TypeScript without spending extra effort in some way (Deep immutability through freezing, separating inner fields, etc.), I believe. I don't think we need to go that far here. 

I'll make a separate set of attributes called ExecutorContext perhaps, which contains information like status.

morgsmccauley: Yeah, I think `Object.freeze` or `as const` is what we want, but we probably don't need to commit to that at this point.

runner/src/server/runner-service.ts
darunrs: Yep that's correct. The redis stream is the "executorId" for V1 indexers. In fact, list functions API returns the redisStream as the executor Id for V1 indexers.

morgsmccauley: So in that case can we extract the `account_id` and `function_name` from the redis stream and populate that information here

darunrs: I hadn't populated it before since I thought it wasn't needed. What's the information used for given executorId is all that's necessary to drop an indexer?

morgsmccauley: The `executorId` is all that is needed to stop the executor, but we need to know which indexer that ID corresponds to in order to stop the right one.

We could to the `executorId`/redis stream parsing on the Coordinator side, but I want to avoid making that assumption since it won't be true in V2.

darunrs: Oh I see. Ok, sorry I understand what you're getting at now. Yeah, let me go ahead and do that parsing.

darunrs:
Example listing of V1 and V2 indexers together. I manually populated streams with 2 and then added one more using the API.  

```
start:  {
  executorId: '5fa3c3ee92d875791598e775738d1fd98a889a10979b600c154e40f9965892c8'
}
list:  {
  executors: [
    {
      executorId: 'flatirons.near/social_blockheight:real_time:stream',
      accountId: 'flatirons.near',
      functionName: 'social_blockheight',
      status: 'RUNNING',
      version: [Long]
    },
    {
      executorId: 'flatirons.near/sweat_blockheight:real_time:stream',
      accountId: 'flatirons.near',
      functionName: 'sweat_blockheight',
      status: 'RUNNING',
      version: [Long]
    },
    {
      executorId: '5fa3c3ee92d875791598e775738d1fd98a889a10979b600c154e40f9965892c8',
      accountId: 'darunrs.near',
      functionName: 'test_sweat_blockheight',
      status: 'RUNNING',
      version: [Long]
    }
  ]
}
```

morgsmccauley (approved):
Great work :)

PR #491, opened 2024-01-08
feat: Create rust GRPC client for Runner

### LINKED ISSUE #468: Create Rust GRPC client for Runner

morgsmccauley:
This PR creates a rust based GRPC client for Runner, for use within Coordinator V2. For now, this exists as its own crate within the top-level directory. The Coordinator PR was becoming quite large so I decided to separate this out.

Additionally, I've made a couple changes to the Runner proto:
- Rename the package: `spec` -> `runner`
- Remove `executorId` from `StartExecutorRequest` - it should be deterministic and assigned internally
- Add `version` to the executor - this will be used to determine whether an executor should be restarted


runner/protos/runner.proto
```diff
@@ -14,12 +14,13 @@ service Runner {
 
 // Start Executor Request 
 message StartExecutorRequest {
-    string executor_id = 1;
     string redis_stream = 2;
```
darunrs, line 17: The IDs for the fields should be reset. Right now they start at 2 instead of 1.

runner/protos/runner.proto
```diff
@@ -14,12 +14,13 @@ service Runner {
...
     string account_id = 3;
     string function_name = 4;
     string code = 5;
     string schema = 6;
+    // Block height corresponding to the created/updated height of the indexer
+    uint64 version = 7;
```
darunrs, line 23: The comment mentions block height whereas the field is version. What's actually being passed here?

runner/src/server/runner-service.ts
```diff
@@ -1,41 +1,64 @@
...
       // Validate request
-      const validationResult = validateStartExecutorRequest(call.request, streamHandlers);
+      const validationResult = validateStartExecutorRequest(call.request);
       if (validationResult !== null) {
         callback(validationResult, null);
         return;
       }
 
+      const { accountId, functionName, code, schema, redisStream, version } = call.request;
+      const executorId = hashString(`${accountId}/${functionName}`);
```
darunrs, line 37: If we're using accountId and function name anyway, is there a need for hashing here? Couldn't we just use accountId/functionName directly? I had envisioned executorId would be something coordinator decides (And it includes whatever information it wants to generate it, could be function name, account id, maybe machine ID, version, etc. in the future, whatever it is) and it simply passes it over to Runner so that it can be stored for access in that way. Coordinator can then pass that, in the future to one of many runners, or maybe a load balancer, depending on how we set things up. Did you have any ideas on leveraging determinism of the ID for anything or is it more just because its a good quality to have.

runner/src/server/runner-service.ts
```diff
@@ -69,13 +92,15 @@ function getRunnerService (StreamHandlerType: typeof StreamHandler): RunnerHandl
       const response: ExecutorInfo__Output[] = [];
       try {
         streamHandlers.forEach((handler, executorId) => {
+          console.log({ streamHandlers, handler, executorId });
```
darunrs, line 95: I assume you mean to remove this logging statement before merging?

darunrs (commented):
So it looks like you went with the approach of having client be a separate folder and keeping a copy of the runner proto file. I wasn't sure about that approach since we'd have duplicate proto files and I felt like the build process would be a bit wonky since (New) Coordinator would then have to copy the folder in, and then build it separately, and then I assume do something with the build output? I wasn't really sure how that would end up looking. What's the integration in Docker like? Have you tried using it in new coordinator, existing coordinator, or block streamer as a test?

runner/protos/runner.proto
morgsmccauley: Normally we shouldn't (https://protobuf.dev/programming-guides/proto3/#consequences), but since the endpoint isn't live yet, in this case we can :)

runner/protos/runner.proto
morgsmccauley: Block Height. The intention behind this field is to map the executor/block stream to the created/updated block height of the registry entity. This allows Coordinator to compare the "versions" to determine whether an update is needed.

`version` will either be `created_at_block_height` or `updated_at_block_height`, naming it either of these would create confusion IMO as it may lead to thinking it corresponds to only one of the fields, and not either.

darunrs: Ahh I see. That makes sense.

runner/protos/runner.proto
darunrs: Ahh good point. Yeah once live we certainly can't do that anymore.

runner/src/server/runner-service.ts
morgsmccauley: Typically the service (Runner) which manages the entities (Executors) assigns the IDs, as they have the most intimate knowledge of these entities and can ensure there are no collisions. I'm not sure there is benefit in Coordinator generating these IDs.

As for hashing, it's generally a good idea to keep IDs opaque. This allows us to change how these IDs are generated going forward without impacting clients, which I have been somewhat thinking about, we may not need deterministic IDs after all. Without opaque IDs, clients might construct the IDs themselves, and then run in to issues when the IDs change on the server side.

runner/src/server/runner-service.ts
morgsmccauley: Yes :)

morgsmccauley:
> So it looks like you went with the approach of having client be a separate folder and keeping a copy of the runner proto file. I wasn't sure about that approach since we'd have duplicate proto files and I felt like the build process would be a bit wonky since (New) Coordinator would then have to copy the folder in, and then build it separately, and then I assume do something with the build output? I wasn't really sure how that would end up looking. What's the integration in Docker like? Have you tried using it in new coordinator, existing coordinator, or block streamer as a test?

This is being used in https://github.com/near/queryapi/pull/444 now. The Docker context needs to be at the root directory anyway since Coordinator depends on two separate crates.

Duplicate proto files might be something we just have to live with until we migrate Runner to Rust 😅

runner/src/server/runner-service.ts
darunrs: Gotcha. Yeah you're right that Coordinator managing the IDs is not particularly valuable since it will be given the ID back when it creates the executor, which it can use for stopping an executor later. This is actually better as Runner being the ID creator means Coordinator is less "stateful". Like, if Coordinator dies and we restart it and have it load the existing runner streams and their IDs, then having Coordinator assume those IDs makes more sense if Runner created them, rather than some previous instance of Coordinator.

And yeah your points on changing ID generation was what I was considering. Changing the way we ID the executors is probably not too difficult either, if we continue with the approach of Runner creating it and passing it back.

darunrs (approved):
LGTM! I understand what you mean by having the client at root. I'll learn whatever strategy you use for linking so I know it for the future. And yeah, relying on proto files for codegen is indeed one of the cons of using protobuf. We might have to do it again if we eventually allow developers to stop and start their indexers from the UI, but that's likely a long time from now, if we ever do so. 

One last thing, I noticed you are using the hard coded host:port values. That's also fine for now, though in my PR I did address that by setting that up. Its up to you if you want to include that in this PR. I feel like it's a good place to do so. I or you can of course do so in a future PR too.

morgsmccauley:
> One last thing, I noticed you are using the hard coded host:port values. That's also fine for now, though in my PR I did address that by setting that up. Its up to you if you want to include that in this PR. I feel like it's a good place to do so. I or you can of course do so in a future PR too.

Let's get this merged and we can address in a future PR :)

PR #502, opened 2024-01-16
refactor: Remove hard-coded shard count

### LINKED ISSUE #382: Remove hardcoded shard number from prototype
In the initial prototype where S3 fetch time was a major impediment, number of shards was hardcoded to 4 in the runner. 
https://github.com/near/queryapi/blob/main/runner/src/lake-client/lake-client.ts#L80

Now that we have some pre-fetching strategies let's return to pulling that from the block.json or move it to config.

morgsmccauley:
Currently, the `shard`/`chunk` count is hard-coded to 4 so that we can fetch the block header and shards in parallel. This PR removes the hard-coded value by using the chunk count specified in the block header to fetch the relevant shards.

### DIFF
--- a/block-server/handler.js
+++ b/block-server/handler.js
@@ -35,16 +35,12 @@ const normalizeBlockHeight = function(block_height) {
 }
 
 const fetchStreamerMessage = async function(block_height, options) {
-    const blockPromise = fetchBlockPromise(block_height, options);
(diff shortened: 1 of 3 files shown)


darunrs (approved):
Nice idea to use block chunks to get shard count!

PR #503, opened 2024-01-16
refactor: Configure `coordinator`/`block-streamer` via environment

### LINKED ISSUE #421: Create Control service

morgsmccauley:
- refactor: Configure `coordinator` via environment
- refactor: Configure `block-streamer` via environment

### DIFF
--- a/block-streamer/src/main.rs
+++ b/block-streamer/src/main.rs
@@ -18,9 +18,12 @@ async fn main() -> anyhow::Result<()> {
         .with(tracing_subscriber::EnvFilter::from_default_env())
         .init();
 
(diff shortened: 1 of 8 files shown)


darunrs (approved):
LGTM! Can you also add the env variables to Docker Compose as well?

morgsmccauley:
> LGTM! Can you also add the env variables to Docker Compose as well?

These services aren't even part of Docker Compose at all as they don't have `Dockerfile`s yet 😄 - I'll be sure to add them when they do

PR #515, opened 2024-01-17
feat: Capture errors thrown within Coordinator

### LINKED ISSUE #421: Create Control service

morgsmccauley:
Currently, errors thrown within Coordinator V2 will bubble up to `main()` and cause the entire application to exit. This PR captures those errors, handling them accordingly. 

Errors are handled in either of the following ways:
1. Exponentially retry - data which is 'critical' to the control, i.e. the indexer registry, executor/stream list, will be continuously retried, blocking the control loop from further progress as it would not make sense to continue without this information.
2. Swallowed - actions such as starting/stopping executors/streams will be logged and swallowed. This is preferable over exponential retries as individual failures will not block the progress of the control loop, therefore allowing other indexers to be acted on. Skipping should be fine in this case as it will be retried in the next loop.

I expect this behaviour to evolve over time as we learn more about the system, the important thing here is that Coordinator will not crash on errors.

### DIFF
--- a/coordinator/src/block_streams_handler.rs
+++ b/coordinator/src/block_streams_handler.rs
@@ -1,12 +1,11 @@
 use anyhow::Context;
-use tonic::transport::channel::Channel;
-use tonic::Request;
-
 use block_streamer::block_streamer_client::BlockStreamerClient;
 use block_streamer::{
     start_stream_request::Rule, ActionAnyRule, ActionFunctionCallRule, ListStreamsRequest,
     StartStreamRequest, Status, StopStreamRequest, StreamInfo,
 };
+use tonic::transport::channel::Channel;
+use tonic::Request;
 
 #[cfg(not(test))]
 pub use BlockStreamsHandlerImpl as BlockStreamsHandler;
@@ -20,9 +19,10 @@ pub struct BlockStreamsHandlerImpl {
 #[cfg_attr(test, mockall::automock)]
 impl BlockStreamsHandlerImpl {
     pub async fn connect(block_streamer_url: String) -> anyhow::Result<Self> {
-        let client = BlockStreamerClient::connect(block_streamer_url)
-            .await
-            .context("Unable to connect to Block Streamer")?;
+        let channel = Channel::from_shared(block_streamer_url)
+            .context("Block Streamer URL is invalid")?
+            .connect_lazy();
+        let client = BlockStreamerClient::new(channel);
 
         Ok(Self { client })
     }
@@ -32,17 +32,25 @@ impl BlockStreamsHandlerImpl {
             .client
             .clone()
             .list_streams(Request::new(ListStreamsRequest {}))
-            .await?;
+            .await
+            .context("Failed to list streams")?;
 
         Ok(response.into_inner().streams)
     }
 
     pub async fn stop(&self, stream_id: String) -> anyhow::Result<()> {
-        let request = Request::new(StopStreamRequest { stream_id });
+        let request = Request::new(StopStreamRequest {
+            stream_id: stream_id.clone(),
+        });
 
         tracing::debug!("Sending stop stream request: {:#?}", request);
 
-        let _ = self.client.clone().stop_stream(request).await?;
+        let _ = self
+            .client
+            .clone()
+            .stop_stream(request)
+            .await
+            .context(format!("Failed to stop stream: {stream_id}"))?;
 
         Ok(())
     }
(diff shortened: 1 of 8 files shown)


coordinator/src/block_streams_handler.rs
```diff
@@ -19,29 +20,44 @@ pub struct BlockStreamsHandlerImpl {
...
 #[cfg_attr(test, mockall::automock)]
 impl BlockStreamsHandlerImpl {
-    pub async fn connect(block_streamer_url: String) -> anyhow::Result<Self> {
-        let client = BlockStreamerClient::connect(block_streamer_url)
-            .await
-            .context("Unable to connect to Block Streamer")?;
+    pub fn connect(block_streamer_url: String) -> anyhow::Result<Self> {
+        let channel = Channel::from_shared(block_streamer_url)
+            .context("Block Streamer URL is invalid")?
+            .connect_lazy();
```
morgsmccauley, line 26: Defer connection to when calls are made. This allows us to consolidate retries (in the code below) rather than also having retry logic here.

coordinator/src/main.rs
```diff
@@ -29,21 +35,26 @@ async fn main() -> anyhow::Result<()> {
...
         let indexer_registry = registry.fetch().await?;
 
-        synchronise_executors(&indexer_registry, &mut executors_handler).await?;
-        synchronise_block_streams(&indexer_registry, &redis_client, &mut block_streams_handler)
-            .await?;
+        tokio::try_join!(
+            synchronise_executors(&indexer_registry, &executors_handler),
+            synchronise_block_streams(&indexer_registry, &redis_client, &block_streams_handler),
+            async {
+                sleep(CONTROL_LOOP_THROTTLE_SECONDS).await;
```
morgsmccauley, line 48: Forcing a minimum loop duration

coordinator/src/block_streams_handler.rs
```diff
@@ -19,29 +20,44 @@ pub struct BlockStreamsHandlerImpl {
...
-            .client
-            .list_streams(Request::new(ListStreamsRequest {}))
-            .await?;
-
-        Ok(response.into_inner().streams)
+    pub async fn list(&self) -> anyhow::Result<Vec<StreamInfo>> {
+        exponential_retry(|| async {
+            let response = self
+                .client
+                .clone()
```
morgsmccauley, line 36: Cloning a gRPC/tonic client is cheap and allows us to avoid holding a mutable reference - https://docs.rs/tonic/latest/tonic/transport/channel/struct.Channel.html#multiplexing-requests

coordinator/src/block_streams_handler.rs
darunrs: I see. So basically when we make the call, the channel connection AND the call success is within the same retry loop?

coordinator/src/block_streams_handler.rs
```diff
@@ -19,29 +20,44 @@ pub struct BlockStreamsHandlerImpl {
...
 
-    pub async fn list(&mut self) -> anyhow::Result<Vec<StreamInfo>> {
-        let response = self
-            .client
-            .list_streams(Request::new(ListStreamsRequest {}))
-            .await?;
-
-        Ok(response.into_inner().streams)
+    pub async fn list(&self) -> anyhow::Result<Vec<StreamInfo>> {
+        exponential_retry(|| async {
```
darunrs, line 33: Does exponential retry have a limit or does it increase forever? Might be good to have an upper limit so that we don't necessarily have to restart coordinator too if the error was say block streamer or runner side.

coordinator/src/block_streams_handler.rs
```diff
@@ -19,29 +20,44 @@ pub struct BlockStreamsHandlerImpl {
...
 
         tracing::debug!("Sending stop stream request: {:#?}", request);
 
-        let _ = self.client.stop_stream(request).await?;
+        let _ = self
+            .client
+            .clone()
+            .stop_stream(Request::new(request.clone()))
+            .await
+            .map_err(|e| {
```
darunrs, line 58: What happens if a stop fails, is swallowed, and the subsequent start (e.g. an indexer update) is successful? Is there a mechanism in block stream to prevent duplicates? It may be worth skipping with continue if we don't have any duplication mechanisms in place.

coordinator/src/block_streams_handler.rs
morgsmccauley: Yes, that's correct

coordinator/src/block_streams_handler.rs
morgsmccauley: No it will just increase forever, good call, we should cap the delay seconds, i'll do that in a follow up PR.

coordinator/src/block_streams_handler.rs
morgsmccauley: Block Streamer service will throw an error if we try to start the same stream for a given indexer

darunrs (approved):

PR #516, opened 2024-01-18
feat: Support Deployment of Block Streamer

### LINKED ISSUE #443: Setup infra for Block Streamer

darunrs:
Block Streamer needs a Dockerfile in order to build the image and be deployable either locally or in GCP. I've created one for the service and updated the compose file to set the correct env variables needed. This also sserves as a record for what ENV variables need to be set during Terraform deployments to GCP. 

In addition, there's a small issue with running QueryApi locally where if the user runs Runner locally through yarn start, provisioning of resources through hasura fails because it populates the postgres host as 'localhost' whereas it should be 'postgres', as the calls are processed inside the hasura docker. This is due to PGHOST being both used and also passed to Hasura inside Runner. To fix this, I created a separate env variable called PGHOST_HASURA which can be set to 'postgres'. This way, Runner can use 'localhost' while it passes 'postgres' to hasura's docker.

### DIFF
--- a/block-streamer/Dockerfile
+++ b/block-streamer/Dockerfile
@@ -1,0 +1,16 @@
+FROM rust:1.75 AS build
+ARG CARGO_BUILD_MODE=debug
+WORKDIR /tmp/
+COPY block-streamer/ block-streamer/
+COPY registry/types/ registry/types/
+WORKDIR /tmp/block-streamer/
+RUN apt update && apt install -yy protobuf-compiler
+RUN cargo build --package block-streamer
+
+FROM ubuntu:22.04
+ARG CARGO_BUILD_MODE=debug
+ENV RUST_LOG=info
+RUN apt update && apt install -yy openssl ca-certificates
+USER nobody
+COPY --from=build /tmp/block-streamer/target/$CARGO_BUILD_MODE/block-streamer /block-streamer
+ENTRYPOINT ["/block-streamer"]
--- a/block-streamer/Dockerfile
+++ b/block-streamer/Dockerfile
@@ -5,7 +5,11 @@ COPY block-streamer/ block-streamer/
 COPY registry/types/ registry/types/
 WORKDIR /tmp/block-streamer/
 RUN apt update && apt install -yy protobuf-compiler
-RUN cargo build --package block-streamer
+RUN if [ "$CARGO_BUILD_MODE" = "debug" ]; then \
+        cargo build --package block-streamer; \
+    else \
+        cargo build --release --package block-streamer; \
+    fi
 
 FROM ubuntu:22.04
 ARG CARGO_BUILD_MODE=debug
--- a/block-streamer/Dockerfile
+++ b/block-streamer/Dockerfile
@@ -13,7 +13,6 @@ RUN if [ "$CARGO_BUILD_MODE" = "debug" ]; then \
 
 FROM ubuntu:22.04
 ARG CARGO_BUILD_MODE=release
-ENV RUST_LOG=info
 RUN apt update && apt install -yy openssl ca-certificates
 USER nobody
 COPY --from=build /tmp/block-streamer/target/$CARGO_BUILD_MODE/block-streamer /block-streamer
--- a/block-streamer/Dockerfile
+++ b/block-streamer/Dockerfile
@@ -1,5 +1,5 @@
 FROM rust:1.75 AS build
-ARG CARGO_BUILD_MODE=debug
+ARG CARGO_BUILD_MODE=release
 WORKDIR /tmp/
 COPY block-streamer/ block-streamer/
 COPY registry/types/ registry/types/
@@ -12,7 +12,7 @@ RUN if [ "$CARGO_BUILD_MODE" = "debug" ]; then \
     fi
 
 FROM ubuntu:22.04
-ARG CARGO_BUILD_MODE=debug
+ARG CARGO_BUILD_MODE=release
 ENV RUST_LOG=info
 RUN apt update && apt install -yy openssl ca-certificates
 USER nobody
--- a/block-streamer/src/main.rs
+++ b/block-streamer/src/main.rs
@@ -23,17 +23,18 @@ async fn main() -> anyhow::Result<()> {
 
     tracing::info!("Starting Block Streamer Service...");
 
+    tracing::info!("Connecting to Redis...");
     let redis_client = std::sync::Arc::new(redis::RedisClient::connect(&redis_url).await?);
-
-    tracing::info!("Connected to Redis");
 
     let aws_config = aws_config::from_env().load().await;
     let s3_config = aws_sdk_s3::Config::from(&aws_config);
     let s3_client = crate::s3_client::S3Client::new(s3_config.clone());
 
+    tracing::info!("Connecting to Delta Lake...");
     let delta_lake_client =
         std::sync::Arc::new(crate::delta_lake_client::DeltaLakeClient::new(s3_client));
 
+    tracing::info!("Starting gRPC Server...");
     server::init(&server_port, redis_client, delta_lake_client, s3_config).await?;
 
     Ok(())
--- a/block-streamer/src/main.rs
+++ b/block-streamer/src/main.rs
@@ -25,6 +25,8 @@ async fn main() -> anyhow::Result<()> {
 
     let redis_client = std::sync::Arc::new(redis::RedisClient::connect(&redis_url).await?);
 
+    tracing::info!("Connected to Redis");
+
     let aws_config = aws_config::from_env().load().await;
     let s3_config = aws_sdk_s3::Config::from(&aws_config);
     let s3_client = crate::s3_client::S3Client::new(s3_config.clone());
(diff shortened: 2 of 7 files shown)


darunrs:
The testing I did locally was relatively simple: 

1. I built the images using docker compose. 
2. I ran the resources, Coordinator V1, and block streamer image locally in Docker.
3. I ran Runner V2 and Coordinator V2 locally through CLI. 
4. I hard coded Coordinator V2 to only init one indexer. 
5. Verified Block Stream and Executor were initialized, that blocks were in stream, and that Runner was running them alongside V1 executors. 

THere's probably some edge case testing needed as well. Open to ideas on what things I should try. 

Something I noticed though is that I was getting a lot of S3 errors locally when running all current dev indexers in V2. Might be a problem with running locally and not in GCP, but its something to consider. Caching might help if the indexers are reasonably close in terms of what blocks they're querying.

block-streamer/src/main.rs
```diff
@@ -23,15 +23,18 @@ async fn main() -> anyhow::Result<()> {
...
 
     let aws_config = aws_config::from_env().load().await;
     let s3_config = aws_sdk_s3::Config::from(&aws_config);
     let s3_client = crate::s3_client::S3Client::new(s3_config.clone());
 
+    tracing::info!("Connecting to Delta Lake...");
     let delta_lake_client =
         std::sync::Arc::new(crate::delta_lake_client::DeltaLakeClient::new(s3_client));
 
+    tracing::info!("Starting gRPC Server...");
```
morgsmccauley, line 37: We already have a log for this in `block-streamer/src/server/mod.rs`, can we remove either so we just have one?

block-streamer/src/server/mod.rs
```diff
@@ -10,7 +10,7 @@ pub async fn init(
     delta_lake_client: std::sync::Arc<crate::delta_lake_client::DeltaLakeClient>,
     lake_s3_config: aws_sdk_s3::Config,
 ) -> anyhow::Result<()> {
-    let addr = format!("[::1]:{}", port).parse()?;
+    let addr = format!("[::]:{}", port).parse()?;
```
morgsmccauley, line 13: What was the issue here?

block-streamer/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM rust:1.75 AS build
+ARG CARGO_BUILD_MODE=debug
```
morgsmccauley, line 2: can we default to `release` please?

block-streamer/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN apt update && apt install -yy protobuf-compiler
+RUN if [ "$CARGO_BUILD_MODE" = "debug" ]; then \
+        cargo build --package block-streamer; \
+    else \
+        cargo build --release --package block-streamer; \
+    fi
+
+FROM ubuntu:22.04
+ARG CARGO_BUILD_MODE=debug
+ENV RUST_LOG=info
```
morgsmccauley, line 16: Can this still be overridden from the outside?

block-streamer/Dockerfile
darunrs: Ah yes, I forgot to change it to release again. Will change!

block-streamer/Dockerfile
darunrs: So that's a good point. I can look into that.

block-streamer/src/server/mod.rs
darunrs: I was getting the following error: 
`Error: transport error
Caused by:
    0: error creating server listener: Cannot assign requested address (os error 99)
    1: Cannot assign requested address (os error 99)`

I tried :: instead, and that worked.

darunrs: Looking into it some more now, I don't think that's actually what we want. :: is a wildcard so I don't know for sure what address it's attaching the port on. I'll investigate this some more.

block-streamer/src/main.rs
darunrs: Ah you're right. I'll remove the one I added.

block-streamer/Dockerfile
darunrs: It can be overridden using docker compose file's ENV. I'm not sure about GCP deployment. But for GCP it oughta be release anyway.

block-streamer/src/server/mod.rs
darunrs: Ok I think I figured this all out. So, when you put ::1, it binds to the loopback address, which is localhost (127.0.0.1). This means connections from inside the container itself loop back. If I set up the address as ::1, then the server listens on localhost:port. The problem is this means the server is only reachable from inside the container, regardless of if you did port forwarding or not. The other reason is probably some IPv6 compatibility issues. I tried 127.0.0.1 instead of ::1 and that did not give the error. 

In addition, the reason that I had localhost:port working in Runner V2 was because it was running locally too. I had Block Streamer running in docker while Runner and Coordinator were ran through yarn start and cargo run, respectively. So, COordinator and Runner could communicate through localhost. I tried again with Runner V2 in DOcker, and Coordinator V2, running on the hsot machine, could nto reach either Block Streamer OR Runner. I changed both Block Streamer and Runner V2 to put their server on 0.0.0.0 and it worked. 

The reason 0.0.0.0 works is because it binds the port number to all available IP Addresses. The container is aware of several: the IP Address assigned by the docker subnet, 127.0.0.1, the ethernet IP address from which subnet traffic is routed to, and probably some others. That last one is the important one. By using 0.0.0.0, we ensure the server is aware of any network traffic going through that port. 

So, now with that in mind, this felt a little awkward to me because it felt a little unsecure to listen to all IP addresses in the container for that port. But, given we run these services in GCP in a subnet, that's probably fine? We just need to make sure we don't expose those ports unless we want them to be callable as well. I haven't experimented with specifying an ip address, instead of using 0.0.0.0. Maybe that would work. I couldn't find any information saying that is the right way to do it. People generally just say use 0.0.0.0 and configure your firewalls and access separately. 

I think we can proceed with 0.0.0.0 for now and revisit when we're setting up networking later to decide how to route everything. But 0.0.0.0 will at least ensure we can fully test functionality. 127.0.0.1 does not seem to work. 

Of coruse, let me know if you find conflicting information or results!

block-streamer/Dockerfile
morgsmccauley: I'm referring to `RUST_LOG` here - let's remove, as it should be set in the terraform config anyway?

morgsmccauley (approved):
LGTM just the `RUST_LOG` and `0.0.0.0` change to make :)

block-streamer/Dockerfile
darunrs: Oh sorry you mean that. I had to put that in because leaving it out prevents tracing logs from appearing inside Docker. Now of course we can set that in docker compose and terraform too if we want. But I felt like it made sense to leave it here as we probably wouldn't override it unless we really wanted to, in which case we totally can by placing it there. Removing it means we need to specify it in deployment explicitly without any default. 

Essentially, removing it here means we don't have a default setting. It's probably a minor change but if we want to be explicit about all env variables, doing so in the same place, I'm totally fine going with that.

morgsmccauley: Yeah i'd prefer to keep it in one place to avoid confusion

darunrs:
Block Streamer currently uses [::] which is dual stack 0.0.0.0, meaning it works for both IPv4 and IPv6 addresses. I changed Runner's server to [::] and that seems to have worked as well.

morgsmccauley:
> Block Streamer currently uses [::] which is dual stack 0.0.0.0, meaning it works for both IPv4 and IPv6 addresses. I changed Runner's server to [::] and that seems to have worked as well.

@darunrs I may be wrong, but it seems [::] and `0.0.0.0` are for ipv6 and ipv4 respectively, not the same. If they are, it might be worth checking if specifying ipv6 will conflict with the [firewall config](https://github.com/near/near-ops/pull/1545/files#diff-b4eee016b1bfb66169250d595651fb61a5996c8bd5c4e911ba56f20a1f866b0cR95), which is set to `0.0.0.0`. I wonder if GCP will only expose ipv4 traffic, and not ipv6.

darunrs:
> > Block Streamer currently uses [::] which is dual stack 0.0.0.0, meaning it works for both IPv4 and IPv6 addresses. I changed Runner's server to [::] and that seems to have worked as well.
> 
> @darunrs I may be wrong, but it seems `[::]` and `0.0.0.0` are for ipv6 and ipv4 respectively, not the same. If they are, it might be worth checking if specifying ipv6 will conflict with the [firewall config](https://github.com/near/near-ops/pull/1545/files#diff-b4eee016b1bfb66169250d595651fb61a5996c8bd5c4e911ba56f20a1f866b0cR95), which is set to `0.0.0.0`. With this config, I wonder if GCP will only expose ipv4 traffic, and not ipv6.

Good catch. Our VPC Network currently only supports IPv4. We don't really have a reason to change that. I'll update the addresses to expose on 0.0.0.0 explicitly instead!

PR #517, opened 2024-01-19
feat: Enable Block Streams Start from V1 Interruption

### LINKED ISSUE #417: Enable dedicated streams to start from interruption
After https://github.com/near/queryapi/issues/418, each Indexer will have it's own dedicated stream. To persist these streams across restarts, we need to store the `last_indexed_block` in Redis, so that it can continue where it left off, essentially "Start from interruption".

This will need to be implemented in Coordinator V1, so when the new Control Plane starts, it's able to continue where it left off. But, this should also be implemented in #418, within the new Block Streamer service.

darunrs:
Coordinator V2 currently defaults to starting block streams with the start_block_height value populated in the registry. In order for a smooth migration from V1 to V2, we instead want Coordinator V2 to create block streams starting where the current indexers left off. 

Coordinator V1 now writes a last_published_block value for all indexers it manages on each matching block. 

Coordinator V2's start_block_height setting workflow needs refactoring. This is due to the fact that the exisitng flow defauls to using start_block_height as its present in the registry even after we begin processing a stream from that block height. We need to refactor the flow to ensure we use the start block height once, and otherwise use redis last published block height, or fallback if not present. 

New flow to handle edge cases will be tracked by https://github.com/near/queryapi/issues/521.


coordinator/src/main.rs
```diff
@@ -112,37 +112,48 @@ async fn synchronise_block_streams(
...
-                ))
-                .await
-            {
-                last_published_block
-            } else if let Some(updated_at_block_height) = indexer_config.updated_at_block_height {
-                updated_at_block_height
-            } else {
-                indexer_config.created_at_block_height
-            };
+            if start_block_height == None {
```
morgsmccauley, line 130: I don't think this refactoring is necessary. `start_block_height` will be set to `None` for non-historical flows, therefore it should always take priority when set. We shouldn't need to consider 'updating an existing' vs 'starting for the first time' we can just start a block stream with the desired configuration, regardless of the current configuration.

So the flows we are left with are:
- Use `start_block_height` when set - developer is triggering historical backfill
- Use `last_published_block` when `start_block_height` not set - developer updated code/schema etc. so continue where we left off
- Use `updated_at_block_height`/`created_at_block_height` - first time encountering indexer so just take the block height at which it was published

I checked out the branch, restored the original logic, and ran the tests, the only failure was:
```
test tests::block_stream::uses_last_published_block_height_when_restarting_existing_indexer
_block_stream ... FAILED
```

So even with the original logic we get the desired behaviour, as all other tests, including your new ones, pass. This failure is expected - the indexer has `start_block_height` defined and should therefore be used. Setting it to `None` makes the test case pass. 

Can we please restore the original logic?

morgsmccauley (changes requested):

coordinator/src/main.rs
darunrs: So with the original flow, it always used start_block_height because the set value persists in the registry. So, if I kicked off a backfill 2 days ago and set the block height, and it ran and completed, it would continue to exist in the registry. If I restarted the block stream, it would see the start_block_height field populated even though the backfill had completed previously.

morgsmccauley: Ah I understand now, the currently logic will use `start_block_height` for every fresh start, which is incorrect. Great catch :).

So we need to ensure `start_block_height` is only used _once_. The logic here will use it if:
1. The new version doesn't match the existing block stream version, or
2. There is no existing block stream and no `last_published_block`

But this will still cause issues if the Block Streamer service is restarted _while_ a new indexer version is published with a `start_block_height` - No block stream will exist, so Coordinator will start a new one using `last_published_block`, while it should have `start_block_height`.

We need to persist some state so that Coordinator can start a stream only once, and then use `last_published_block` otherwise.

One approach could be to add the version to the redis stream, i.e. `morgs.near/test:block_stream:123`, where `123` is the `updated_at_block_height`/`created_at_block_height`. So the flow would be:
1. If an existing block stream exists, and the `version` is different - stop it
2. If `start_block_height` exists
    1. if a redis stream exists for the current version - use it with `last_published_block`
    2. else - create it and use `start_block_height
3. else
    1. if a redis stream exists for the current version - use it with `last_published_block`
    2. else - create it and use `updated_at_block_height`/`created_at_block_height`

So in essence, we only use `last_published_block` if a redis stream exists for the current version, otherwise we start a new one.

Does this make sense? Do you see any issues with it?

darunrs: Yeah the error case you mentioned is true. Thanks for catching that! I think the approach broadly works. There's some additional things to consider now. We need to make sure the stream name written by Coordinator V1 matches what V2 expects.

In addition, for the case of no start block height, how do we migrate the existing messages? I imagine we could rename the stream? Is that approach equal, better, or worse than storing the version separately. I'm thinking having it separate might be easier to work with? Like, {indexer full name}:version or something. 

Let me know what you think, but I definitely think storing the version in Redis is a great way to go for covering that edge case.

morgsmccauley: Coordinator V1 will always use the the same stream names; `real-time`, and `historical`. That shouldn't be a concern for V2?

V1 will not need to know about the versioned redis streams, V2 will just move the messages from the streams mentioned above to a new versioned stream.

Let's revert the logic for now and raise another ticket which we can discuss further to make sure we've fully captured all cases?

darunrs: Gotcha. And I think renaming should be doable. We'll want to time the process, if we go with migrating the messages. Just to make sure it's fast. And yeah, we can revert the behavior for now and discuss it. Some time tomorrow should work? Go ahead and drop a meeting whenever works for you. Or we can wait until Wednesday's meeting.

morgsmccauley (approved):

PR #518, opened 2024-01-23
feat: Only start indexers set within the `allowlist`

### LINKED ISSUE #501: Selectively start indexers
As described in the rollout plan (https://github.com/near/queryapi/issues/422) we need the ability to have Coordinator V2 only start indexers within certain accounts, while also having Coordinator V1 ignore those indexers.

This will allow us to progressively rollout the change.

morgsmccauley:
This PR updates Coordinator V2 to read an `allowlist` from Redis, and only start accounts which have been added to that list. The `allowlist` is a single key in Redis stored as stringified JSON, e.g.:
```json
[{ "account_id": "morgs.near" }]
```

I've used JSON as each entry will eventually contain more fields, such as an acknowledgement from Coordinator V1, and a flag to set when completing the migration.

Additionally, I've added various logs across Coordinator V2.

### DIFF
--- a/coordinator/src/main.rs
+++ b/coordinator/src/main.rs
@@ -41,9 +41,11 @@ async fn main() -> anyhow::Result<()> {
     loop {
         let indexer_registry = registry.fetch().await?;
 
+        let allowlist: Vec<String> = redis_client.get("allowlist").await?;
+
         let indexer_registry = indexer_registry
             .into_iter()
-            .filter(|(account_id, _)| ["morgs.near"].contains(&account_id.as_str()))
+            .filter(|(account_id, _)| allowlist.contains(&account_id.to_string()))
             .collect();
 
         tokio::try_join!(
--- a/coordinator/src/main.rs
+++ b/coordinator/src/main.rs
@@ -1,5 +1,6 @@
 use std::time::Duration;
 
+use near_primitives::types::AccountId;
 use tokio::time::sleep;
 use tracing_subscriber::prelude::*;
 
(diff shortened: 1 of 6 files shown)


coordinator/src/main.rs
```diff
@@ -94,6 +142,13 @@ async fn synchronise_executors(
     }
 
     for unregistered_executor in active_executors {
+        tracing::info!(
+            account_id = unregistered_executor.account_id.as_str(),
+            function_name = unregistered_executor.function_name,
+            registry_version = unregistered_executor.version,
+            "Stopping executor"
```
darunrs, line 149: nit: Perhaps we should rename this to "Stopping unregistered executor" instead? Same for block streamer below.

coordinator/src/redis.rs
```diff
@@ -25,15 +25,17 @@ impl RedisClientImpl {
 
     pub async fn get<T, U>(&self, key: T) -> anyhow::Result<U>
     where
-        T: ToRedisArgs + Debug + Send + Sync + 'static,
-        U: FromRedisValue + Send + Sync + 'static,
+        T: ToRedisArgs + Debug + 'static,
+        U: FromRedisValue + Debug + 'static,
```
darunrs, line 29: What are Send, Sync, and Debug here? Are they traits that are required of the type that is passed in?

darunrs (commented):
Looks mostly good to me! What about allowlist in Coordinator V1? I feel like it makes sense to do both in the same PR since their behavior is strongly linked. What's the plan for if we remove an account from the allowlist? Or will we not do it at all? As in, if something goes wrong, hot fix instead of revert by removing from allowlist. I think hot-fixing makes more sense personally.

coordinator/src/redis.rs
morgsmccauley: Yes, they are "trait bounds" which are enforced/required for the arguments passed in.

darunrs (approved):
Gotcha. LGTM then!

coordinator/src/main.rs
morgsmccauley: Yeah good call, I'll update. I have another logging PR, I'll do it in that one.

PR #526, opened 2024-01-24
feat: Logging & Error updates

morgsmccauley:
- feat: Cap exponential retry to 30 secs
- feat: Debug log stream/executor responses
- feat: Add more logging to coordinator

### DIFF
--- a/coordinator/src/block_streams_handler.rs
+++ b/coordinator/src/block_streams_handler.rs
@@ -38,7 +38,11 @@ impl BlockStreamsHandlerImpl {
                 .await
                 .context("Failed to list streams")?;
 
-            Ok(response.into_inner().streams)
+            let streams = response.into_inner().streams;
+
+            tracing::debug!("List streams response: {:#?}", streams);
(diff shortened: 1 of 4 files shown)


darunrs (approved):
Nice! I'm curious how tracing::debug behaves in the GCP container. It shouldn't display in the container itself, which means Docker doesn't pick it up. That's good since we still have a hunch that Docker logs is the cause of the compute engine storage being consumed. But I wonder if it still shows up in the GCP logs?

morgsmccauley:
> Nice! I'm curious how tracing::debug behaves in the GCP container. It shouldn't display in the container itself, which means Docker doesn't pick it up. That's good since we still have a hunch that Docker logs is the cause of the compute engine storage being consumed. But I wonder if it still shows up in the GCP logs?

It depends on the `RUST_LOG` env var, if set to `debug`, the application will write these logs to stdout, and will therefore be surfaced to GCP. Otherwise they will not be output anywhere.

PR #527, opened 2024-01-25
feat: Auto migrate indexers to Control Plane

### LINKED ISSUE #520: Migrate Indexers to Control Plane
Indexers should either run on Coordinator V1 or V2, and a process needs to be set up to migrate these indexers across. This work should build on top of the existing `allowlist`.

For a given account, the flow will roughly be:
1. Coordinator V1 ignores accounts within the `allowlist` and sets the `v1_ack` - we need an acknowledge to ensure V2 doesn't start before V1 stopped
2. Coordinator V2 starts indexers within the `allowlist` which have the `v1_ack` set
3. Coordinator V2:
    1. Remove the indexer from the `streams` Redis Set, preventing Runner from starting it again
    2. Stop the current historical/real-time executors
    3. Move all messages from the historical/real-time streams to a new single stream
    4. Using the stream above, start a Block Stream from last_published_block, and start the corresponding Executor
    5. Set the `migration_complete` flag to prevent the migration from running again

morgsmccauley:
This PR builds on the Redis `allowlist` to auto migrate indexers from the existing infrastructure to the Control Plane. An account migration requires coordination between both the V1 and V2 architecture - the indexer must be removed/ignored from V1, and then correctly configured within V2.

## Allowlist shape
Each `allowlist` entry now contain the following properties:
```rs
pub struct AllowlistEntry {
    account_id: AccountId, // The account which should be migrated
    v1_ack: bool,          // True if Coordinator V1 has acknowledged the entry
    migrated: bool,        // True if the migration was successful
    failed: bool,          // True if the migration failed
}
```

## Coordinator V1
For Coordinator V1, the `allowlist` is really a Denylist, the code/types have therefore been named as such. Accounts within the "denylist" should be ignored completely by V1. Because we cannot guarantee the timing of when this "ignore" actually happens, a flag (`v1_ack`) will be set from V1. V2 will only take over once this flag has been set.

Accounts within the "denylist" will be filtered out of the in-memory registry. Any new indexer "registrations" will also be ignored. In-progress historical backfills haven't been considered as we'll disable this functionality anyway.

## Coordinator V2
Once acknowledged by V1, Coordinator V2 will attempt to migrate all functions under the relevant account. The steps for migration are:
1. Remove the streams from the Redis `streams` set - preventing Runner from starting these indexers implicitly
2. Stop the existing V1 executors which have already been started via the `streams` set
3. Merge the existing historical (if it exists) and real-time streams

Once migrated, accounts which have `v1_ack && migrated && !failed` will be exposed to the control loop, prompting V2 to act on these indexers.

### `migrated` flag
For now, the `migrated` flag will not be set on success, preventing V2 from running the indexer on the new architecture. There are some issues around V2 continuing from the right block correctly, so it's best to not run them for now. This allows us to test the migration in isolation, not worrying about what V2 does after that. I'll add this logic back in once https://github.com/near/queryapi/issues/536 is complete.

### `failed` flag
If any part of the migration fails, the `failed` flag will be set for that account. It would take a significant amount of time to cover all the edge cases in code so it would be faster to just set this flag to ignore the account, fix the migration manually and then reset the `failed` flag.


indexer/storage/src/lib.rs
```diff
@@ -198,3 +203,24 @@ pub async fn get_last_indexed_block(
         .query_async(&mut redis_connection_manager.clone())
         .await?)
 }
+
+// `redis::transaction`s currently don't work with async connections, so we have to create a _new_
+// blocking connection to atmoically update a value.
+pub fn atomic_update<K, O, N, F>(redis_url: &str, key: K, update_fn: F) -> anyhow::Result<()>
```
morgsmccauley, line 209: This isn't ideal - blocking code blocks all async code, but given this happens infrequently, and is only temporary it should be ok.

coordinator/src/migration.rs
```diff
@@ -0,0 +1,475 @@
...
+
+    use std::collections::HashMap;
+
+    use mockall::predicate;
+    use registry_types::{IndexerRule, IndexerRuleKind, MatchingRule, Status};
+
+    use crate::registry::IndexerConfig;
+
+    #[tokio::test]
+    async fn ignores_migrated_indexers() {
```
morgsmccauley, line 275: Could definitely expand on these cases but as Redis is mocked it wouldn't provide much benefit, the time would be better spent testing in Dev.

indexer/queryapi_coordinator/src/main.rs
```diff
@@ -38,8 +40,17 @@ pub(crate) struct QueryApiContext<'a> {
     pub redis_connection_manager: &'a ConnectionManager,
     pub indexer_registry: &'a SharedIndexerRegistry,
     pub streamers: &'a Streamers,
+    pub redis_url: &'a str,
 }
 
+#[derive(serde::Deserialize, serde::Serialize, Debug)]
+struct DenylistEntry {
+    account_id: AccountId,
+    v1_ack: bool,
```
morgsmccauley, line 49: Haven't included `migrated`/`failed` because they're not needed - `serde` will just ignore them.

indexer/queryapi_coordinator/src/main.rs
```diff
@@ -118,14 +137,85 @@ async fn main() -> anyhow::Result<()> {
...
+    redis_url: &str,
+) -> IndexerRegistry {
+    let mut filtered_registry = HashMap::new();
+
+    for (account_id, indexer) in indexer_registry.into_iter() {
+        let account_in_deny_list = denylist.iter().find(|entry| entry.account_id == account_id);
+
+        match account_in_deny_list {
+            Some(account_in_deny_list) => {
+                tracing::info!(
```
darunrs, line 184: This will be printed on every ignored indexer for every block coordinator publishes right? This would probably clog up the logs. Maybe its better to collect either ignored or not-ignored indexers in a list and print it all together at the end, instead.

darunrs (approved):
LGTM! Really cool work on this!

indexer/queryapi_coordinator/src/main.rs
morgsmccauley: No - because the registry is mutated as we go, only "new" removals will be logged

PR #529, opened 2024-01-27
fix: Properly handle case sensitive and quoted columns

### LINKED ISSUE #447: Indexers with quoted case-sensitive column names will have context.db calls fail against them

INSERT INTO nearpavel_near_staking_v3."receipts" (**blockHeight**, blockHash, blockTimestamp, blockDate, depositIn, receiptId, predecessorId, signerId, validatorId) VALUES ('106896621', '8o26GZfSt4z7TrbHdHhDCXttXkG3hUFNkLCL33y52c1d', '1701388833845.7158', '2023-12-01', '900000000000000000000000', 'AZhuyEwThxEHUyUcudiNJ49pBhMekgjgokj9tmfHEgkg', '2ee9129228b532b25181ec6cbd1e09bf058e5169c9a9606fec69b5f635423d97', '2ee9129228b532b25181ec6cbd1e09bf058e5169c9a9606fec69b5f635423d97', 'astro-stakers.poolv1.near') ON CONFLICT (receiptId) DO UPDATE SET depositIn = excluded.depositIn RETURNING *

column \"blockheight\" of relation \"receipts\" does not exist

darunrs:
The current implementation of context.db made an assumption that since unquoted identifiers are non-case sensitive, they would correctly apply to quoted versions as well. However, this was found out to be false. If the schema defines a table under posts, and has a column called `"blockheight"`, executing an SQL statement with `blockheight` will not work as it does not match `"blockheight"`. 

In order for quotes in the DDL schema to be supported, context.db needs to maintain whether or not the identifier was quoted. The library used to generate the AST, and get table/column names does not preserve quotes in the name (Though there is an [issue](https://github.com/taozhi8833998/node-sql-parser/issues/1607) for that feature). 

The new approach is as follows:

- Continue to build AST, which gives back unquoted identifiers
- Use regex to split schema into an array of CREATE TABLE statements. They are detected as starting with CREATE TABLE and a semicolon. 
- Using both AST and string statements, find the relevant identifier and capture it as is, with quotes if they are present using another regex. I advance the index in the statement I am searching as well, to minimize edge cases related to mismatches. 
- Store the mappings and deliver data to context.db functions and DmlHandler.

The additional process takes roughly an extra 0.6 to 0.7ms. 

Eventually, we can migrate this work to outside of runFunctions once Control Plane is deployed, as we can expect that code and schema will remain unchanged unless an event occurs (restart).


darunrs:
THe changes require an integration test, by creating some local indexers with edge case schemas. Will do that when possible and comment result and address any gaps, if there are any. I believe the unit tests capture the additional cases well and exisitng tests pass with minimal corrections.

darunrs:
Honestly don't like this approach and would rather wait for support from, or contribute a change to, the node-sql-parser dependency. Will close this.

PR #544, opened 2024-01-30
fix: Prevent Coordinator from stopping V1 executors

### LINKED ISSUE #543: Coordinator not respecting `allowlist`
After starting Coordinator V2 for the first time, we observed lots of "Stop Executor" requests. With the empty `allowlist` in place, we should not be stopping any executors.

morgsmccauley:
As V1 executors are returned via Runner gRPC, Coordinator V2 will attempt to synchronise them. These executors should be completely ignored, within the synchronisation logic.

### DIFF
--- a/coordinator/src/block_streams_handler.rs
+++ b/coordinator/src/block_streams_handler.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(test, allow(dead_code))]
+
 use anyhow::Context;
 use block_streamer::block_streamer_client::BlockStreamerClient;
 use block_streamer::{
--- a/runner/src/server/runner-service.ts
+++ b/runner/src/server/runner-service.ts
@@ -102,7 +102,7 @@ function getRunnerService (executors: Map<string, StreamHandler>, StreamHandlerT
             config = {
               account_id: accountId,
               function_name: functionName,
-              version: -1, // Ensure Coordinator V2 sees version mismatch
+              version: 0, // Ensure Coordinator V2 sees version mismatch
               code: '',
               schema: '',
             };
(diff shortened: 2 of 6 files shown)


coordinator/src/block_streams_handler.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(test, allow(dead_code))]
```
morgsmccauley, line 1: Since the `impl` is mocked in `test` builds, the actual implementation warns `dead_code`, as it isn't used with the tests. This suppresses these errors for `test` builds, code that is not used in normal builds will still warn.

runner/src/server/runner-service.ts
```diff
@@ -102,7 +102,7 @@ function getRunnerService (executors: Map<string, StreamHandler>, StreamHandlerT
             config = {
               account_id: accountId,
               function_name: functionName,
-              version: -1, // Ensure Coordinator V2 sees version mismatch
```
morgsmccauley, line 105: As `version` is `uint64` this ends up being a the largest possible 64 bit int.

darunrs (approved):
The changes look good to me! 

We still need to add an allowlist block on Block Streamer, and also add logs to it (For when it successfully completes a start or stop command). We could do that in a separate PR but we just gotta be careful as once we deploy this PR and  the triggers run, V2 coordinator will go and kick off every block streamer it can, which spikes Runner/Redis load. We would want to either implement the block streamer allowlist in this PR too, or be ready to action on dev. I'll leave it up to you!

Also, I haven't actually flushed the streams yet. Do you mind doing it? Sorry!

morgsmccauley:
@darunrs Implementing the `allowlist` in Block Streamer isn't necessary - Coordinator will not attempt to manage any indexers which aren't in the `allowlist`. The only reason this change here is necessary is because the V1 executors are started (via the redis set) and Coordinator believes they shouldn't exist.

I'll add logging in a follow up PR :)

> Also, I haven't actually flushed the streams yet. Do you mind doing it? Sorry!
Yup I can do it :)

PR #547, opened 2024-02-02
fix: Executors would crash when DmlHandler.create times out

### LINKED ISSUE #532: Indexers occasionally crash due to timeout on Hasura credentials fetch when creating DmlHandler
DmlHandler is lazy loaded before running user's indexer code. The problem is part of the class creation involves polling Hasura for credentials for the user's DB. These are used to then initialize a postgres client. The hasura call can fail due to timeout. If this happens, the error is stored inside the promise until the DML Handler is awaited. This happens when context.db is next called and this call is not within a try catch. The result is that the error is not caught and bubbles up. Occasionally, this could happen at an unexpected part of code which causes the error to somehow travel all the way up to the stream handler, crashing it. 

We can fix the issue in two phases:

Phase 1 (Immediate patch)
Add a catch for await dmlHandler's create promise to handle if the timeout error occurred. We can simply fail for now and let Runner retry as it normally would. 

Phase 2 (Integration into Control Plane)
Migrate the Hasura call outside of runFunction call path as part of https://github.com/near/queryapi/issues/426. We can do this because when the executor is started, we can expect Coordinator to let Runner know of provisioning status, and we don't have to do guess work on if Hasura even has user DB credentials or not. 

In addition, moving the Hasura calls outside of runFunctions will improve latency of calls, speeding up backfill performance. Actual impact of migrating the calls out will be known once Instrumentation is completed and released.

darunrs:
In each invocation of runFunction, on a block, DML Handler is created. It's responsible for making calls to the database. Part of setting it up involves fetching the credentials for the user's DB from Hasura, and creating a PG Client. This takes time so the process was run through an unawaited async request. While other setup and some of user code is ran, the setting up of DML Handler would be completed. The first context.db call would await its completion and subsequent calls would have it ready. 

However, it was observed that when the call to Hasura for the DB credentials times out, the error, instead of propagating into a try catch, would instead be considered by the Javascript runtime as an unhandled Promise Exception, and would terminate the worker thread, stopping the indexer. 

In order to fix this problem, we need to transition away from keeping DmlHandler.create as an unresolved Promise across multiple contexts. The approach I've decided to take is to defer the creation of the Hasura call promise until the first call of context.db. This adds more latency to the first context.db call as it now must wait for the entire process to complete. However, this also does not penalize Indexers that don't use context.db as their code does not need to connect to Hasura unless needed. 

Very soon, we will in fact overhaul this existing logic by migrating the Hasura credentials call away from runFunctions. This eliminates the underlying problem of unresolved promises as none remain afterward. So, the focus here is to address the bug, which is a critical problem, without too much change, as the workflow will be refactored again soon anyway. 

I also fix a small bug where context.db calls were getting logged under the wrong indexer logs table function name.


runner/src/indexer/indexer.ts
```diff
@@ -298,7 +287,8 @@ export default class Indexer {
       }, {});
       return result;
     } catch (error) {
-      console.warn('Caught error when generating context.db methods. Building no functions. You can still use other context object methods.\n', error);
+      const errorContent = error as Error;
+      console.warn('Caught error when generating context.db methods. Building no functions. You can still use other context object methods.', errorContent.message);
```
darunrs, line 291: Schemas that don't adhere to the limitations of context.db emit a very verbose error message. That message is helpful in the frontend when you can change the schema, but not in the backend. Returning just the message prevents GCP logs from looking cluttered and also reduces log congestion.

runner/src/dml-handler/dml-handler.ts
```diff
@@ -4,29 +4,50 @@ import HasuraClient from '../hasura-client/hasura-client';
...
+
+  async initialize (): Promise<PgClientModule> {
+    if (!this.getPgClientPromise) {
+      this.getPgClientPromise = this.getPgClient();
+    }
+    try {
+      return await this.getPgClientPromise;
+    } catch (e) {
+      this.initializeFailure = true;
+      throw e;
```
darunrs, line 32: This error is now thrown into the user's context. Their code controls if it is re-thrown, swallowed, logged, etc.

runner/src/indexer/indexer.ts
```diff
@@ -230,60 +235,44 @@ export default class Indexer {
...
         const funcForTable = {
           [`${sanitizedTableName}`]: {
             insert: async (objectsToInsert: any) => {
               // Write log before calling insert
-              await this.writeLog(`context.db.${sanitizedTableName}.insert`, blockHeight, defaultLog + '.insert',
-                `Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName} on schema ${schemaName}`);
-
-              // Wait for initialiiation of DmlHandler
-              const dmlHandler = await dmlHandlerLazyLoader;
+              await this.writeLog(functionName, blockHeight,
```
darunrs, line 242: Replaced incorrect functionName with the correct one. Context.db logs will now appear in the user's indexer logs.

runner/src/dml-handler/dml-handler.ts
```diff
@@ -4,29 +4,44 @@ import HasuraClient from '../hasura-client/hasura-client';
...
+  }
+
+  async initialize (): Promise<PgClientModule> {
+    if (!this.getPgClientPromise) {
+      this.getPgClientPromise = this.getPgClient();
+    }
+    return await this.getPgClientPromise;
+  }
+
+  async getPgClient (): Promise<PgClientModule> {
```
darunrs, line 30: If there is an error in this function, it is now thrown into the user's context. Their code controls if it is re-thrown, swallowed, logged, etc.

gabehamilton (approved):

PR #548, opened 2024-02-05
feat: Add "StartBlock::Continue" to registry contract/types

### LINKED ISSUE #536: Expand start block types to include "Continue"
Indexers should be able to continue indexing from the last indexed block, prior to update.

### LINKED ISSUE #539: Update the contract to allow only admins to use this type

morgsmccauley:
The start block parameter (`start_block_height`) has two states, signifying "Start from block" and "Start from latest" only. With the removal of the continuous "real-time" process, which essentially runs Indexers "From Interruption", the need for a third state arises, representing "Interruption" arises. This PR expands the registry contract/types to accommodate this change. This work was a good opportunity to refactor the existing types, removing unnecessary fields and tailoring it to the new architecture.

## Registry Types Refactoring
- `filter: IndexerRule` -> `rule: Rule`: `IndexerRule` contained a lot of noise/data which was not needed. `id`, `name`, and `indexer_rule_kind` have all been removed. The only useful part here is `MatchingRule` which has been renamed to just `Rule`.
- `schema` is now non-optional: This field is always required so it makes sense to convey that in the code. Having it as an `Option` created lots of unnecessary checks throughout the system.
- `start_block_height` replaced by `start_block`: The latter being an `enum` to represent "From Latest", "From Block", and "From Interruption"

## Public methods/API
To minimise the disruption of the existing contract consumers, the public API remains unchanged, i.e. the core methods (`list_indexer_functions`, `register_indexer_function`, etc.) use/return the same types. As the underlying data will be migrated to the new types, these methods infer the old types from the new ones. New methods have been created to work with the new types (`register`, `list_all`, etc.) allowing clients to move over when possible, as opposed to creating a breaking change.

### DIFF
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -294,7 +294,7 @@ mod tests {
             )
             .unwrap(),
             function_name: "test".to_string(),
-            indexer_rule: registry_types::IndexerRule {
+            indexer_rule: registry_types::OldIndexerRule {
                 indexer_rule_kind: registry_types::IndexerRuleKind::Action,
                 matching_rule: registry_types::MatchingRule::ActionAny {
                     affected_account_id: "queryapi.dataplatform.near".to_string(),
--- a/coordinator/src/main.rs
+++ b/coordinator/src/main.rs
@@ -255,7 +255,7 @@ mod tests {
     use mockall::predicate;
     use std::collections::HashMap;
 
-    use registry_types::{IndexerRule, IndexerRuleKind, MatchingRule, Status};
+    use registry_types::{IndexerRuleKind, MatchingRule, OldIndexerRule as IndexerRule, Status};
 
     use crate::registry::IndexerConfig;
 
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -268,6 +268,22 @@ impl Contract {
             &account_id
         );
 
+        match &rule {
+            Rule::ActionAny {
+                affected_account_id,
+                ..
+            }
+            | Rule::ActionFunctionCall {
+                affected_account_id,
+                ..
+            } => {
+                if affected_account_id == "*" {
+                    self.assert_roles(vec![Role::Owner]);
+                }
+            }
+            _ => {}
+        }
+
         let account_indexers =
             self.registry
                 .entry(account_id.clone())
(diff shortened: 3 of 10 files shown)


block-streamer/src/block_stream.rs
```diff
@@ -294,7 +294,7 @@ mod tests {
             )
             .unwrap(),
             function_name: "test".to_string(),
-            indexer_rule: registry_types::IndexerRule {
+            indexer_rule: registry_types::OldIndexerRule {
```
morgsmccauley, line 297: Will be migrated to new types in a future PR

coordinator/src/main.rs
```diff
@@ -274,7 +274,7 @@ mod tests {
     use mockall::predicate;
     use std::collections::HashMap;
 
-    use registry_types::{IndexerRule, IndexerRuleKind, MatchingRule, Status};
+    use registry_types::{IndexerRuleKind, MatchingRule, OldIndexerRule as IndexerRule, Status};
```
morgsmccauley, line 277: Will be migrated to new types in a future PR

registry/contract/src/lib.rs
```diff
@@ -5,50 +5,45 @@ use near_sdk::store::UnorderedMap;
...
-pub enum StorageKeys {
+enum StorageKeys {
     Registry,            // can be removed after migration
     Account(CryptoHash), // can be removed after migration
     RegistryV1,
     AccountV1(CryptoHash),
     RegistryV2,
     AccountV2(CryptoHash),
+    RegistryV3,
+    AccountV3(CryptoHash),
```
darunrs, line 46: What's the purpose of these? I'm a bit confused how these are used in conjunction with various types like IndexersByAccount.

darunrs (approved):
I assume we're going forward with Update, or some version of that instead of FromInterruption? 

Otherwise it looks good to me!

Seems like the approach was to create new functions to avoid having to build in backwards compatibility and bloat existing code further, migrate our services over to new functions, and then return and clean up the old code?

registry/contract/src/lib.rs
morgsmccauley: I don't completely understand, but afaik, we need to use a unique storage key for each collection we store. Re-using an existing key could lead us to read stale state and therefore fail deserialization. This is essentially a helper for creating unique keys.

In saying that, the migration actually ignores existing state so it's probably ok to ignore these, but I kept doing it just incase 😅

morgsmccauley:
> I assume we're going forward with Update, or some version of that instead of FromInterruption?
Yup, I'll rename to `Continue`

> Seems like the approach was to create new functions to avoid having to build in backwards compatibility and bloat existing code further, migrate our services over to new functions, and then return and clean up the old code?
Yes exactly, V2 will use the new functions, while V1 still uses the existing. I'll then remove the existing functions once V2 is live.

morgsmccauley:
This has been deployed to dev, will let it sit for a bit and then to prod

PR #550, opened 2024-02-06
fix: Crashed Runner Executors would continue to display RUNNING

### LINKED ISSUE #533: Indexer Status frozen as RUNNING if unexpectedly crashes
Indexer Status remains as RUNNING if the executor in Runner crashes in the middle of runFunction invocation. This is because the status is set to STOPPED only if the runFunction invocation fails and its error is caught. So, if the error was somehow not caught and the error was thrown at the Executor class level, it will fail there without any error handling, crashing the worker thread. 

To resolve this issue, we need to ensure that StreamHandler is able to write the status and blockheight to postgres if it detects the worker thread failed.

darunrs:
Runner sets the status of an Indexer after a successful run or a failed run whose error was caught. However, if the executor itself crashed, the error would not be caught as the worker is terminated. As a result, the status of the indexer would continue to display RUNNING, which was incorrect and misleading. 

I updated the status by ensuring crashed workers have the STOPPED status set. In addition, I added a new FAILING status which will now be set when an indexer is still running, but failing on the same block.

### DIFF
--- a/runner/src/stream-handler/stream-handler.ts
+++ b/runner/src/stream-handler/stream-handler.ts
@@ -2,9 +2,11 @@ import path from 'path';
 import { Worker, isMainThread } from 'worker_threads';
 
 import { registerWorkerMetrics } from '../metrics';
+import Indexer from '../indexer';
 
 export enum Status {
   RUNNING = 'RUNNING',
+  FAILING = 'FAILING',
   STOPPED = 'STOPPED',
 }
 export interface IndexerConfig {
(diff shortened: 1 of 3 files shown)


runner/src/stream-handler/stream-handler.ts
```diff
@@ -54,6 +56,11 @@ export default class StreamHandler {
     if (this.indexerConfig !== undefined) {
       this.executorContext.status = Status.STOPPED;
     }
+    const indexer = new Indexer();
```
darunrs, line 59: Converting the setStatus function to static would have required more changes than necessary. Plus, creating an Indexer class is very cheap. And I was hoping to consider refactoring the set status, write logs, etc. all at once. Will discuss offline in team meeting.

gabehamilton (approved):

PR #552, opened 2024-02-06
fix: Various Control Plane migration fixes

### LINKED ISSUE #520: Migrate Indexers to Control Plane

morgsmccauley:


coordinator/src/redis.rs
```diff
@@ -34,16 +34,17 @@ impl RedisClientImpl {
         })
     }
 
-    pub async fn get<T, U>(&self, key: T) -> anyhow::Result<U>
+    pub async fn get<T, U>(&self, key: T) -> anyhow::Result<Option<U>>
```
morgsmccauley, line 37: Not actually directly related, but quite a nice change to have

indexer/queryapi_coordinator/src/main.rs
```diff
@@ -43,10 +43,14 @@ pub(crate) struct QueryApiContext<'a> {
     pub redis_url: &'a str,
 }
 
+// TODO add missing keys to prevent deserialization error on v2 side
 #[derive(serde::Deserialize, serde::Serialize, Debug)]
 struct DenylistEntry {
     account_id: AccountId,
     v1_ack: bool,
+    migrated: bool,
```
morgsmccauley, line 51: This can still be deserialized without these fields, but it means they will be removed when serializing again.

coordinator/src/migration.rs
```diff
@@ -104,8 +113,7 @@ async fn migrate_account(
             .context("Failed to merge streams")?;
     }
 
-    // TODO Uncomment when V2 correctly continues from V1 stop point
-    // set_migrated_flag(redis_client, account_id)?;
+    set_migrated_flag(redis_client, account_id)?;
```
morgsmccauley, line 116: We need to set the `migrated` flag to prevent further attempts to migrate

coordinator/src/migration.rs
```diff
@@ -14,12 +14,17 @@ pub struct AllowlistEntry {
     v1_ack: bool,
     migrated: bool,
     failed: bool,
+    v2_control: bool,
```
morgsmccauley, line 17: Since we have to set the `migrated` flag, this allows us to still prevent V2 from taking control. Providing time to sanity check the migration itself.

morgsmccauley:
Going to merge these to unblock testing - feel free to review in retrospect

PR #553, opened 2024-02-07
feat: Handle `StartBlock` options within Coordinator & Block Streamer

### LINKED ISSUE #521: Handle new `StartBlock` config options
Context: https://github.com/near/queryapi/pull/517#discussion_r1462362712

We should start using the updated `StartFrom` options as opposed to the `start_from_block` parameter.

morgsmccauley:
This PR adds support for the updated Registry types, including the new `StartBlock` config option, across the Block Streamer/Coordinator.

With `StartBlock`, the logic within Coordinator is much more straight forward - we no longer need to guess whether we should "continue" or "start over", that is build in to the configuration options.

This only really affects the handling of Block Streams, the executor flow remains the same: always restart when a new version is published.

To summarise the Block Stream synchronisation process:
- `StartBlock::Continue` - Resumes process, keeping the current data in the Redis Stream, and start the Block Stream from `last_published_block`
- `StartBlock::Latest` - Starts a new process, clears the Redis Stream, starts the Block Stream from the registry version, essentially being latest
- `StartBlock::Height(u64)` - Starts a new process, clears the Redis Stream, starts the Block Stream from the height configured

Additionally, Accounts/Indexers which have just been migrated, and also streams which have been stopped but have unchanged versions (e.g. after Block Streamer restart), will be treated the same as `StartBlock::Continue`.


block-streamer/src/redis.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg_attr(test, allow(dead_code))]
```
morgsmccauley, line 1: Cargo thinks that the real implementation is "unused", and the mock version is used instead. This suppresses those warnings.

coordinator/src/block_streams/synchronise.rs
```diff
@@ -0,0 +1,598 @@
+use registry_types::StartBlock;
```
morgsmccauley, line 1: Moved synchronisation logic our or `main.rs` and in to its own mod

coordinator/src/block_streams/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod handler;
```
morgsmccauley, line 1: Moved synchronisation logic out of `main.rs` and it to its own mod. Moved the "handler" struct next to it too.

coordinator/src/block_streams/synchronise.rs
```diff
@@ -0,0 +1,598 @@
...
+
+        block_streams_handler
+            .stop(unregistered_block_stream.stream_id)
+            .await?;
+    }
+
+    Ok(())
+}
+
+#[tracing::instrument(
```
morgsmccauley, line 60: Creates a `tracing::span` which automatically adds this context to all log messages called within/below this function.

block-streamer/src/block_stream.rs
```diff
@@ -123,20 +132,59 @@ pub(crate) async fn start_block_stream(
...
         start_block_height,
-        "Starting block stream",
+        delta_lake_client,
+        redis_client.clone(),
+        indexer,
+        redis_stream.clone(),
+    )
+    .await
+    .unwrap_or_else(|err| {
+        tracing::warn!(
```
morgsmccauley, line 146: There's currently an issue with the Delta Lake pipeline - which highlighted that any failures here prevent kill the entire block stream. Changed it to capture/log these errors, and then just continue with Near Lake Framework. Not ideal, but better than just stopping.

coordinator/src/block_streams/synchronise.rs
```diff
@@ -0,0 +1,599 @@
...
+                redis_client,
+                block_streams_handler,
+            )
+            .await
+            .map_err(|err| {
+                tracing::error!(
+                    account_id = account_id.as_str(),
+                    function_name,
+                    version = indexer_config.get_registry_version(),
+                    "failed to sync block stream: {err:?}"
```
morgsmccauley, line 38: Swallow errors, they will be retired on the next control loop

coordinator/src/migration.rs
```diff
@@ -230,6 +239,20 @@ async fn merge_streams(
     }
 }
 
+async fn update_stream_version(
```
morgsmccauley, line 242: We need to distinguish between "new" and "migrated" streams, as they are handled differently

coordinator/src/block_streams/synchronise.rs
```diff
@@ -0,0 +1,676 @@
...
+}
+
+async fn get_continuation_block_height(
+    indexer_config: &IndexerConfig,
+    redis_client: &RedisClient,
+) -> anyhow::Result<u64> {
+    redis_client
+        .get_last_published_block(indexer_config)
+        .await?
+        .ok_or(anyhow::anyhow!("Indexer has no `last_published_block`"))
```
darunrs, line 192: What does this do? Is our intended behavior to return some other block height available to us? Like the version?

coordinator/src/executors/synchronise.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    for (account_id, indexers) in indexer_registry.iter() {
+        for (function_name, indexer_config) in indexers.iter() {
+            let active_executor = active_executors
+                .iter()
+                .position(|stream| {
+                    stream.account_id == account_id.to_string()
+                        && &stream.function_name == function_name
+                })
+                .map(|index| active_executors.swap_remove(index));
```
darunrs, line 28: Now that I think about, what's the behavior if the account and function name is the same for multiple indexers? Does swap_remove stop at the first instance?

darunrs (approved):
Awesome work on the migration! LGTM! Just some small comments.

coordinator/src/block_streams/synchronise.rs
morgsmccauley: This fails the block stream creation. I didn't want to make any assumptions about where to start the block stream from as we may be incorrect, resulting in undesired behaviour 🤔

coordinator/src/executors/synchronise.rs
morgsmccauley: `.position` finds the _first_ occurrence , so yes, this will take the first executor. Runner prevents duplicate executors thought right? In saying that though we should probably handle the case here.

darunrs: Runner sort of does it through the deterministic hashing. I agree that it'd be nicer to have that called out in Coordinator since it would indicate an issue with our contract. And we can verify it here, instead of relying on our hashing mechanism in Runner, which might change.

coordinator/src/block_streams/synchronise.rs
darunrs: That makes sense to me!

PR #556, opened 2024-02-13
fix: Various block streamer issues

morgsmccauley:


block-streamer/src/block_stream.rs
```diff
@@ -123,17 +132,52 @@ pub(crate) async fn start_block_stream(
...
+) -> anyhow::Result<u64> {
     let latest_block_metadata = delta_lake_client.get_latest_block_metadata().await?;
-    let last_indexed_block = latest_block_metadata
+    let last_indexed_block_from_metadata = latest_block_metadata
         .last_indexed_block
-        .parse::<near_indexer_primitives::types::BlockHeight>()?;
+        .parse::<near_indexer_primitives::types::BlockHeight>()
+        .context("Failed to parse Delta Lake metadata")?;
+
+    if start_block_height >= last_indexed_block_from_metadata {
```
morgsmccauley, line 178: This is essentially for `StartBlock::Latest`, we dont need to check delta lake in this case.

block-streamer/src/delta_lake_client.rs
```diff
@@ -246,6 +246,7 @@ impl DeltaLakeClientImpl {
                 }
             })
             .flat_map(|index_file| index_file.heights)
+            .filter(|block_height| *block_height >= start_block_height)
```
morgsmccauley, line 249: Previously we would add all heights for the day the height falls within. This was especially an issue for `StartBlock::Latest` as we would add many blocks initially where there should be essentially none.

block-streamer/src/block_stream.rs
```diff
@@ -123,17 +132,52 @@ pub(crate) async fn start_block_stream(
...
+    )
+    .await?;
+
+    let last_indexed_near_lake_block = process_near_lake_blocks(
+        last_indexed_delta_lake_block,
+        lake_s3_config,
+        lake_prefetch_size,
+        redis_client,
+        indexer,
+        redis_stream,
```
darunrs, line 152: So previously in delta lake, we passed a clone of this. Here we pass it directly. Is the reason that we end the function afterward, and are ok with it consuming the ownership or the function, and the end of this function releases it? 

Mainly just trying to understand Rust best practices.

darunrs (approved):
Awesome work! I like the refactoring of the block processing.

block-streamer/src/block_stream.rs
morgsmccauley: Yes, it's fine to take ownership here because `redis_stream` is no longer used from here

PR #561, opened 2024-02-14
Prod Release 15/02/24

morgsmccauley:
- fix: Various Control Plane migration fixes (#552)
- feat: Add "StartBlock::Continue" to registry contract/types (#548)
- fix: Labels showing up in Grafana and Composed Runner fails due to Region Missing
- feat: Log Worker crashes in Indexer logs
- feat: Add metrics for memory footprint
- fix: Set current block height for listed V1 indexers and fix rust formatting
- feat: Remove executor metrics on stop (#557)
- fix: Prevent skipping blocks from Redis Stream (#558)
- feat: Handle `StartBlock` options within Coordinator & Block Streamer (#553)
- fix: Various block streamer issues (#556)


indexer/queryapi_coordinator/src/main.rs
```diff
@@ -138,8 +141,9 @@ async fn main() -> anyhow::Result<()> {
 }
 
 async fn fetch_denylist(redis_connection_manager: &ConnectionManager) -> anyhow::Result<Denylist> {
-    let raw_denylist: String =
-        storage::get(redis_connection_manager, storage::DENYLIST_KEY).await?;
+    let raw_denylist: String = storage::get(redis_connection_manager, storage::DENYLIST_KEY)
+        .await
+        .unwrap_or("".to_owned());
```
darunrs, line 146: This should be returning "[]" in order to not cause problems in fact. I can either push a quick PR to merge that change or we can let it through, since it isn't a problem if there's a denylist anyway (You can add it in redis before merging).

registry/contract/src/lib.rs
```diff
@@ -5,50 +5,45 @@ use near_sdk::store::UnorderedMap;
...
 };
 
 type FunctionName = String;
-// Define the contract structure
-#[near_bindgen]
+
 #[derive(BorshDeserialize, BorshSerialize, Debug)]
-pub struct Contract {
-    registry: IndexersByAccount,
+pub struct OldContract {
```
darunrs, line 15: Have you deployed the updates to prod's registry already?

darunrs (approved):
I think it all looks good to me. Two comments on the changes before we send them out.

indexer/queryapi_coordinator/src/main.rs
morgsmccauley: `[]` has already been added to Redis :)

registry/contract/src/lib.rs
morgsmccauley: Yup 👍🏼

PR #562, opened 2024-02-15
feat: Improve Schema Error Diplay with Glyphs

Kevin101Zhang:
Accidentally overwrote some changes here - https://github.com/near/queryapi/pull/555.
Checkout a branch a resolved the latest changes for removing glyphsuccess color from green to transparent and removing trycatch.

Glyphs margins to showcase a better error message to the user. The Glyph is either red or green. Green indicating it has not caught an errors and red indicating it has found an error. When an error is found it will come with a hoverable error message as well as the line it was caught on.

Procedure:
Onload of the schema page the glyphs will mount and check if there is an error based off the users previously written schema or their locally stored code.

If the user has an error: It will highlight the line the error was caught on with a red glyph and a hoverable message.
If the user does not have an error it will remain on line 1 with a green glyph. Onhover it will tell the user there are no errors.

If the user navigates away from the schema tab with an error it will prompt the user with a Alert that there was an error and to please return back to the schema tab and correct the schema.

The Glyph functionality with run on a debounced event of 500ms based off the users onChange typing activity on the schema page. When triggered it will prettify the users schema and will attempt to generate types in a try/catch. If it hits the error then it will relay the line number the error was caught on and the err message which the glyph will represent.

The Glyph only takes place on schema.sql as for indexLogic we have monaco's support for intellisense/catching errors in the JS env.

In the case in the video where the glyph is on line one and hovering over the glyph showcases an error message with a line number - this is when prettier is throwing an error and can be added through string matching to correctly display the glyph location

https://www.loom.com/share/78d0c6435115470597b9e8448976fda4?sid=fba10322-2ded-4b14-8178-8a5b51143148


darunrs (approved):
LGTM! Copy over the description from the previous PR, and do a squash and merge.

darunrs:
Name the PR something like: `feat: Improve Schema Error with Glyphs` or something too.

PR #567, opened 2024-02-16
feat: Support Wildcard and * contract filters

### LINKED ISSUE #541: Update the frontend to dis-/allow use of this filter

Kevin101Zhang:
added the * filter for contracts.
A valid contract is defined as 

```
function validateContractId(accountId) {
  accountId = accountId.trim();
  
  const isLengthValid = accountId.length >= 2 && accountId.length <= 64;
  if (!isLengthValid) return false; 

  //test if the string starts with a '*.' and remove it if it does
  const isWildCard = WILD_CARD_REGEX.test(accountId);
  isWildCard ? accountId = accountId.slice(2) : null;

  //test if rest of string is valid accounting for/not isWildCard
  const isRegexValid = CONTRACT_NAME_REGEX.test(accountId);
  return isRegexValid;
}

note: early termination for performance
```

The function take in account all **invalid** contract names on [nomicon](https://nomicon.io/DataStructures/Account)
Please view test for more details.

Regex Broken down Below:

CONTRACT_NAME_REGEX:

`/^(([a-z\d]+[-_])*[a-z\d]+(\.([a-z\d]+[-_])*[a-z\d]+)*\.([a-z\d]+)|([a-z\d]+))$/`

1. **^**: Asserts the start of the string.

2. **(**: Begins a capturing group.
3. **([a-z\d]+[-_])***: Matches a sequence of lowercase letters or digits, possibly followed by hyphens or underscores. The sequence may repeat zero or more times.
4. **[a-z\d]+**: Matches a sequence of one or more lowercase letters or digits.
5. **(\.([a-z\d]+[-_])*[a-z\d]+)***: Allows for zero or more occurrences of a dot . followed by another sequence.
6. **\.([a-z\d]+)**: Matches a dot followed by a sequence of lowercase letters or digits.
7. **|**: OR operator.
8. **([a-z\d]+)**: Matches a standalone sequence of lowercase letters or digits.
9. **)$**: Asserts the end of the string.
WILD_CARD_REGEX:

`/\*\./`
1. **\***: Matches a literal asterisk.
2. **\.**: Matches a literal dot.

EDIT: Added more test please review
Added jest to support function testing.

It seems there's a problem getting Babel and the VM to work smoothly together. I'm going to dig deeper into it. Right now, I'm adjusting the test file to recreate the functions instead of using the usual require/import method, which seems to cause trouble with the VM code. It does mean there's some duplicated code in the testing files, but the test work.


frontend/src/components/Editor/__tests__/validator.test.js
```diff
@@ -0,0 +1,107 @@
+const CONTRACT_NAME_REGEX = /^(?:\*|(?:[a-z\d]+[-_])*[a-z\d]+)(?:\.(?:\*|(?:[a-z\d]+[-_])*[a-z\d]+))*\.near$/i;
+
+function validateContractId(accountId) {
```
darunrs, line 3: You should always import anything you are testing, rather than repeating. The last thing you want is for the thing you intended to test to diverge from the code you copied here.

frontend/src/constants/RegexExp.js
```diff
@@ -1 +1,2 @@
-export const CONTRACT_NAME_REGEX = RegExp(/^(\*|([a-z\d]+[-_])*[a-z\d]+)(\.*(\*|([a-z\d]+[-_])*[a-z\d]+))*\.(\w+)$/);
+export const CONTRACT_NAME_REGEX = /^(?:\*|(?:[a-z\d]+[-_])*[a-z\d]+)(?:\.(?:\*|(?:[a-z\d]+[-_])*[a-z\d]+))*\.near$/i;
```
darunrs, line 1: This already supported *.near and other things. Let's try to change it as little as possible, since it seems to have already been working. Let's just make the safe change of allowing * in addition to whatever we currently have!

frontend/jest.config.js
```diff
@@ -0,0 +1,15 @@
+// jest.config.js
```
darunrs, line 1: Nice job adding unit tests here!

darunrs (commented):
Some comments on the updated regex, and testing methodology. Great job adding jest here!

frontend/src/components/Editor/__tests__/validator.test.js
```diff
@@ -0,0 +1,107 @@
...
+  const isWildcard = accountId.trim() === "*" || accountId.trim() === "*.near";
+  const isLengthValid = accountId.length >= 2 && accountId.length <= 64;
```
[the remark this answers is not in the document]
darunrs: Found this as a start: https://nomicon.io/DataStructures/Account

frontend/src/components/Editor/__tests__/validator.test.js
Kevin101Zhang: I mentioned in the description there are some issues with Babel and the VM to work smoothly together. Will try again with the import later on

frontend/src/constants/RegexExp.js
Kevin101Zhang: ok!

frontend/src/components/Editor/__tests__/validator.test.js
```diff
@@ -0,0 +1,107 @@
...
+    expect(validateContractIds(spacedIds)).toBe(false);
+  });
+
+  test('it should return true for a mix of valid and wildcard contract IDs', () => {
+    const mixedIds = 'contract1.near, *.near, contract3.near';
+    expect(validateContractIds(mixedIds)).toBe(true);
+  });
+
+  test('it should return false for an invalid wildcard contract ID', () => {
+    const invalidWildcard = '*invalid.near';
```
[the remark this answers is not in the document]
darunrs: Right so as discussed in the meeting, we allow star at the beginning only separated by a `.`. E.g. `*.near`, `*.tg`, `*.some.sub.account` but NOT `some.*.account`, `account.*`, `someone.*` or `*near`. We also allow just `*`.

frontend/src/components/Editor/__tests__/validator.test.js
```diff
@@ -0,0 +1,226 @@
+const CONTRACT_NAME_REGEX = RegExp(/^(([a-z\d]+[-_])*[a-z\d]+(\.([a-z\d]+[-_])*[a-z\d]+)*\.([a-z\d]+)|([a-z\d]+))$/);
```
darunrs, line 1: Can you try adding a babel.config.js to the frontend folder and trying imports again? I feel like that might work? 

Put this in the config:
`module.exports = {presets: ['@babel/preset-env']}`

frontend/src/utils/validators.js
```diff
@@ -1,15 +1,21 @@
...
-    CONTRACT_NAME_REGEX.test(accountId)
-  );
+function validateContractId(accountId) {
+  accountId = accountId.trim();
+  const isLengthValid = accountId.length >= 2 && accountId.length <= 64;
+  if (!isLengthValid) return false;
+
+  //test if the string starts with a '*.' and remove it if it does
+  const isWildCard = WILD_CARD_REGEX.test(accountId);
+  isWildCard ? accountId = accountId.slice(2) : null;
```
darunrs, line 14: Could be re-written as 
`accountId = isWildCard ? accountId.slice(2) : accountId;`

I think having a set operation in the middle is a wee bit awkward.

frontend/src/constants/RegexExp.js
```diff
@@ -1 +1,2 @@
-export const CONTRACT_NAME_REGEX = RegExp(/^(\*|([a-z\d]+[-_])*[a-z\d]+)(\.*(\*|([a-z\d]+[-_])*[a-z\d]+))*\.(\w+)$/);
+export const CONTRACT_NAME_REGEX = RegExp(/^(([a-z\d]+[-_])*[a-z\d]+(\.([a-z\d]+[-_])*[a-z\d]+)*\.([a-z\d]+)|([a-z\d]+))$/);
```
darunrs, line 1: Nice job on the regex!

darunrs (commented):
Mostly looks good to me. 

I just want to clarify once more that we want to support* as that seems correct to me. I checked with Pavel and he confirmed what I thought about supporting `*` alone. 

Also dropped a suggestion for fixing that babel issue. If it works, awesome.  If not, no big deal, we can defer it. Leave a TODO in the test file to replace with imports later.

frontend/src/components/Editor/__tests__/validator.test.js
```diff
@@ -0,0 +1,226 @@
...
+const WILD_CARD_REGEX = RegExp(/\*\./);
+// const INVALID_ACCOUNT = 'system';
+
+function validateContractId(accountId) {
+  accountId = accountId.trim();
+  const isLengthValid = accountId.length >= 2 && accountId.length <= 64;
+  if (!isLengthValid) return false;
+
+  //test if the string starts with a '*.' and remove it if it does
+  const isWildCard = WILD_CARD_REGEX.test(accountId);
```
pkudinov, line 11: I think WILD_CARD_REGEX will match `abc*.near` as well, and we only need it to match `*.a.b.c.near` type of account

frontend/src/components/Editor/__tests__/validator.test.js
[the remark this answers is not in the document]
pkudinov: Indeed, `*` should be a valid filter, as we discussed during the call

pkudinov (dismissed):

frontend/src/components/Editor/__tests__/validator.test.js
darunrs: This is true. We probably need something more like `/^\*./`

frontend/src/components/Editor/__tests__/validator.test.js
Kevin101Zhang: The problem was specifying babel to only run on test as babel fails on our actual production code. Happy to sync to try to resolve this

frontend/src/components/Editor/__tests__/validator.test.js
Kevin101Zhang: WILD_CARD_REGEX is testing if the string starts with '*.'
The code right now suceeds for *.a.b.c.near and fails for abc*.near.

darunrs: Oh no I didn't say we shouldn't. I said we should, even if the validate function would reject it anyway.

darunrs (approved):
Looks good to me. There's a merge conflict but otherwise, should be good. 

Add a comment above the original validate functions that they're duplicated in the test files due to a Babel config issue, and a TODO to address that. 

Also, can you rename the PR to something like `feat: Support Wildcard and * contract filters`?

PR #570, opened 2024-02-20
feat: Remove `allowlist` from Coordinator V2

### LINKED ISSUE #568: Remove `allowlist`

morgsmccauley:
Now that both dev/prod have been completely migrated, we can remove the `allowlist`. This makes the V2 architecture the default.

We don't need to remove the `denylist` from Coordinator V2, that process has already been stopped.


darunrs (approved):
LGTM!

PR #571, opened 2024-02-20
fix: view and monaco glyph interactions with debug and diffviedw

### LINKED ISSUE #569: Glyph view on debug mode
error on diffView/debug mode due to rendering of unmounted glyphs.

Image

Kevin101Zhang:
Logic fix for handling monaco glyph mounting.

Originally it will error on diffView/debug mode due to rendering of unmounted glyphs.
Screenshot 2024-02-20 at 3 05 05 PM

Video post fix:
https://www.loom.com/share/dd797612590d4f12a27e7216ab009b68?sid=4ec944fd-e718-4851-b231-4339ffab5bd8

### DIFF
--- a/frontend/src/components/Editor/Editor.jsx
+++ b/frontend/src/components/Editor/Editor.jsx
@@ -364,11 +364,7 @@ const Editor = ({ actionButtonText }) => {
         [
           {
             range: new monaco.Range(1, 1, 1, 1),
-            options: {
-              isWholeLine: true,
-              glyphMarginClassName: "glyphSuccess",
-              glyphMarginHoverMessage: { value: "" },
-            },
+            options: {},
           },
         ]
       );
--- a/frontend/src/components/Editor/Editor.jsx
+++ b/frontend/src/components/Editor/Editor.jsx
@@ -359,24 +359,27 @@ const Editor = ({ actionButtonText }) => {
   }
 
   function handleEditorWillMount(editor, monaco) {
-    const decorations = editor.deltaDecorations([],
-      [
-        {
-          range: new monaco.Range(1, 1, 1, 1),
-          options: {
-            isWholeLine: true,
-            glyphMarginClassName: "glyphSuccess",
-            glyphMarginHoverMessage: { value: "" },
+    if (!diffView) {
+      const decorations = editor.deltaDecorations([],
+        [
+          {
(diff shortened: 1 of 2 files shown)


frontend/src/components/Editor/Editor.jsx
```diff
@@ -359,24 +359,27 @@ const Editor = ({ actionButtonText }) => {
...
-    const decorations = editor.deltaDecorations([],
-      [
-        {
-          range: new monaco.Range(1, 1, 1, 1),
-          options: {
-            isWholeLine: true,
-            glyphMarginClassName: "glyphSuccess",
-            glyphMarginHoverMessage: { value: "" },
+    if (!diffView) {
+      const decorations = editor.deltaDecorations([],
```
darunrs, line 363: What purpose does this particular delta decoration actually do? Is this a leftover from when you were doing the green box? If so, can we just remove it? 

Also, I believe WillMount is only done on init. 

I think there's a small bug with context.db where on initial reload, it doesn't set the types. You have to change tabs to get it to load, or click the button. This was because of a race condition where WillMount may be done before the editor is actually available. My guess is that is the cause of the error, not that it's setting decorations that don't exist.

darunrs (commented):
A comment on the fix.

frontend/src/components/Editor/Editor.jsx
Kevin101Zhang: Its necessary to mount the glyph here. 
We do this to grab the decoration that takes a shape ['W;3'] and the monaco editor ref. 
When we run parseGlyph it is overwriting the one decoration in line

`monacoEditorRef.current.deltaDecorations(
      [decorations],
      [
        {
          range: new monaco.Range(startLine, startColumn, endLine, endColumn),
          options: {
            isWholeLine: true,
            glyphMarginClassName: error ? "glyphError" : "glyphSuccess",
            glyphMarginHoverMessage: { value: displayedError },
          },
        },
      ]
    );`
    
The other option is to create the the decoration for the first time in parseGlyphError. I tried this approach but it makes the code more messy since we are instantiating the glyph here.

Kevin101Zhang: Will go ahead and push the version without options.

darunrs (approved):

darunrs:
Ok sounds good!

PR #572, opened 2024-02-20
feat: Support Star Contract Filter in Block Streamer

### LINKED ISSUE #540: Update the streamer to support this filter

darunrs:
Block Streamer will fail to process a * contract filter as it will try to fetch index files for all account IDs. This is not only extremely inefficient since we aren't planning to skip anything, but will also fail as we limit the number of accounts fetched as a sanity check. 

Block Streamer, if the filter is "*" specifically, now skips fetching index files and begins reading from Lake directly instead.

This retains the behavior of block streamer reading block data before pushing the height to Redis. There is an intended migration of Runner's prefetch into Block Streamer. This change works well towards that goal. However, obviously there is double work performed here: We fetch the block, only to just put the height into Redis. If, in the short term and before we migrate prefetch from Runner to Block Streamer, we need to speed things up, we can list files form S3 instead. However, this work would be removed after the mentioned migration of work.

### DIFF
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -190,7 +190,7 @@ async fn process_delta_lake_blocks(
                 .any(|account_id| DELTA_LAKE_SKIP_ACCOUNTS.contains(&account_id.trim()))
             {
                 tracing::debug!(
-                    "Skipping fetching index files from delta lake due to wildcard contract filter {}",
+                    "Skipping fetching index files from delta lake due to wildcard contract filter present in {}",
                     affected_account_id
                 );
                 return Ok(start_block_height);
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -9,6 +9,7 @@ use registry_types::Rule;
 /// The number of blocks to prefetch within `near-lake-framework`. The internal default is 100, but
 /// we need this configurable for testing purposes.
 const LAKE_PREFETCH_SIZE: usize = 100;
+const DELTA_LAKE_SKIP_ACCOUNTS: [&str; 4] = ["*", "*.near", "*.kaiching", "*.tg"];
 
 pub struct Task {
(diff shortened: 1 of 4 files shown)


darunrs:
I did a simple manual test: I called start_stream with a modified filter of * and observed that it began pulling block heights immediately. I also verified manually that the heights were in the expected sequential order, including skipping skipped heights.

block-streamer/src/block_stream.rs
```diff
@@ -358,4 +365,67 @@ mod tests {
...
+                    last_indexed_block: "107503700".to_string(),
+                    processed_at_utc: "".to_string(),
+                    first_indexed_block: "".to_string(),
+                    last_indexed_block_date: "".to_string(),
+                    first_indexed_block_date: "".to_string(),
+                })
+            });
+        mock_delta_lake_client
+            .expect_list_matching_block_heights()
+            .times(0);
```
morgsmccauley, line 385:
```suggestion
            .never();
```
FYI :)

morgsmccauley (approved):

block-streamer/src/block_stream.rs
```diff
@@ -184,6 +184,13 @@ async fn process_delta_lake_blocks(
             affected_account_id,
             ..
         } => {
+            if affected_account_id.eq("*") {
+                tracing::info!(
```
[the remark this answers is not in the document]
darunrs: Gotcha! And yeah, I changed it to info just to find the log in all the debug log noise haha. Was on the fence for switching it back. But yeah I think with what we currently log, it makes sense to keep it as debug.

block-streamer/src/block_stream.rs
darunrs: Oh haha. Thanks! Will change.

darunrs:
As per discussion in team meeting, I'll expand the list of things to skip delta to *, *.near, *.kaiching, *.tg to start.

PR #575, opened 2024-02-22
fix: Use non-archival RPC to deploy widgets

morgsmccauley:
Archival RPC, the default for near/bos CLI seems to result in transactions expiring. non-RPC does not face the same issue. 
 This PR updates the widget deploy CI to use non-archival RPC.


darunrs (approved):
Interesting that we merely needed to add a step to change this. Was the default to use archival specified somewhere else?

morgsmccauley:
> Interesting that we merely needed to add a step to change this. Was the default to use archival specified somewhere else?

@darunrs Yes bos/near CLI default to archival

PR #576, opened 2024-02-22
chore: Remove legacy code

morgsmccauley:
Removes:
- Coordinator V1
- SQS Runner
- Relevant GH actions


darunrs (approved):
LGTM! Just need to also remove Coordinator V1 from the docker compose file too. 

If you're up for it, could we also update the README too?

morgsmccauley:
> LGTM! Just need to also remove Coordinator V1 from the docker compose file too.

Done

>If you're up for it, could we also update the README too?

And Done!

PR #579, opened 2024-02-23
feat: Add metrics to Block Streamer

morgsmccauley:
This PR adds the following metrics to Block Streamer:
- `LAST_PROCESSED_BLOCK` - height of last block seen, will be used to calculate the lag from tip of network
- `PROCESSED_BLOCKS_COUNT` - count of blocks seen, used to calculate BPS
- `PUBLISHED_BLOCKS_COUNT` - count of blocks published to Redis Stream, to determine whether the block stream is publishing messages
- `LOGS_COUNT` - count logs by level, so we can alert on number of error logs etc.

I've also slightly refactored `RedisClient` to support this change.

### DIFF
--- a/block-streamer/src/indexer_config.rs
+++ b/block-streamer/src/indexer_config.rs
@@ -21,4 +21,9 @@ impl IndexerConfig {
         self.get_full_name().hash(&mut hasher);
         hasher.finish().to_string()
     }
+
+    pub fn last_processed_block_key(&self) -> String {
+        // TODO: rename to `last_processed_block`
+        format!("{}:last_published_block", self.get_full_name())
+    }
 }
--- a/block-streamer/src/main.rs
+++ b/block-streamer/src/main.rs
@@ -16,6 +16,7 @@ mod test_utils;
 async fn main() -> anyhow::Result<()> {
     tracing_subscriber::registry()
         .with(tracing_subscriber::fmt::layer())
+        .with(metrics::LogCounter)
         .with(tracing_subscriber::EnvFilter::from_default_env())
         .init();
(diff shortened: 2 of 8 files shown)


block-streamer/src/main.rs
```diff
@@ -15,26 +16,36 @@ mod test_utils;
 async fn main() -> anyhow::Result<()> {
     tracing_subscriber::registry()
         .with(tracing_subscriber::fmt::layer())
+        .with(metrics::LogCounter)
         .with(tracing_subscriber::EnvFilter::from_default_env())
         .init();
 
     let redis_url = std::env::var("REDIS_URL").expect("REDIS_URL is not set");
-    let server_port = std::env::var("SERVER_PORT").expect("SERVER_PORT is not set");
+    let grpc_port = std::env::var("GRPC_PORT").expect("GRPC_PORT is not set");
```
morgsmccauley, line 24: Will also make this change in Terraform

block-streamer/src/indexer_config.rs
```diff
@@ -21,4 +21,9 @@ impl IndexerConfig {
         self.get_full_name().hash(&mut hasher);
         hasher.finish().to_string()
     }
+
+    pub fn last_processed_block_key(&self) -> String {
+        // TODO: rename to `last_processed_block`
+        format!("{}:last_published_block", self.get_full_name())
```
morgsmccauley, line 27: "published" should mean it actually got added to the Redis Stream

darunrs (approved):
LGTM! Looking forward to seeing what the metrics show us!

PR #580, opened 2024-02-23
Prod Release 23/02/24

darunrs:
- feat: Add metrics to Block Streamer (https://github.com/near/queryapi/pull/579)
- [Feat: added CORS matching all routes (https://github.com/near/queryapi/pull/578)

### DIFF
--- a/block-streamer/Cargo.lock
+++ b/block-streamer/Cargo.lock
@@ -3,12 +3,198 @@
 version = 3
 
 [[package]]
+name = "actix-codec"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


darunrs:
We can't quite merge this yet because: 
1. We need to apply changes to prod to scrape block streamer and set METRICS_PORT env var for it too. I suspect Block Streamer would crash without that change. 
2. Terraform apply seems to indicate a replacement of our instances (Including Coordinator V1) due to a COS update. I believe that V1 running briefly is fine, and we can also just shut down the instance immediately too, to be totally sure. Wouldn't mind giving that a go on any day other than Friday. 

I'd be happy to briefly work with @morgsmccauley to merge this on his Monday. Shouldn't be too hard. Just don't want to take that nonzero low risk of nonsense happening in deployment before the weekend.

morgsmccauley:
@darunrs All sorted, I'll merge this now

PR #582, opened 2024-02-26
feat: Cache Streamer Message from Block Streamer

morgsmccauley:


runner/src/redis-client/redis-client.ts
```diff
@@ -20,7 +20,7 @@ export default class RedisClient {
   SMALLEST_STREAM_ID = '0';
   LARGEST_STREAM_ID = '+';
   STREAMS_SET_KEY = 'streams';
-  STREAMER_MESSAGE_HASH_KEY_BASE = 'streamer:message:';
+  STREAMER_MESSAGE_HASH_KEY_BASE = 'streamer_message:';
```
morgsmccauley, line 23: `:` signifies parent, but we don't need that sort of nesting here

darunrs (approved):
LGTM! I think the set command for Redis has never even taken a single ms to complete, so I think we have no problems if all our streamers write on each block. 

Only issue is now we're setting a block even during historical (index files done or a * indexer), but a 60sec expiry would prevent that from being a problem, I think. We can loop back to it after observing any changes to memory usage, I think.

PR #583, opened 2024-02-26
fix: update middleware for CORS

Kevin101Zhang:
Looks like the middleware file was pushed on a dev branch and was ignored when testing for CORS options.
Fix: removed-  frontend/src/pages/middleware.js
out CORS is created from nextjs looking at next.config.js which is a sibling of package.json


darunrs:
Still not sure why so many commits are still showing up but the squash will get rid of them anyway I guess. Just make sure to checkout off main after pulling when this is merged, to get rid of whatever those commits are.

darunrs (approved):

PR #584, opened 2024-02-26
fix: Increase timeout of VM run function

darunrs:
Some indexers such as eduohe.near/access_keys potentially exceeds the timeout of the VM while running. This may be due to latency of DB calls, which is the heaviest parts of the code. We are planning to drive down the latency of these calls, but this is a stopgap to allow indexers to continue running in the mean time.


darunrs:
Just pushing it out since its a simple change, and is blocking.

PR #585, opened 2024-02-27
feat: Cache Database Credentials

### LINKED ISSUE #581: Store Hasura Credentials Fetch
Runner fetches user database credentials from hasura every invocation on a block, which is wasteful since these credentials don't change, and slows down compute. This step is required for running context.db functions. So, we need to refactor Runner workers to remember the credentials after their first fetch, while also ensuring the fetch is deferred until after provisioning, in the case of new indexers.

darunrs:
Currently, Runner fetches DB credentials every invocation of runFunction because provisioning of the user's DB occurs in Runner. As such, we couldn't be sure if we could fetch credentials for the user's DB from Hasura or not. The fetch is slow and increases the latency of any indexers which leverage context.db.

I updated the code to cache the credentials locally after the first successful fetch. This allows all future requests to succeed as these credentials will never change after they are set. The changes were aimed to be minimal as we intend to migrate provisioning out of Runner. In fact, the changes aid in the effort as the credentials fetch is done through the Provisioner class rather than separately in DmlHandler.

### DIFF
--- a/runner/src/index.ts
+++ b/runner/src/index.ts
@@ -1,33 +1,12 @@
 import { startServer as startMetricsServer } from './metrics';
-import RedisClient from './redis-client';
 import startRunnerServer from './server/runner-server';
 import StreamHandler from './stream-handler';
 
 const executors = new Map<string, StreamHandler>();
-const redisClient = new RedisClient();
-const grpcServer = startRunnerServer(executors);
+startRunnerServer(executors);
 
 startMetricsServer().catch((err) => {
   console.error('Failed to start metrics server', err);
 });
 
-void (async function main () {
-  try {
-    const STREAM_HANDLER_THROTTLE_MS = 500;
-    while (true) {
-      const streamKeys = await redisClient.getStreams();
-      streamKeys.forEach((streamKey) => {
-        if (executors.get(streamKey) === undefined) {
-          const streamHandler = new StreamHandler(streamKey);
-          executors.set(streamKey, streamHandler);
-        }
-      });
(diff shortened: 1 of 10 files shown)


runner/src/index.ts
```diff
@@ -1,33 +1,12 @@
...
-      await new Promise((resolve) =>
-        setTimeout(resolve, STREAM_HANDLER_THROTTLE_MS),
-      );
-    }
-  } finally {
-    await redisClient.disconnect();
-    grpcServer.forceShutdown();
-  }
-})();
+void (async function main () {})();
```
darunrs, line 12: Went ahead and cleaned up the Redis scanning here too. I can also move this to happen in another PR, if needed.

runner/src/stream-handler/worker.ts
```diff
@@ -85,8 +84,17 @@ async function blockQueueConsumer (workerContext: WorkerContext, streamKey: stri
   const indexer = new Indexer();
   const isHistorical = workerContext.streamType === 'historical';
   let streamMessageId = '';
-  let indexerName = streamKey.split(':')[0];
   let currBlockHeight = 0;
+  let indexerName = `${workerContext.indexerConfig.account_id}/${workerContext.indexerConfig.function_name}`;
```
darunrs, line 88: Migrated to using config passed into worker, since we no longer use storage after V2 migration.

darunrs: Ah yes you're totally right. I'll update that!

PR #589, opened 2024-02-29
feat: Reduce Runner Logs and DB Writes

### LINKED ISSUE #587: Remove unnecessary Runner logs
Runner logs excessively, both to the machine, and to postgres. Most of these logs are not necessary or contain too much information. 

The plan is to reduce logs the following ways.
1. Only output failure log for first failure (Instead of every 10 seconds)
2. Stop outputting "Running function on X block" logs to GCP
3. Stop outputting "Success: Ran function on X block" logs to GCP
4. Remove DB logs for hasura calls
5. Remove context.db logging 

To support 1, Runner needs a way to know when a failure was the first. So, it needs to remember the state of the Runner locally. This is a good opportunity to introduce a refactor to stop setting of Runner status every loop. We can instead only write when the status changes. 

In addition, there is some overlap with this epic. So, to orient these changes towards our intentions there, I will also introduce a rudimentary log level into Runner. This value will be INFO by default, and the extra logs will be at the DEBUG level. This ensures these logs are automatically skipped while also adding some backbone for expected future implementation.

darunrs:
Runner logs excessively, both to the machine, and to postgres. Most of these logs are not necessary or contain too much information.

The plan is to reduce logs the following ways.

1. Only output failure log for first failure (Instead of every 10 seconds)
2. Stop outputting "Running function on X block" logs to GCP
3. Stop outputting "Success: Ran function on X block" logs to GCP
4. Remove DB logs for hasura calls
5. Remove context.db logging

To support 1, Runner needs a way to know when a failure was the first. So, it needs to remember the state of the Runner locally. This is a good opportunity to introduce a refactor to stop setting of Runner status every loop. We can instead only write when the status changes.

In addition, there is some overlap with this epic. So, to orient these changes towards our intentions there, I will also introduce a rudimentary log level into Runner. This value will be INFO by default, and the extra logs will be at the DEBUG level. This ensures these logs are automatically skipped while also adding some backbone for expected future implementation.

### DIFF
--- a/runner/src/indexer/__snapshots__/indexer.test.ts.snap
+++ b/runner/src/indexer/__snapshots__/indexer.test.ts.snap
@@ -1,5 +1,215 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Indexer unit tests Indexer log level respected by writeLog 1`] = `
+[
+  [
+    "mock-hasura-endpoint/v1/graphql",
+    {
+      "body": "{"query":"mutation writeLog($function_name: String!, $block_height: numeric!, $message: String!){\\n                insert_indexer_log_entries_one(object: {function_name: $function_name, block_height: $block_height, message: $message}) {id}\\n             }","variables":{"function_name":"buildnear.testnet/test","block_height":456,"message":"Running function buildnear.testnet/test on block 456, lag is: NaNms from block timestamp"}}",
+      "headers": {
+        "Content-Type": "application/json",
+        "X-Hasura-Admin-Secret": "mock-hasura-secret",
+        "X-Hasura-Role": "append",
+        "X-Hasura-Use-Backend-Only-Permissions": "true",
+      },
+      "method": "POST",
+    },
+  ],
+  [
+    "mock-hasura-endpoint/v1/graphql",
+    {
+      "body": "{"query":"\\n                mutation SetStatus($function_name: String, $status: String) {\\n                  insert_indexer_state_one(object: {function_name: $function_name, status: $status, current_block_height: 0 }, on_conflict: { constraint: indexer_state_pkey, update_columns: status }) {\\n                    function_name\\n                    status\\n                  }\\n                }\\n            ","variables":{"function_name":"buildnear.testnet/test","status":"RUNNING"}}",
+      "headers": {
+        "Content-Type": "application/json",
+        "X-Hasura-Admin-Secret": "mock-hasura-secret",
+        "X-Hasura-Role": "append",
+        "X-Hasura-Use-Backend-Only-Permissions": "true",
+      },
+      "method": "POST",
+    },
+  ],
+  [
+    "mock-hasura-endpoint/v1/graphql",
+    {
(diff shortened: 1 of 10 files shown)


runner/src/indexer/__snapshots__/indexer.test.ts.snap
```diff
@@ -1,5 +1,215 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Indexer unit tests Indexer log level respected by writeLog 1`] = `
```
darunrs, line 3: These snapshots will be removed after we migrate these calls to direct DB calls.

runner/src/indexer/indexer.test.ts
```diff
@@ -1054,6 +1059,82 @@ CREATE TABLE
...
+        header: {
+          height: blockHeight
+        }
+      },
+      shards: {}
+    } as unknown as StreamerMessage) as unknown as Block;
+    const functions: Record<string, any> = {};
+    functions['buildnear.testnet/test'] = {
+      code: `
+            console.debug('debug log');
```
darunrs, line 1094: An interesting caveat to console.[anything] is that these are async whereas normally one would not consider a console.log statement to be async. So, people tend to not await these, meaning it's possible to see out of order logs.

runner/src/indexer/indexer.ts
```diff
@@ -75,41 +76,41 @@ export default class Indexer {
...
         } catch (e) {
           const error = e as Error;
-          simultaneousPromises.push(this.writeLog(functionName, blockHeight, 'Failed to get database connection parameters', error.message));
+          simultaneousPromises.push(this.writeLog(LogLevel.ERROR, functionName, blockHeight, 'Failed to get database connection parameters', error.message));
           throw error;
         }
 
-        await this.setStatus(functionName, blockHeight, 'RUNNING');
+        // TODO: Prevent unnecesary reruns of set status
+        simultaneousPromises.push(this.setStatus(functionName, blockHeight, 'RUNNING'));
```
darunrs, line 113: Runs every loop. No need to let this run in a blocking way by awaiting here. Also, as a result of running every loop, if an indexer is failing, it writes its running, and then after it fails again, writes its failing. So, it's repeatedly overwriting its own status.

runner/src/indexer/indexer.test.ts
gabehamilton: maybe we need a sequence number? Having them execute in parallel is nice and out of order resolution could be fine if we then order them in the display by sequence number.

runner/src/indexer/indexer.ts
gabehamilton: If it's async could the failure be written first then be overwritten by this 'RUNNING'?

gabehamilton (approved):

runner/src/indexer/indexer.test.ts
darunrs: Ah that's an interesting idea! We're in the planning phases of redesigning the logs table. I'll bring that up as an idea to discuss!

PR #592, opened 2024-03-01
feat: fixed reload button, unmount grid

### LINKED ISSUE #573: "Reload Data" button doesn't load fresh logs
https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=nearpavel.near/staking_v4

image

Kevin101Zhang:
Fixed the reload button, added unmounting of the grid. 
IMO - we really don't need a library(gridjs) for a logging table with simple functions as this is the only area we use it as well.
Happy to create an ticket to migrate for a todo in the future.

Otherwise demo here:

https://www.loom.com/share/c35ca732212e48f2ab11d17238f39812


frontend/src/components/Logs/IndexerLogs.jsx
```diff
@@ -202,19 +203,16 @@ const IndexerLogsComponent = () => {
...
+    grid.render(document.getElementById('grid-logs-container'));
 
-    grid.render(indexerLogsRef.current);
-  };
-
-  useEffect(() => {
-    initializeTable();
+    return () => {
+      grid.destroy();
+    };
```
darunrs, line 211: What does this piece of code do?

darunrs (approved):
LGTM! Nice simple changes. Regarding your question about migration, can you explain what the benefit of the migration would be? Like if removing/migrating results in verifiably faster loading time. Also how long the migration would take.

Kevin101Zhang:
> LGTM! Nice simple changes. Regarding your question about migration, can you explain what the benefit of the migration would be? Like if removing/migrating results in verifiably faster loading time. Also how long the migration would take.

The main benefit is just reducing our dependencies in our package.json file leading to a smaller bundle size, faster load times. The main disadvantages are the package is taking care of virtualization, and pagination. None of the 2 are terribly hard to implement. Its really just a reinventing the wheel. I think its best to revisit when we have the new schema and new frontend detailed out since it doesnt make as much sense to build upon a design we want to revamp

PR #602, opened 2024-03-13
feat: Instrument Runner Service

### LINKED ISSUE #530: Add Instrumentation to Runner
While Grafana/Promethus provides general metrics to assess performance and faults, the granularity of data provided is not sufficient to understand performance characteristics of Runner. We will want to know in greater detail what is contributing to the latency of a runFunction execution, so that we can both detect and prioritize improvements to the workflow, aiming for the fastest backfill performance we can achieve. 

The instrumentation can be gradually improved as we tackle various latency contributors. 

Ideally, we will also integrate the instrumentation library into Grafana (and/or Prometheus) to reuse existing metrics infrastructure.

darunrs:
Runner is lacking instrumentation. It is responsible for many things and it's become hard to understand what tasks contribute to the overall latency of an indexer. In addition, we are now at a point where we need to drive down latencies to facilitate new * indexer use cases such as access keys.

I've chosen to instrument Runner with OpenTelemetry. Tracing generally requires 3 items: An instrumented service, a trace collector, and a trace visualizer. The service is responsible for collecting and transmitting trace data to the collector. The collector should be able to receive trace data with little fuss to prevent performance impacts to the instrumented service. The collector then processes the trace data and transmits the processed data to the visualizer. The visualizer visualizes trace data and allows for filtering on traces. 

The benefit of OpenTelemetry over other options like Zipkin and Jaeger is that GCP already supports ingesting OpenTelemetry data. As such, we don't need to provision a collector ourselves, and can instead leverage GCP's existing collector & visualizer Tracing service. For local development, traces can be output to console, a Zipkin all-in-one container or to GCP (Requires Cloud Trace Agent role and specifying project ID). This is done by simply initializing the NodeSDK differently. 

In addition, we do not want to enable traces in prod yet, so by not specifying any exporter. This creates a No-Op Trace Exporter which won't attempt to record traces. 

No code changes were made changing code execution path. All tests pass with no changes, aside from having to replace snapshots due to changes in tabbing of mutation strings. I have manually verified mutation strings are still the same by stripping whitespace and checking against original.


runner/src/indexer/indexer.ts
```diff
@@ -100,32 +102,41 @@ export default class Indexer {
...
 
-        const modifiedFunction = this.transformIndexerFunction(indexerFunction.code);
-        try {
-          await vm.run(modifiedFunction);
-        } catch (e) {
-          const error = e as Error;
-          await this.writeLog(LogLevel.ERROR, functionName, blockHeight, 'Error running IndexerFunction', error.message);
-          throw e;
-        }
+        await this.tracer.startActiveSpan('run indexer code', async (runIndexerCodeSpan: Span) => {
```
darunrs, line 128: I use setActiveSpan when I anticipate more traces to be made inside the code block that I want to establish a child status for. In this case, any context.db or hasura calls made in the vm should be children of the `run indexer code` span.

runner/src/instrumentation/tracer.ts
```diff
@@ -0,0 +1,87 @@
...
+  console.debug('Using GCP Exporter. Traces exported to GCP Trace.');
+  const sdk = new NodeSDK({
+    resource: new Resource({
+      [SEMRESATTRS_SERVICE_NAME]: 'queryapi-runner',
+      [SEMRESATTRS_SERVICE_VERSION]: '1.0',
+    }),
+    traceExporter: new TraceExporter(),
+    spanProcessors: [new BatchSpanProcessor(new TraceExporter(
+      {
+        projectId: process.env.GCP_PROJECT_ID ?? ''
```
darunrs, line 40: An empty project ID will send the traces nowhere. I figured this was the safer default to use.

runner/src/instrumentation/tracer.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  PeriodicExportingMetricReader,
+  ConsoleMetricExporter,
+} from '@opentelemetry/sdk-metrics';
+import { ZipkinExporter } from '@opentelemetry/exporter-zipkin';
+import { TraceExporter } from '@google-cloud/opentelemetry-cloud-trace-exporter';
+import { TraceIdRatioBasedSampler } from '@opentelemetry/sdk-trace-node';
+
+export default function setUpTracerExport (): void {
+  switch (process.env.TRACING_EXPORTER) {
```
darunrs, line 14: The only change here is what exporter we set below. The rest of the code is the same. SO, swapping out endpoint is actually quite cheap.

runner/src/stream-handler/worker.ts
```diff
@@ -101,59 +111,91 @@ async function blockQueueConsumer (workerContext: WorkerContext, streamKey: stri
...
+            await indexer.runFunctions(block, functions, isHistorical, { provision: true });
+          } finally {
+            runFunctionsSpan.end();
+          }
+        });
+
+        const postRunSpan = tracer.startSpan('Delete redis message and shift queue', {}, context.active());
+        parentPort?.postMessage({ type: WorkerMessageType.STATUS, data: { status: Status.RUNNING } });
+        await workerContext.redisClient.deleteStreamMessage(streamKey, streamMessageId);
+        workerContext.queue.shift();
```
gabehamilton, line 161: If this errors will it escape this function's try/catch since it's not awaited?

gabehamilton (approved):

runner/src/stream-handler/worker.ts
```diff
@@ -1,21 +1,26 @@
...
 }
-interface QueueMessage {
+interface BlockPromise {
   block: Block
   streamMessageId: string
 }
-type PrefetchQueue = Array<Promise<QueueMessage>>;
+interface QueueMessage {
+  promise: Promise<BlockPromise>
+  block_height: number
```
darunrs, line 21: I needed the block height to set the trace information prior to waiting for the block to download. So, I reworked the Queue message to allow reading of the height before awaiting block data.

runner/src/stream-handler/worker.ts
darunrs: Are you referring to the queue shift? If so, yep I spotted that and put an await on it. startSpan is sync. It doesn't need an await. I'll be pushing the change up. I'm actually reverting to not modify the QueueMessage. Fewer changes to worker and putting the block height in the root span doesn't help much since we sample only 10% anyway. The block height is listed in a child span later.

PR #606, opened 2024-03-18
Support WHERE col IN (...) in context.db.table.select and delete

pkudinov:
With this PR, we can use `context.db.Table.select({column_name: ['a.near', 'b.near']})`. I think it is best to keep it within the same select method instead of introducing a new one like `selectMany`. The same support is added for `delete`.

Frontend support is added. I also improved parameter naming to reflect SQL statements like `where`, `values` and `set`. Here is the sample output:
```TS
declare interface ActionsIndexItem {
  block_date: Date;
  receiver_id: string;
  first_block_height: number;
  bitmap: string;
}

declare interface ActionsIndexPartial {
  block_date?: Date;
  receiver_id?: string;
  first_block_height?: number;
  bitmap?: string;
}

declare interface ActionsIndexFilter {
  block_date?: Date | Date[];
  receiver_id?: string | string[];
  first_block_height?: number | number[];
  bitmap?: string | string[];
}

type ActionsIndexColumns = "block_date" | "receiver_id" | "first_block_height" | "bitmap";

declare const context: {
	graphql: (operation, variables) => Promise<any>,
	set: (key, value) => Promise<any>,
	log: (...log) => Promise<any>,
	fetchFromSocialApi: (path, options) => Promise<Response>,
	db: {
		ActionsIndex: {
			insert: (values: ActionsIndexItem | ActionsIndexItem[]) => Promise<ActionsIndexItem[]>;
			select: (where: ActionsIndexFilter, limit = null) => Promise<ActionsIndexItem[]>;
			update: (where: ActionsIndexPartial, set: ActionsIndexPartial) => Promise<ActionsIndexItem[]>;
			upsert: (values: ActionsIndexItem | ActionsIndexItem[], conflictColumns: ActionsIndexColumns[], updateColumns: ActionsIndexColumns[]) => Promise<ActionsIndexItem[]>;
			delete: (where: ActionsIndexFilter) => Promise<ActionsIndexItem[]>;
		},
  }
};
```

### DIFF
--- a/runner/src/dml-handler/dml-handler.test.ts
+++ b/runner/src/dml-handler/dml-handler.test.ts
@@ -1,6 +1,6 @@
 import pgFormat from 'pg-format';
 import DmlHandler from './dml-handler';
-import PgClient from '../pg-client';
+import type PgClient from '../pg-client';
 
 describe('DML Handler tests', () => {
   const getDbConnectionParameters = {
@@ -19,7 +19,7 @@ describe('DML Handler tests', () => {
   beforeEach(() => {
     query = jest.fn().mockReturnValue({ rows: [] });
     pgClient = {
-      query: query, 
+      query,
       format: pgFormat
     } as unknown as PgClient;
   });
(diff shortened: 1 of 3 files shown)


runner/src/dml-handler/dml-handler.test.ts
```diff
@@ -1,6 +1,6 @@
 import pgFormat from 'pg-format';
 import DmlHandler from './dml-handler';
-import PgClient from '../pg-client';
+import type PgClient from '../pg-client';
```
pkudinov, line 3: @darunrs please note this change after I hit `ESLint fix this file`

darunrs (approved):
LGTM! Nice changes!

darunrs:
Let me check if the frontend allows for this too.

darunrs:
Yeah it will not. Let me suggest the frontend fix to allow this...

darunrs:
let tsTypeBase = columnDetails.type + " | " + columnDetails.type + "[]";
let tsType = columnDetails.nullable ? tsTypeBase + " | null" : tsTypeBase;

https://github.com/near/queryapi/blob/main/frontend/src/utils/pgSchemaTypeGen.js#L172

darunrs (approved):
Some small comments. Otherwise, it seems to work! I tried it out in the frontend and ran a simple indexer in the backend that does selects and deletes and observed the right behavior.

PR #607, opened 2024-03-19
Cache provisioning status

### LINKED ISSUE #603: Cache Provisioning Status

pkudinov:
Checking provisioning status through Hasura takes 70-100ms on Dev. This PR caches the provisioning status inside of `Provisioner` and does make extra requests to Hasura on every run.

### DIFF
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -86,7 +86,7 @@ export default class Indexer {
 
         if (options.provision && !indexerFunction.provisioned) {
           try {
-            if (!await this.deps.provisioner.isUserApiProvisioned(indexerFunction.account_id, indexerFunction.function_name)) {
+            if (!this.deps.provisioner.isUserApiProvisioned(indexerFunction.account_id, indexerFunction.function_name)) {
               await this.setStatus(functionName, blockHeight, 'PROVISIONING');
               simultaneousPromises.push(this.writeLog(LogLevel.INFO, functionName, blockHeight, 'Provisioning endpoint: starting'));
 
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -86,7 +86,7 @@ export default class Indexer {
 
         if (options.provision && !indexerFunction.provisioned) {
           try {
-            if (!this.deps.provisioner.isUserApiProvisioned(indexerFunction.account_id, indexerFunction.function_name)) {
+            if (!await this.deps.provisioner.fetchUserApiProvisioningStatus(indexerFunction.account_id, indexerFunction.function_name)) {
               await this.setStatus(functionName, blockHeight, 'PROVISIONING');
               simultaneousPromises.push(this.writeLog(LogLevel.INFO, functionName, blockHeight, 'Provisioning endpoint: starting'));
 
--- a/runner/src/provisioner/provisioner.ts
+++ b/runner/src/provisioner/provisioner.ts
@@ -52,14 +52,8 @@ export default class Provisioner {
   }
 
   private setProvisioned (accountId: string, functionName: string): void {
-    const accountIndexers = this.#hasBeenProvisioned[accountId] ?? {};
-    this.#hasBeenProvisioned = {
-      ...this.#hasBeenProvisioned,
-      [accountId]: {
-        ...accountIndexers,
-        [functionName]: true
-      }
-    };
+    this.#hasBeenProvisioned[accountId] ??= {};
+    this.#hasBeenProvisioned[accountId][functionName] = true;
   }
 
   async createDatabase (name: string): Promise<void> {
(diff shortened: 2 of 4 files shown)


runner/src/provisioner/provisioner.ts
```diff
@@ -44,6 +45,23 @@ export default class Provisioner {
       .replace(/\//g, '0');
   }
 
+  isUserApiProvisioned(accountId: string, functionName: string): boolean {
+    const accountIndexers = this.#hasBeenProvisioned[accountId];
+    if (!accountIndexers) { return false; }
+    return accountIndexers[functionName];
+  }
+
+  private setProvisioned(accountId: string, functionName: string) {
```
darunrs, line 54: Should return void. Also, I think Typescript complains if there isn't a space between the function name and the parenthesis. You'll want to do `isUserApiProvisioned (...)` and `setProvisioned (...)`.

runner/src/provisioner/provisioner.ts
```diff
@@ -44,6 +45,23 @@ export default class Provisioner {
...
+    const accountIndexers = this.#hasBeenProvisioned[accountId] ?? {};
+    this.#hasBeenProvisioned = {
+      ...this.#hasBeenProvisioned,
+      [accountId]: {
+        ...accountIndexers,
+        [functionName]: true
+      }
+    }
```
darunrs, line 62: I think there's a missing semicolon here.

runner/src/indexer/indexer.ts
```diff
@@ -86,7 +86,7 @@ export default class Indexer {
 
         if (options.provision && !indexerFunction.provisioned) {
           try {
-            if (!await this.deps.provisioner.isUserApiProvisioned(indexerFunction.account_id, indexerFunction.function_name)) {
+            if (!this.deps.provisioner.isUserApiProvisioned(indexerFunction.account_id, indexerFunction.function_name)) {
```
darunrs, line 89: This should be calling `fetchUserApiProvisioningStatus`, not `isUserApiProvisioned`. Or you can switch the names. Probably better to just switch the names or else you'll need to rename indexer tests.

runner/src/provisioner/provisioner.ts
```diff
@@ -44,6 +45,23 @@ export default class Provisioner {
       .replace(/\//g, '0');
   }
 
+  isUserApiProvisioned(accountId: string, functionName: string): boolean {
```
darunrs, line 48: Considering this function is meant to be used in another function, this should become private.

darunrs (changes requested):

runner/src/provisioner/provisioner.ts
```diff
@@ -24,6 +24,7 @@ export interface DatabaseConnectionParameters {
 
 export default class Provisioner {
   tracer: Tracer = trace.getTracer('queryapi-runner-provisioner');
+  #hasBeenProvisioned: Record<string, Record<string, boolean>> = {};
```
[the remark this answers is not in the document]
pkudinov: I think this is a great answer: https://stackoverflow.com/a/59641565

TL;DR; – this makes the field truly private unlike the use of `private` keyword.

runner/src/provisioner/provisioner.ts
pkudinov: Oh, my ESLint stopped working. I fixed this

runner/src/indexer/indexer.ts
pkudinov: Correct, changed to `fetchUserApiProvisioningStatus`. Not sure why the tests should be changed?

runner/src/provisioner/provisioner.ts
pkudinov: This is only made public for tests, which is not ideal of course, but does not require too many changes

runner/src/indexer/indexer.ts
pkudinov: Got it, and fixed tests. I think `fetchUserApiProvisioningStatus` is better naming to reflect that this is an expensive call

runner/src/provisioner/provisioner.ts
```diff
@@ -44,6 +45,23 @@ export default class Provisioner {
...
+
+  private setProvisioned (accountId: string, functionName: string): void {
+    const accountIndexers = this.#hasBeenProvisioned[accountId] ?? {};
+    this.#hasBeenProvisioned = {
+      ...this.#hasBeenProvisioned,
+      [accountId]: {
+        ...accountIndexers,
+        [functionName]: true
+      }
+    };
```
morgsmccauley, line 62:
```suggestion
    this.#hasBeenProvisioned[accountId] ??= {};
    this.#hasBeenProvisioned[accountId][functionName] = true;
```
We can use Nullish coalescing assignment (`??=`) to ensure the account record exists, and just mutate the single field rather than replacing the entire object.

darunrs (approved):
LGTM!

PR #608, opened 2024-03-19
feat: Introducing Logging Class (Disabled Usage of Logger)

### LINKED ISSUE #299: Implement Logging and Setting Status/BlockHeight for new Tables
- Partition by date
- Don't set primary key to `block_height`
- Index by severity
- Introduce the new Hasura endpoint for logs
- Add log schema provisioning step

```[tasklist]
### Tasks
- [ ] https://github.com/near/queryapi/issues/499
- [ ] https://github.com/near/queryapi/issues/586
- [ ] https://github.com/near/queryapi/issues/631
- [ ] https://github.com/near/queryapi/issues/638
- [ ] https://github.com/near/queryapi/issues/650
```

Blocked
- Delete the table on indexer deletion

Kevin101Zhang:
Added code for writing logs to new logs table through Postgres instead of Hasura. 

https://www.loom.com/share/ff21d7099cac403d9152c905f7e4ddcc?sid=5828ae99-377b-4510-ac8c-76c02fd232f2


Kevin101Zhang:
cases until completion:

1. cron statements (currently commented out)
2. add cases where user's indexer is already provisioned and we must provision only the logs table
3. update provision.test.ts & hasura-client.test.ts

runner/src/indexer/indexer.ts
```diff
@@ -336,27 +353,32 @@ export default class Indexer {
...
+    const schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
     const parsedMessage: string = message
       .map(m => typeof m === 'object' ? JSON.stringify(m) : m)
       .join(':');
 
     const mutation =
-            `mutation writeLog($function_name: String!, $block_height: numeric!, $message: String!){
-                insert_indexer_log_entries_one(object: {function_name: $function_name, block_height: $block_height, message: $message}) {id}
-             }`;
+            `mutation InsertLogs($block_height: numeric, $log_date: date, $log_level: String, $log_type: String, $message: String) {
```
pkudinov, line 366: This must be done through SQL query – we want to avoid using GraphQL for such operations.

pkudinov (dismissed):

runner/src/indexer/indexer.ts
Kevin101Zhang: ok I will change it to a sql equivalent

pkudinov: And we should make batch INSERT calls to reduce the number of queries to the database. Basically logs should be buffered in-memory and written in the single query.

runner/src/provisioner/provisioner.ts
```diff
@@ -141,16 +155,19 @@ export default class Provisioner {
           const tableNames = await this.getTableNames(schemaName, HasuraClient.DEFAULT_DATABASE);
           await this.hasuraClient.untrackTables(HasuraClient.DEFAULT_DATABASE, schemaName, tableNames);
         }
-
+
```
morgsmccauley, line 158: 🔪

runner/src/hasura-client/hasura-client.ts
```diff
@@ -123,12 +123,81 @@ export default class HasuraClient {
...
   async createSchema (source: string, schemaName: string): Promise<any> {
-    return await this.executeSql(`CREATE schema ${schemaName}`, {
+    //toFix: change back and remove Drop Schema
+    return await this.executeSql(`DROP SCHEMA IF EXISTS ${schemaName} CASCADE;CREATE schema ${schemaName}`, {
       source,
       readOnly: false,
     });
   }
 
+  async createLogs (source: string, schemaName: string): Promise<any> {
```
morgsmccauley, line 133:
```suggestion
  async createLogsTable (source: string, schemaName: string): Promise<any> {
```
This seems more clear

runner/src/hasura-client/hasura-client.ts
```diff
@@ -123,12 +123,81 @@ export default class HasuraClient {
...
+  END
+  $func$;
+
+  --  GRANT CREATE, USAGE ON SCHEMA cron TO ${source};
+  --  SELECT cron.schedule('${schemaName}_logs_create_partition', '0 1 * * *', $$SELECT fn_create_partition('${schemaName}.logs', CURRENT_DATE, '1 day', '2 day')$$);
+  --  SELECT cron.schedule('${schemaName}_logs_delete_partition', '0 2 * * *', $$SELECT fn_delete_partition('${schemaName}.logs', CURRENT_DATE, '-15 day', '-14 day')$$);
+      `,
+      { source, readOnly: false }
+    );
+  }
```
morgsmccauley, line 200:
```suggestion
  }
  
```

runner/src/hasura-client/hasura-client.ts
```diff
@@ -146,7 +215,6 @@ export default class HasuraClient {
         source,
       }
     );
-
```
morgsmccauley, line 149: 🔪

runner/src/hasura-client/hasura-client.ts
```diff
@@ -327,6 +395,33 @@ export default class HasuraClient {
     );
   }
 
+  async addPermissionsToLogsTable (schemaName: string, source: string, tableNames: string[], roleName: string, permissions: string[]): Promise<any> {
```
morgsmccauley, line 398: This shouldn't be necessary right? `addPermissionsToTables` should pick up the logs tables automatically

runner/src/hasura-client/hasura-client.ts
```diff
@@ -123,12 +123,81 @@ export default class HasuraClient {
...
       readOnly: false,
     });
   }
 
+  async createLogs (source: string, schemaName: string): Promise<any> {
+    return await this.executeSql(
+      `
+      set schema '${schemaName}';
+    
+      CREATE TABLE logs (
```
morgsmccauley, line 138: I would prefix all created entities with `__`, i.e. `__logs` to:
1. Signify that this is internal/metadata
2. Avoid conflicts with user created entities

runner/src/indexer/indexer.ts
```diff
@@ -9,6 +9,13 @@ import { type IndexerBehavior, LogLevel, Status } from '../stream-handler/stream
 import { type DatabaseConnectionParameters } from '../provisioner/provisioner';
 import assert from 'assert';
 
+const enum LogLevelNew {
```
morgsmccauley, line 12: Following on from the sync discussion in the weekly - an extracted `Logging` `class` would be good here.

morgsmccauley (changes requested):

runner/src/hasura-client/hasura-client.ts
pkudinov: I wonder if it's worth storing this SQL in a separate SQL file for cleaner code.

runner/src/hasura-client/hasura-client.ts
Kevin101Zhang: Im not too familiar why exactly, but I need the append role here.
Screenshot 2024-03-20 at 4 27 54 PM

darunrs: The append role is a default role Hasura uses to make writes to the default schema tables (state and logs). The role is defined in Indexer here and you can see it is passed into the graphql query function as the role to be used when executing those queries. The reason is that Hasura only allows append and admin to write to these tables. To clarify, this is a **hasura** permission. As far as I can tell, the append role is not an actual Postgres role. Just a virtual role of sorts only recognized by Hasura. 

Regardless, it is not necessary for your logs table to give the append role write access because of two reasons:
1. You can use the user's role to make these writes instead of some generic one like append. 
2. You are making writes to the new logs table through Postgres, not Hasura. As I mentioned earlier, you can't even use append if you wanted to, through Postgres. And we wouldn't want to. We would want to use the user's role since it's in their schema. 

However, we do have the question of how we allow queries to be visible through Hasura (for the frontend). What role does the frontend use for querying logs? Is it append? If so, should we provision a new role which has read access the new logs tables, or reuse append even if append will no longer append anything anymore? 

We'll need to know the answers to those questions and then decide how to structure permissions for roles to read from the new tables. 

Finally, I don't understand why you basically redefined the above addPermissionsToTable function which no changes to its inner code? You can just call the original function again but pass in the append role instead of userRole with whatever permissions it should have.

morgsmccauley: I'm not sure what the frontend does currently, but going forward we should use the users account ID as the role for accessing logs, just like we do for their other data. This provides a single role for controlling access to the users data - we should view logs as being "owned" by the user and should therefore restricted as such.

Therefore, the existing `addPermissionsToTables` method _should_ work, although it will add `update`/`insert`/`delete`, which may be unwanted given we want to write directly from Postgres.

runner/src/indexer/indexer.ts
```diff
@@ -285,8 +289,8 @@ export default class Indexer {
               return await this.tracer.startActiveSpan('Call context db insert', async (insertSpan: Span) => {
                 try {
                   // Write log before calling insert
-                  await this.writeLog(LogLevel.DEBUG, functionName, blockHeight,
-                    `Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`);
+                  await this.writeLogOld(LogLevel.DEBUG, functionName, blockHeight,`Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`);
+                  await this.writeLog(blockHeight, functionName, new Date(), new Date(), 'system', LogLevel.DEBUG, `Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`);
```
darunrs, line 293: See my other comment about running these in parallel.

runner/src/indexer/indexer.ts
```diff
@@ -391,32 +395,13 @@ export default class Indexer {
     }
   }
 
-  async writeLog (logLevel: LogLevel, functionName: string, blockHeight: number, ...message: any[]): Promise<any> {
+  async writeLog (blockHeight: number, functionName:string, logDate: Date, logTimestamp: Date, logType: string, logLevel: number, ...message: any ): Promise<any> {
```
darunrs, line 398: As mentioned before, there is no need to have logDate and logTimestamp both.

darunrs (changes requested):
Nice improvements. Left various comments. I think once you've addressed all of these, you can open up the PR to get full feedback. But, you also need to create a PR for [the infra repo](https://github.com/near/near-ops/tree/master/provisioning/terraform/data_stack/queryapi) to add cron. And make sure that the cron works. Make sure you understand what dependency on infrastructure there is for this code change. If Infra changes need to be released first, don't merge this PR until the infra changes are up.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,60 @@
...
+          database: databaseConnectionParameters.database,
+        });
+        return new IndexerLogger(pgClient);
+      }
+  
+  async writeLog(
+    blockHeight: number,
+    functionName: string,
+    logDate: Date,
+    logTimestamp: Date,
```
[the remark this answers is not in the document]
Kevin101Zhang: Just to confirm the comment here - I should pass a single instance of Date.now() into the writeLog (logTimestamp) and generate logdate from this param?

As opposed to 
` await this.writeLog(blockHeight, functionName, new Date(), new Date(), 'system', LogLevel.DEBUG,  ...log)`
we do
` await this.writeLog(blockHeight, functionName, new Date(), 'system', LogLevel.DEBUG,  ...log)`

and in the logger-class will use the single Date.now() to extrace log-date.

runner/src/indexer/indexer.ts
```diff
@@ -195,19 +197,22 @@ export default class Indexer {
...
       error: async (...log) => {
-        return await this.writeLog(LogLevel.ERROR, functionName, blockHeight, ...log);
+        await this.writeLog(blockHeight, functionName, new Date(), new Date(), 'system', LogLevel.ERROR,  ...log)
+        return await this.writeLogOld(LogLevel.ERROR, functionName, blockHeight, ...log);
       },
       fetchFromSocialApi: async (path, options) => {
         return await this.deps.fetch(`https://api.near.social${path}`, options);
       },
-      db: this.buildDatabaseContext(functionName, schemaName, schema, blockHeight)
+      db: this.buildDatabaseContext(schemaName, functionName, schema, blockHeight)
```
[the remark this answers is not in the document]
Kevin101Zhang: Thanks - unsure when this took place.

runner/src/indexer-logger/indexer-logger.ts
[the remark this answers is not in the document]
darunrs: Correct. Generate logDate from logTimestamp.

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
...
+-- Detach partition 
+EXECUTE 'ALTER TABLE ' || _tbl || ' DETACH PARTITION ' || _tbl || '_p' || _partition_name;
+EXECUTE 'DROP TABLE '  || _tbl || '_p' || _partition_name;
+END
+$func$;
+
+-- GRANT CREATE, USAGE ON SCHEMA cron TO ${source};
+-- SELECT cron.schedule('${schemaName}___logs_create_partition', '0 1 * * *', $$SELECT fn_create_partition('${schemaName}.logs', CURRENT_DATE, '1 day', '2 day')$$);
+-- SELECT cron.schedule('${schemaName}___logs_delete_partition', '0 2 * * *', $$SELECT fn_delete_partition('${schemaName}.logs', CURRENT_DATE, '-15 day', '-14 day')$$);
+`;
```
Kevin101Zhang, line 61: added space

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,88 @@
...
+      entry.blockHeight,
+      this.formatDate(entry.logTimestamp),
+      entry.logTimestamp,
+      entry.logType,
+      LogLevel[entry.logLevel],
+      entry.message
+  ]);
+
+    const valuePlaceholders = logEntries.map((_, index) => `($${index * 6 + 1}, $${index * 6 + 2}, $${index * 6 + 3}, $${index * 6 + 4}, $${index * 6 + 5}, $${index * 6 + 6})`).join(', ');
+    const query =
```
morgsmccauley, line 81: We should be using `pg-format` here to avoid sql injection. It even helps with inserting multiple objects so you wouldn't need to calculate placeholders etc. https://github.com/datalanche/node-pg-format?tab=readme-ov-file#-arrays-and-objects

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  private extractDateParts(date: Date): { year: number, month: number, day: number } {
+    return {
+        year: date.getFullYear(),
+        month: date.getMonth(),
+        day: date.getDate()
+    };
+  }
+
+  formatDate(date: Date): Date {
```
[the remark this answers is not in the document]
morgsmccauley: I'm not sure this conversion is needed - node-postgres will do the [conversion](https://node-postgres.com/features/types#date--timestamp--timestamptz) for us, we can just pass the provided timestamp to both `log_date` and `log_timestamp`

runner/src/indexer/__snapshots__/indexer.test.ts.snap
```diff
@@ -1,215 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Indexer unit tests Indexer log level respected by writeLog 1`] = `
```
morgsmccauley, line 3: This removal seems incorrect?

runner/src/provisioner/provisioner.ts
```diff
@@ -99,6 +99,10 @@ export default class Provisioner {
 
     const schemaExists = await this.hasuraClient.doesSchemaExist(databaseName, schemaName);
     if (schemaExists) {
+      const tableNames = await this.getTableNames(schemaName, sanitizedAccountId);
+      if (!tableNames.includes('__logs')) {
+        await this.createLogsTable(schemaName, databaseName);
```
morgsmccauley, line 104: Let's not add side-effects to this method - at surface value, I'd expect `fetchUserApiProvisioningStatus` to tell me whether things have been provisioned correctly, and not act of the fact.

We should "manually" create the `__logs` table for all existing users, making this unnecessary.

runner/src/hasura-client/hasura-client.ts
```diff
@@ -129,6 +129,17 @@ export default class HasuraClient {
     });
   }
 
+  async createLogsTable (source: string, schemaName: string): Promise<any> {
+    const ddl = logsTableDDL(source, schemaName);
```
morgsmccauley, line 133: The logs DDL should be imported in to `Provisioner`, not here. Hasura is just an interface and shouldn't have business specific logic.

runner/src/provisioner/provisioner.ts
```diff
@@ -159,25 +167,29 @@ export default class Provisioner {
...
-          if (await this.hasuraClient.doesSchemaExist(HasuraClient.DEFAULT_DATABASE, schemaName)) {
+        // Check if the schema and logs table exist
+        if (await this.hasuraClient.doesSchemaExist(HasuraClient.DEFAULT_DATABASE, schemaName)) {
             const tableNames = await this.getTableNames(schemaName, HasuraClient.DEFAULT_DATABASE);
             await this.hasuraClient.untrackTables(HasuraClient.DEFAULT_DATABASE, schemaName, tableNames);
-          }
+            if (!tableNames.includes('__logs')) {
+                await this.createLogsTable(databaseName, schemaName);
+            }
+        } else {
```
morgsmccauley, line 179: This check is actually unnecessary now, it was for an previous migration - we can remove it.

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
...
+    block_height NUMERIC(20) NOT NULL,
+    log_date DATE NOT NULL,
+    log_timestamp TIMESTAMP NOT NULL,
+    log_type TEXT NOT NULL,
+    log_level TEXT NOT NULL,
+    message TEXT NOT NULL,
+    PRIMARY KEY (log_date, id)
+) PARTITION BY RANGE (log_date);
+
+CREATE INDEX logs_log_timestamp_idx ON __logs USING btree (log_timestamp);
```
morgsmccauley, line 13:
```suggestion
CREATE INDEX __logs_log_timestamp_idx ON __logs USING btree (log_timestamp);
```
While conflicts are unlikely, I would still prefix this just to signify it's internal

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
+export const logsTableDDL = (source: string, schemaName: string) => `
```
morgsmccauley, line 1: `source` is not needed, and we can use the pg function `current_schema()` also making `schemaName` unnecessary.

Without the need to inject values I think this would make more sense as a `sql` file.

runner/src/stream-handler/stream-handler.ts
```diff
@@ -87,7 +89,7 @@ export default class StreamHandler {
     indexer.setStatus(functionName, 0, Status.STOPPED).catch((e) => {
       console.error(`Failed to set status STOPPED for stream: ${this.streamKey}`, e);
     });
-    indexer.writeLog(LogLevel.ERROR, functionName, this.executorContext.block_height,
+    indexer.writeLogOld(LogLevel.ERROR, functionName, this.executorContext.block_height,
```
morgsmccauley, line 92: We should also write to the new logs table

runner/src/indexer/indexer.ts
```diff
@@ -14,14 +17,16 @@ interface Dependencies {
   fetch: typeof fetch
   provisioner: Provisioner
   DmlHandler: typeof DmlHandler
-  parser: Parser
+  parser: Parser,
```
morgsmccauley, line 20: Can you run eslint over the project, this comes up as an error.

runner/src/indexer/indexer.ts
```diff
@@ -72,48 +79,49 @@ export default class Indexer {
...
+        const runningMessage = `Running function ${functionName} on block ${blockHeight}, lag is: ${lag?.toString()}ms from block timestamp`;
+        simultaneousPromises.push(this.writeLogOld(LogLevel.INFO, functionName, blockHeight, runningMessage));
+        writeLogBatch.push({ blockHeight, functionName, logTimestamp: new Date(), logType: LogType.SYSTEM, logLevel: LogLevel.INFO, message: runningMessage });
         // Cache database credentials after provisioning
         const credentialsFetchSpan = this.tracer.startSpan('fetch database connection parameters');
         try {
-          this.database_connection_parameters = this.database_connection_parameters ??
-            await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName);
+          this.database_connection_parameters = this.database_connection_parameters ?? await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName);
+          this.indexer_logger = this.indexer_logger ?? IndexerLogger.create(this.database_connection_parameters!);
```
morgsmccauley, line 115: We should not use `!`, this can lead to runtime errors as we are asserting this is defined, when in fact it can not be.

runner/src/indexer/indexer.ts
```diff
@@ -195,14 +204,28 @@ export default class Indexer {
         }
       },
       debug: async (...log) => {
-        return await this.writeLog(LogLevel.DEBUG, functionName, blockHeight, ...log);
+        await Promise.all([
+          this.writeLog(blockHeight, functionName, new Date(), LogType.SYSTEM, LogLevel.DEBUG, ...log),
+          this.writeLogOld(LogLevel.DEBUG, functionName, blockHeight, ...log)
```
morgsmccauley, line 209: You could probably just call `writeLogOld` from `writeLog` rather than duplicating the calls everywhere.

runner/src/indexer/indexer.ts
```diff
@@ -143,7 +152,8 @@ export default class Indexer {
         await this.setStatus(functionName, blockHeight, Status.FAILING);
         throw e;
       } finally {
-        await Promise.all(simultaneousPromises);
+        //do we want to trace this promise?
+        await Promise.all([...simultaneousPromises, this.indexer_logger?.writeLogBatch(writeLogBatch) ?? Promise.resolve()]);
```
morgsmccauley, line 156:
```suggestion
        await Promise.all([...simultaneousPromises, this.indexer_logger?.writeLogBatch(writeLogBatch)]);
```

morgsmccauley: I think `Promise.all([undefined])` is ok

morgsmccauley (changes requested):

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
...
+CREATE TABLE __logs (
+    id BIGSERIAL NOT NULL,
+    block_height NUMERIC(20) NOT NULL,
+    log_date DATE NOT NULL,
+    log_timestamp TIMESTAMP NOT NULL,
+    log_type TEXT NOT NULL,
+    log_level TEXT NOT NULL,
```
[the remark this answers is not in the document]
Kevin101Zhang: I agree. I'll double check with Eduardo to ensure there won't be any additional compatibility or naming conflicts after removing the "log_" prefixes.

runner/src/hasura-client/hasura-client.ts
```diff
@@ -129,6 +129,17 @@ export default class HasuraClient {
     });
   }
 
+  async createLogsTable (source: string, schemaName: string): Promise<any> {
```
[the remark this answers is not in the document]
Kevin101Zhang: Just to confirm, are we using runSql twice.

runSql(logs schema)
runSql(user's schema)

or are we going to combine both ddl to one connection.

runSql(logs schema + user's schema)
    
```
  async runSql (source: string, schemaName: string, userSchemaDDL: string, logsSchemaDDL): Promise<any> {
    return await this.executeSql(
      `
      set schema '${schemaName}';
      ${logsSchemaDDL}
      ${userSchemaDDL}
      `,
      { source, readOnly: false }
    );
  }
```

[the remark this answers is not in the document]
morgsmccauley: The first option - `runSql` takes arbitrary sql and executes it

runner/src/provisioner/provisioner.ts
Kevin101Zhang: Are you suggesting to remove that code and not to add the logs table for users with an existing provisioned schema, but instead to allow the logs to fail and log a message indicating that the schema does not exist?

runner/src/indexer/indexer.ts
Kevin101Zhang: Without the assertion, ts will raise an error. How about this?
```
if (this.database_connection_parameters) {
    this.indexer_logger = this.indexer_logger ?? IndexerLogger.create(this.database_connection_parameters);
}
```

runner/src/provisioner/schemas/logs-table.ts
Kevin101Zhang: There is going to be an added layer of complexity since ts ignores files than json / TS and JS. To get that file into output folder on compilation would only work if I follow the instructions in here.

https://stackoverflow.com/questions/65435921/tsconfig-to-copy-sql-file-to-dist
https://github.com/microsoft/TypeScript/issues/30835

morgsmccauley: Fair point, let's leave as `.ts` then 👍🏼

runner/src/indexer/indexer.ts
morgsmccauley: I didn't realise it actually will be defined at this point since we're assigning it on the line above. To work around the TS error your solution is fine, this will also work:
```ts
this.indexer_logger = this.indexer_logger ?? IndexerLogger.create(this.database_connection_parameters as DatabaseConnectionParameters);
```
We can also use nullish assignment, so this would be the same:
```ts
this.indexer_logger ??= IndexerLogger.create(this.database_connection_parameters as DatabaseConnectionParameters);
```

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
...
+_start := TO_CHAR(date_trunc('day', _date + (_interval_start)::interval), 'YYYY-MM-DD');
+_end := TO_CHAR(date_trunc('day', _date + (_interval_end)::interval), 'YYYY-MM-DD');
+_partition_name := TO_CHAR(date_trunc('day', _date + (_interval_start)::interval), 'YYYYMMDD');
+-- Detach partition 
+EXECUTE 'ALTER TABLE ' || _tbl || ' DETACH PARTITION ' || _tbl || '_p' || _partition_name;
+EXECUTE 'DROP TABLE '  || _tbl || '_p' || _partition_name;
+END
+$func$;
+
+-- GRANT CREATE, USAGE ON SCHEMA cron TO PUBLIC;
```
darunrs, line 58: You've still got some cron SQL here. Is this no longer necessary after Morgan's PR?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -87,9 +90,10 @@ export default class StreamHandler {
     indexer.setStatus(functionName, 0, Status.STOPPED).catch((e) => {
       console.error(`Failed to set status STOPPED for stream: ${this.streamKey}`, e);
     });
-    indexer.writeLog(LogLevel.ERROR, functionName, this.executorContext.block_height,
-      `Encountered error processing stream: ${this.streamKey}, terminating thread\n${error.toString()}`
-    ).catch((e) => {
+    Promise.all([
+      indexer.writeLogOld(LogLevel.ERROR, functionName, this.executorContext.block_height, `Encountered error processing stream: ${this.streamKey}, terminating thread\n${error.toString()}`),
+      indexer.writeLog({blockHeight: this.executorContext.block_height, logTimestamp: new Date(), logType: LogType.SYSTEM, logLevel: LogLevel.ERROR, message: `Encountered error processing stream: ${this.streamKey}, terminating thread\n${error.toString()}`})
```
darunrs, line 95: This is probably the one place where you should still just call writeLog rather than add it to some batch.

darunrs (commented):
Here's a new round of comments. It's definitely real close to being ready. I also unresolved some comments which were not actually addressed. 

My suggestion is to resolve the comment only after you made that particular code change and committed it locally. Or resolve them after pushing changes up to the PR. This way, you don't lose track of them. Resolving is not necessary to merging a PR, so do what works to ensure they're handled.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+    const pgClient = pgClientInstance ?? new PgClient({
+      user: databaseConnectionParameters.username,
+      password: databaseConnectionParameters.password,
+      host: process.env.PGHOST,
+      port: Number(databaseConnectionParameters.port),
+      database: databaseConnectionParameters.database,
+    });
+
+    this.pgClient = pgClient;
+    this.schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
```
darunrs, line 40: Morgan had a comment about making this a constructor parameter. I agree, you should do that. Expand the comments in the PR and you should see it. I unresolved it.

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,61 @@
+export const logsTableDDL = (schemaName: string) => `
```
darunrs, line 1: You need to add a return type for this function.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
+import format from 'pg-format';
+import { wrapError } from '../utility';
+import PgClient from '../pg-client';
+import { type DatabaseConnectionParameters } from '../provisioner/provisioner';
+import { LogLevel } from '../stream-handler/stream-handler';
+
+export interface LogEntry {
+  blockHeight: number;
```
darunrs, line 8: These semicolons are unnecessary

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+  message: string;
+}
+
+export enum LogType {
+  SYSTEM = 'system',
+  USER = 'user',
+}
+
+export default class IndexerLogger {
+
```
darunrs, line 21: :knife:

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+  private readonly pgClient: PgClient;
+  private readonly schemaName: string;
+
+  constructor (
+    functionName: string,
+    databaseConnectionParameters: DatabaseConnectionParameters,
+    pgClientInstance: PgClient | undefined = undefined
+  ) {
+
```
darunrs, line 30: :knife:

runner/src/indexer/indexer.ts
darunrs: You still have the issue showing. There's actually all sorts of ESLint errors in this file. Things like improper spacing, unnecessary separators (such as the comma here and the semicolons where you define some interfaces, etc). Please enable proper linting and take a look as it's too many to scour through and mark as comments in this PR.

runner/src/indexer/__snapshots__/indexer.test.ts.snap
```diff
@@ -1,215 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
Kevin101Zhang: I dont remember removing this - adding back

  postgres.Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM postgres:14
```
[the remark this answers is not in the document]
Kevin101Zhang: I just did a git merge and resolved the respective conflicts as Im more comfortable with that workflow to incorporate Morgan's changes

runner/src/indexer/indexer.ts
Kevin101Zhang: Got it! Will incorporate eslint more often

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  const functionName = 'testFunction';
+
+  describe('create', () => {
+    
+    it('should create an instance of IndexerLogger with provided PgClient instance', () => {
+      const indexerLogger = new IndexerLogger(functionName, mockDatabaseConnectionParameters, pgClient);
+      expect(indexerLogger).toBeInstanceOf(IndexerLogger);
+    });
+
+    it('should create an instance of IndexerLogger with a new PgClient instance if none provided', () => {
```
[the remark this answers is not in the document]
Kevin101Zhang: ok that makes sense Ill remove the 'create' portion

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            message: 'Test log message'
+        };
+
+        await indexerLogger.writeLog(logEntry);
+
+        expect(query).toHaveBeenCalled(); // Ensure query method is called
+    });
+
+    it('should handle errors when inserting a single log entry', async () => {
+        // Mocking pgClient.query to throw an error
```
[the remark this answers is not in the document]
Kevin101Zhang: ok that makes sense. Thanks for the insight!

runner/src/provisioner/schemas/logs-table.ts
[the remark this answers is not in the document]
Kevin101Zhang: morgan and I had a sync to just move forward with it reasoning that it was unlikely to conflict

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+  async writeLog(
+    logEntry: LogEntry
+  ): Promise<void> {
+    const { blockHeight, logTimestamp, logType, logLevel, message } = logEntry;
+    const logLevelString = LogLevel[logLevel];
+
+    const values = [blockHeight, logTimestamp, logTimestamp, logType, logLevelString, message];
+
+    const query = format(
+        `INSERT INTO %I.__logs (block_height, log_date, log_timestamp, log_type, log_level, message) VALUES %L`,
```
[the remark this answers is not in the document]
Kevin101Zhang: that makes lots of sense since its redundant and identical now.

runner/src/indexer/indexer.ts
```diff
@@ -72,43 +79,44 @@ export default class Indexer {
...
 
+        const runningMessage = `Running function ${functionName} on block ${blockHeight}, lag is: ${lag?.toString()}ms from block timestamp`;
+        simultaneousPromises.push(this.writeLogOld(LogLevel.INFO, functionName, blockHeight, runningMessage));
+        writeLogBatch.push({ blockHeight, logTimestamp: new Date(), logType: LogType.SYSTEM, logLevel: LogLevel.INFO, message: runningMessage });
         // Cache database credentials after provisioning
         const credentialsFetchSpan = this.tracer.startSpan('fetch database connection parameters');
         try {
-          this.database_connection_parameters = this.database_connection_parameters ??
-            await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName);
+          this.database_connection_parameters = this.database_connection_parameters ?? await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName) as DatabaseConnectionParameters;
```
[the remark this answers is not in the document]
Kevin101Zhang: done

runner/src/indexer-logger/indexer-logger.ts
Kevin101Zhang: Maybe Im a bit confused here but functionName I already modified functionName as a constructors parameter. 
Which then undergoes a sanitization for only non-alphanumeric characters. The sanitized result is stored in this.schemaName.

runner/src/hasura-client/hasura-client.ts
```diff
@@ -129,11 +129,11 @@ export default class HasuraClient {
     });
   }
 
-  async runMigrations (source: string, schemaName: string, migration: string): Promise<any> {
+  async runSql (source: string, schemaName: string, sqlScript: string): Promise<any> {
```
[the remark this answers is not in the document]
Kevin101Zhang: its a better name but there's some redundancy since we already have an executeSql method for hasuraClient. Which may lead to confusion. Happy to modify it to an alternative.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,79 @@
...
+        this.schemaName,
+        [values] 
+    );
+    await wrapError(async () => await this.pgClient.query(query), `Failed to insert log into ${this.schemaName}.__logs table`);
+  }
+
+  async writeLogBatch(logEntries: LogEntry[]): Promise<void> {
```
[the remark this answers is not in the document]
Kevin101Zhang: See comment below:

I can add a trace to the promise in the finally here and this should resolve the writeLogBatch which is now just a writeLog.

```
finally {
        await Promise.all([...simultaneousPromises, this.indexer_logger?.writeLog(logEntries)]);
      }
```

Kevin101Zhang:
I got some conflicting advice on a sync on thursday with Morgan about keeping the writeLog function but just have writeLogOld call writeLog as oppose to having many instances of promise.all([writeLogOld, writeLog])...

I see the above approach to encapsulate the logging further but indexer_behavior would have to be passed to IndexerLogger. Happy to go with this but just wanted to confirm this is what we want as this leads to some refactoring. We also have some issues with tracing right now as combining both calls writeLog/writeLogBatch in logger have some double tracing but this can be resolved by adding a trace to the finally block in runFunctions.

I think it easiest way to not batch the writeLogs in buildContext unless we declare it as a block scope const and execute it towards the end of the function. Unless we want to move the scoping of our current logEntries or repeatedly pass logEntries as a parameter.

Resolved the other comments.

runner/src/hasura-client/hasura-client.ts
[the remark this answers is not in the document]
darunrs: Sure that's fair. In that case, this functionName doesn't accurately represent what it does. It's a wrapper function over executeSql but it does so on a specific schema. So in that case, a better name is `executeSqlOnSchema` or something. runSql just sounds like its a duplicate of executeSql.

runner/src/indexer-logger/indexer-logger.ts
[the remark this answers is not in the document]
darunrs: I don't understand the comment. Are you saying you want to trace this function elsewhere? I'd suggest against that. We can explicitly trace this particular unit of work we care about (writing batch logs) and not need to worry about anything else. If we want to trace other things together, that's out of the scope of this class. I do a similar thing in provisioner, rather than tracing the provisioner chunk of code in Indexer as Indexer's code is bloated enough as it is. I'd like to have Indexer case more about tracing chunks of code, rather than tracing individual function calls. I sort of do the wrong thing with context.db in fact. I should have traced that in DmlHandler. 

What's I'd like to see is something like this:
```
const writeLogsSpan = this.tracer.startSpan('batch write logs');
try {
  await wrapError(DO THE QUERY HERE)
} finally {
  writeLogSpan.end();
}
```

  postgres.Dockerfile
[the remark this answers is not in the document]
darunrs: Yeah any approach works, so long as the github PR diff doesn't include work that isn't part of this PR. If adding this postgres file is part of this PR and not work that has been done in someone else's merged commit, then that's all good. Though, I believe Morgan's added this file already. So I should see a diff showing what changes you're making to this file. Not adding a new file, which is now incorrect.

runner/src/indexer/indexer.ts
```diff
@@ -391,32 +417,19 @@ export default class Indexer {
     }
   }
 
-  async writeLog (logLevel: LogLevel, functionName: string, blockHeight: number, ...message: any[]): Promise<any> {
+  async writeLog (logEntry: LogEntry): Promise<any> {
```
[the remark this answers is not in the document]
Kevin101Zhang: Resolving due to talking async meeting. Migrated logLevel and tracing to logger and writeLog will stay and call writeLogOld.

runner/src/indexer-logger/indexer-logger.ts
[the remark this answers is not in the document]
Kevin101Zhang: Ahh yeah I have it setup just like this. Was this comment before our sync?

runner/src/stream-handler/stream-handler.ts
Kevin101Zhang: In our newly proposed method - Im going to need to duplicate a function call writeLog to keep this functionality.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,77 @@
...
+        entry.logType,
+        LogLevel[entry.logLevel],
+        entry.message
+      ]);
+
+      const query = format(this.logInsertQueryTemplate, this.schemaName, values);
+      await this.pgClient.query(query);
+    }, `Failed to insert ${entriesArray.length > 1 ? 'logs' : 'log'} into the ${this.schemaName}.__logs table`)
+      .finally(() => {
+        writeLogSpan.end();
```
darunrs, line 74: Just to be sure, did you run this locally and check zipkin to see that the traces look correct?

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,77 @@
...
+    return logLevel >= this.loggingLevel;
+  }
+
+  async writeLogs (
+    logEntries: LogEntry | LogEntry[],
+  ): Promise<void> {
+    const entriesArray = (Array.isArray(logEntries) ? logEntries : [logEntries]).filter(entry => this.shouldLog(entry.logLevel)); ;
+    if (entriesArray.length === 0) return;
+
+    const spanMessage = `call writeLog function of IndexerLogger for ${entriesArray.length === 1 ? 'single entry' : `batch of ${entriesArray.length}`}`;
```
darunrs, line 57: Rename the trace to something like `write batch of {number of logs} logs through postgres`.

darunrs (commented):
Before we push this commit in, we need to run some code to create the logs tables and track them in Hasura. We should do so through a script so that we can duplicate the work in dev and prod with just a change of env variables used. 

I need to recover my local environment a bit but afterward I can write that script up and test it locally. We'll use it on dev and then we can proceed.

  postgres.Dockerfile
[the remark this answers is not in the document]
darunrs: So given that Morgan already added this file, let's remove this from the PR then.

docker-compose.yml
```diff
@@ -160,4 +162,4 @@ services:
 volumes:
   postgres:
   redis:
-  grafana:
+  grafana:
```
darunrs, line 165: Add an empty line at the end of this file

runner/src/indexer/indexer.ts
```diff
@@ -14,13 +17,15 @@ interface Dependencies {
...
+  IndexerLogger: typeof IndexerLogger
 };
 
 interface Context {
   graphql: (operation: string, variables?: Record<string, any>) => Promise<any>
   set: (key: string, value: any) => Promise<any>
   debug: (...log: any[]) => Promise<void>
   log: (...log: any[]) => Promise<void>
+  warn: (...log: any[]) => Promise<void>
   error: (...log: any[]) => Promise<void>
```
morgsmccauley, line 29:
```suggestion
  debug: (message: string) => Promise<void>
  log: (message: string) => Promise<void>
  warn: (message: string) => Promise<void>
  error: (message: string) => Promise<void>
```

The variadic arguments make the types really awkward - we need to use `any` as we can't guarantee what is passed in. I'm guessing these are only variadic because the original log function was. Since we are adding new methods it may be a good opportunity to just have a single typed argument like so.

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,57 @@
+export const logsTableDDL = (schemaName: string): string => `
+CREATE TABLE __logs (
+    id BIGSERIAL NOT NULL,
+    block_height NUMERIC(20) NOT NULL,
```
morgsmccauley, line 4: Can we make this nullable - some system logs won't necessarily have a block height

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,77 @@
...
+    this.schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
+    this.loggingLevel = loggingLevel;
+  }
+
+  private shouldLog (logLevel: LogLevel): boolean {
+    return logLevel >= this.loggingLevel;
+  }
+
+  async writeLogs (
+    logEntries: LogEntry | LogEntry[],
```
morgsmccauley, line 52:
```suggestion
    logEntries: LogEntry[],
```
Sorry, I know @darunrs already commented on this, but I would just make this take an array only. That way you don't need conditional logic to work with either type. The caller is still free to a single log entry, but it would have to be an array, which I don't think is too inconvenient.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,77 @@
+import format from 'pg-format';
+import { wrapError } from '../utility';
+import PgClient from '../pg-client';
+import { type DatabaseConnectionParameters } from '../provisioner/provisioner';
+import { LogLevel } from '../stream-handler/stream-handler';
+import { trace } from '@opentelemetry/api';
+
+export interface LogEntry {
+  blockHeight: number
+  logTimestamp: Date
```
morgsmccauley, line 10: From a usability standpoint, I think it's more convenient if we didn't need to pass this in, and instead we create the timestamp within this implementation. This would make the API cleaner and leaves less room for error.

The timestamp would need to correspond to when the `LogEntry` was created, rather than when the `writeLogs` method is called. That way batch logs, which are created ahead of time, would still have the correct timestamp. To achieve this, this could become a class:

```ts
class LogEntry {
  public readonly timestamp: Date;

  constructor (
    public readonly message: string,
    public readonly level: LogLevel,
    public readonly type: LogType,
    public readonly blockHeight?: number
  ) {
    this.timestamp = new Date();
  }

  static systemInfo (message: string, blockHeight?: number): LogEntry {
    return new LogEntry(message, LogLevel.INFO, LogType.SYSTEM, blockHeight);
  }
}
```
Which would be used like so:
```ts
const log = LogEntry.systemInfo('hello');
// later...
writeLog(log);
```

Note that passing an object to `writeLog` directly would still work, as that's all a `class` is really. The `static` methods could be expanded over time to provide convenience.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -0,0 +1,77 @@
...
+  logType: string
```
morgsmccauley, line 11:
```suggestion
  logType: LogType
```

runner/src/stream-handler/stream-handler.ts
```diff
@@ -13,8 +14,10 @@ export enum Status {
 export enum LogLevel {
```
morgsmccauley, line 14: It feels more appropriate for this to live in `IndexerLogger` now

runner/src/stream-handler/stream-handler.ts
```diff
@@ -13,8 +14,10 @@ export enum Status {
...
   DEBUG = 2,
   INFO = 5,
+  WARN = 6,
   ERROR = 8,
```
morgsmccauley, line 18: What do these numbers mean? Should we make them strings instead? e.g. `'ERROR'`

morgsmccauley (commented):
Sorry to create more work. One blocking comment around block height, the rest are just around code cleanliness.

runner/src/indexer-logger/indexer-logger.ts
darunrs: I kinda like the suggestion. The reason I went with Type | Type[] for DMlHandler was actually because its user facing and Monaco types can be wonky. So, I wanted to be flexible with the user code. In this case, the callers of this function are us, so we can expect that we write code correctly.

runner/src/stream-handler/stream-handler.ts
darunrs: This is an optimization I made. We need to compare the incoming log with the log level set in the indexer. So, if the log level was info, we reject debug. If it was error, we reject warn, info, debug. Making the log level a number inside Runner is helpful since I can make that comparison easier by jsut doing LogLevel <= LogLevelConfig.

runner/src/provisioner/schemas/logs-table.ts
Kevin101Zhang: @eduohe I dont see any issues with removing this - tagging to confirm

eduohe: no problem, we can remove the `NOT NULL` constraint

Kevin101Zhang:
Focusing on resolving test

runner/src/indexer-logger/indexer-logger.ts
Kevin101Zhang: @morgsmccauley There have been significant modifications in the PR already. Any thoughts on proceeding with the original changes but create a new PR to establish for this update

runner/tests/integration.test.ts
```diff
@@ -54,6 +54,32 @@ describe('Indexer integration', () => {
       database: postgresContainer.getDatabase(),
     });
 
+    const mockPgClient = {
```
morgsmccauley, line 57: The purpose of integration tests are for integrating with real components, can you please remove these mocks.

You can use the `hasuraClient` to get the real DB credentials and inject those in to `IndexerLogger`. Happy to sync up if you need help here.

runner/src/provisioner/schemas/logs-table.ts
```diff
@@ -0,0 +1,57 @@
...
+    level TEXT NOT NULL,
+    message TEXT NOT NULL,
+    PRIMARY KEY (date, id)
+) PARTITION BY RANGE (date);
+
+CREATE INDEX logs_timestamp_idx ON __logs USING btree (timestamp);
+CREATE INDEX logs_type_idx ON __logs USING btree (type);
+CREATE INDEX logs_level_idx ON __logs USING btree (level);
+CREATE INDEX logs_block_height_idx ON __logs USING btree (block_height);
+CREATE INDEX logs_search_vector_idx ON __logs USING GIN (to_tsvector('english', message));
```
morgsmccauley, line 17: I thought you prefixed these in an earlier commit but it seems to be gone now?

runner/src/provisioner/provisioner.ts
```diff
@@ -237,18 +243,20 @@ export default class Provisioner {
           }
 
           await this.createSchema(databaseName, schemaName);
-          await this.runMigrations(databaseName, schemaName, databaseSchema);
+
+          await this.runLogsSql(databaseName, schemaName);
```
morgsmccauley, line 247: This should be called within `setupPartitionedLogsTable()` and the line below uncommented.

runner/src/indexer/indexer.test.ts
```diff
@@ -1235,8 +1255,8 @@ CREATE TABLE
         })
       });
     const role = 'morgs_near';
-    const indexer = new Indexer(defaultIndexerBehavior, { fetch: mockFetch as unknown as typeof fetch, DmlHandler: genericMockDmlHandler }, undefined, undefined, config);
-    const context = indexer.buildContext(SIMPLE_SCHEMA, INDEXER_NAME, 1, HASURA_ROLE);
+    const indexer = new Indexer(defaultIndexerBehavior, { fetch: mockFetch as unknown as typeof fetch, DmlHandler: genericMockDmlHandler, IndexerLogger: genericMockIndexerLogger }, undefined, undefined, undefined, config);
+    const context = indexer.buildContext(SIMPLE_SCHEMA, INDEXER_NAME, 1, HASURA_ROLE, []);
```
morgsmccauley, line 1259: Could you please add some tests for logging itself? We should assert that the logs themselves actually translate to calls to Postgres

runner/src/indexer/__snapshots__/indexer.test.ts.snap
```diff
@@ -354,19 +354,6 @@ exports[`Indexer unit tests Indexer.runFunctions() catches errors 1`] = `
...
-      "body": "{"query":"\\n      mutation writeLog($function_name: String!, $block_height: numeric!, $message: String!){\\n          insert_indexer_log_entries_one(object: {function_name: $function_name, block_height: $block_height, message: $message}) {id}\\n      }","variables":{"function_name":"morgs.near/test","block_height":82699904,"message":"Running function morgs.near/test on block 82699904, lag is: NaNms from block timestamp"}}",
-      "headers": {
-        "Content-Type": "application/json",
-        "X-Hasura-Admin-Secret": "mock-hasura-secret",
-        "X-Hasura-Role": "append",
-        "X-Hasura-Use-Backend-Only-Permissions": "true",
-      },
-      "method": "POST",
-    },
-  ],
```
morgsmccauley, line 369: This removal means that this (old) log is no called, which is a bug. Do you know why this is the case?

morgsmccauley (dismissed):
There are a couple problems with this PR which I've commented inline.

This PR has become massive, and very confusing to track due to the number of comments. That's partly my fault, sorry. This definitely needs more work, but I think it would create more confusion if done here.

I'd recommend merging this with the main functionality commented out:
- Provisioning of the logs table (`runLogsSql()` call in `Provisioner`)
- Flushing the logs to the database (the `writeLogs` call in `Promise.all` in `Indexer`)

That way where not running any new and potentially problematic code, and can make changes to it under the hood.

With the above disabled, functionality should be exactly as it is on `main`. Therefore `indexer.test.ts` and `integration.test.ts` can be reverted to their original state and should pass. That should give us a fair amount of confidence that what we are pushing is not going to break anything.

With all that done, we should collate a list of pending tasks which we can break up in to future smaller PRs.

Let me know if you need help with any of this. I'm happy to sync up.

runner/src/indexer/indexer.test.ts
Kevin101Zhang: This is done in indexer-logger.test.ts

morgsmccauley: Ah, yes, you're right. But, we're not testing the `IndexerLogger` is actually being used within `Indexer`, i.e. If I have an Indexer which logs message, are we actually calling `IndexerLogger`? That flow is not covered.

runner/tests/integration.test.ts
Kevin101Zhang: Yeah will go ahead and use the real instance here

runner/src/indexer/indexer.test.ts
Kevin101Zhang: It's being covered along with loglevel. there's a specific area in the test called "writeLog is respecting level". Here the we create an indexer and initialize the logger in runfunction. We test to see if the log level is a certain level it calls writeLog in logger.

runner/src/stream-handler/stream-handler.ts
```diff
@@ -87,11 +83,17 @@ export default class StreamHandler {
     indexer.setStatus(functionName, 0, Status.STOPPED).catch((e) => {
       console.error(`Failed to set status STOPPED for stream: ${this.streamKey}`, e);
     });
-    indexer.writeLog(LogLevel.ERROR, functionName, this.executorContext.block_height,
-      `Encountered error processing stream: ${this.streamKey}, terminating thread\n${error.toString()}`
-    ).catch((e) => {
+    
+    indexer.callWriteLog({
```
darunrs, line 87: We should still make the original call, in addition to writeLog. Before this one, please still make the `writeLogOld` call through `indexer`.

runner/tests/integration.test.ts
darunrs, line 57: I don't think we're supposed to use mocks in an integration test. That defeats the point. That being said, I don't think this should be a blocker for you to get this PR out as it's gonna take you more time to figure. out usage of this. So, it's ok for now I think. Maybe @morgsmccauley can chime in here.

darunrs (commented):
Super duper close. Some problems with the implementation of testing and some cleanup you can do with the Indexer code.

runner/src/indexer/indexer.test.ts
darunrs: No what he means is we check how many times and what logs were given to IndexerLogger in Indexer. 

For example, some tests to verify:

We run an indexer which calls 5 logs, some with debug level with an info config. We then expect that IndexerLogger's writeLogs is called once. And then expect it was called with ALL FIVE LOGS since the filtering is done internally. 

In other words, we don't care if IndexerLogger does the right thing. But we do care if Indexer uses the logger class correctly. So, check scenarios like, the indexer throws an error but we still call writeLogs. Or we call it with all logs produced in various scenarios. Stuff like that.

runner/src/indexer/__snapshots__/indexer.test.ts.snap
Kevin101Zhang: Yeah my guess is this is something I missed due to stream handler directly interacting with the writeLogs and forgeting the writeLogsOld. Can go ahead and fix this

runner/src/stream-handler/stream-handler.ts
Kevin101Zhang: ah this slipped my mind - readding back - forgetting writeLogOld here is probably the reason the snapshot is missing a small snippert

darunrs:
Lots of remaining work. so we are going to do a phased rollout of these changes. 

1. We uncommented usage of the new IndexerLogger, but defined it here. All existing tests pass with few modifications and Runner works locally without issue.
2. We are going to modify https://github.com/near/queryapi/pull/636 to conditionally provision the logs table for all indexers. This ensure that we don't have to worry about it again, and we can revert the commit after. 
3. We re-enable usage of IndexerLogger, this time with fully provisioned logs tables. And we can verify the actual functionality works. And we can write up unit tests correctly for that.

darunrs (approved):

runner/src/provisioner/schemas/logs-table.ts
Kevin101Zhang: '__'

runner/src/indexer/indexer.test.ts
Kevin101Zhang: #4

Kevin101Zhang:
4 PRs.

1. Fix all unresolved comments in original PR
 - Prefix log indexes with __
 - Move creation of logs table to abstracted function
 - Update LogEntry to be a class
2. Uncomment provisioning of logs, add unit tests in Provisioner, and ensure integration tests pass
3. Ensure auto-provisioning of logs for existing users works, and merge https://github.com/near/queryapi/pull/636
4. Uncomment logging calls, update unit tests in Indexer, and ensure integration tests pass

PR #613, opened 2024-03-20
feat: Expose `near-lake-primitives` to VM

### LINKED ISSUE #593: Support from borsh-js
At [Devhub](https://github.com/NEAR-DevHub) we (@ailisp and I) are currently developing an indexer for our new proposal flow. We would greatly benefit if the queryapi would support importing [borsch-js](https://github.com/near/borsh?tab=readme-ov-file#implementations) in order to easily deserialize/serialize and read structures saved on chain.

<details>
<summary>Currently, we find ourselves manually writing functions to deserialize and serialize data structures saved on-chain, as illustrated below </summary>

```js
function buildAuthorToPostIdMap(block) {
  const stateChanges = block.streamerMessage.shards
    .flatMap((e) => e.stateChanges)
    .filter(
      (stateChange) =>
        stateChange.change.accountId === "devgovgigs.near" &&
        stateChange.type === "data_update"
    );
  const addOrEditPost = stateChanges
    .map((stateChange) => stateChange.change)
    .filter((change) => base64toHex(change.keyBase64).startsWith("05"))
    .map((c) => ({
      k: Buffer.from(c.keyBase64, "base64"),
      v: Buffer.from(c.valueBase64, "base64"),
    }));
  const authorToPostId = Object.fromEntries(
    addOrEditPost.map((kv) => [
      kv.v.slice(13, 13 + kv.v.slice(9, 13).readUInt32LE()).toString("utf-8"),
      Number(kv.k.slice(1).readBigUInt64LE()),
    ])
  );
  return authorToPostId;
}
```
</details>

It would be much easier to read and index the data if a library like borsch is supported.

@morgsmccauley Please let me know if this aligns with your interests, and I'm more than willing to make a pull request myself.

morgsmccauley:
- Upgrade `near-lake-primitives` to `0.2.0`, which includes `borsh`
- Expose entire `near-lake-primitives` library to VM via `primitives`, e.g. borsh can be accessed via `primitives.borsh.fromBorsh()`


runner/src/indexer/indexer.ts
```diff
@@ -282,7 +283,7 @@ export default class Indexer {
         const funcForTable = {
           [`${sanitizedTableName}`]: {
             insert: async (objectsToInsert: any) => {
-              return await this.tracer.startActiveSpan('Call context db insert', async (insertSpan: Span) => {
```
morgsmccauley, line 285: https://eslint.org/docs/latest/rules/no-return-await

runner/yarn.lock
```diff
@@ -1,6114 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
morgsmccauley, line 1: The project expects npm install and we should stick to that

runner/src/indexer/indexer.ts
darunrs: I'm a bit confused. The article you linked mentions "It can now be slower to remove await rather than keeping it." and also mentions the function not being present in the stack trace if something goes wrong. What benefits do we gain by removing the await? It says performance but it may not be the case anymore.

darunrs:
I believe we need to increase the primitives library version in the frontend too. I don't think we need to make any code changes though, since we pull the primitives type file from node nodules and add the whole file. Could be wrong on that though.

runner/src/indexer/indexer.ts
morgsmccauley: Oops, I completely missed that note 🤦🏼 

It seems `return await` is actually slightly faster. I'll revert and disable this rule.

morgsmccauley: Maybe my eslint was out of date, because it's actually requiring it now 🤷🏼. I'll revert this change, but keep the config as is.

morgsmccauley:
> I believe we need to increase the primitives library version in the frontend too. I don't think we need to make any code changes though, since we pull the primitives type file from node nodules and add the whole file. Could be wrong on that though.

Done, looks like it does pull from `node_modules` so shouldn't require any change

darunrs:
Sorry another thought just occurred to me. The reason we can freeze Block as "block" is because we import it through `import { Block } from "@near-lake/primitives";` and then have users write code in a function which surfaces it as `block` through `async function getBlock(block: Block) {` in the frontend. I believe we might need to make a change to the template code on the frontend to import whatever we want from primitives as `Primitives`. Otherwise, autocomplete might lead users to write code that doesn't reference primitives the way we were expecting.

morgsmccauley:
> Sorry another thought just occurred to me. The reason we can freeze Block as "block" is because we import it through `import { Block } from "@near-lake/primitives";` and then have users write code in a function which surfaces it as `block` through `async function getBlock(block: Block) {` in the frontend. I believe we might need to make a change to the template code on the frontend to import whatever we want from primitives as `Primitives`. Otherwise, autocomplete might lead users to write code that doesn't reference primitives the way we were expecting.

Good point, I don't think we need to update the template code though, we can add types like we do for `context`: https://github.com/near/queryapi/blob/main/frontend/src/utils/formatters.js#L71

morgsmccauley:
> Sorry another thought just occurred to me. The reason we can freeze Block as "block" is because we import it through `import { Block } from "@near-lake/primitives";` and then have users write code in a function which surfaces it as `block` through `async function getBlock(block: Block) {` in the frontend. I believe we might need to make a change to the template code on the frontend to import whatever we want from primitives as `Primitives`. Otherwise, autocomplete might lead users to write code that doesn't reference primitives the way we were expecting.

Good catch - we do need to update the template, most likely to `import * as primitives from "@near-lake/primitives"`, so they have access to the entire primitives library.

It also looks like [type declarations](https://github.com/near/queryapi/blob/main/frontend/primitives.d.ts) are manually defined/included, so those will need to be updated. Would be more convenient if we could pull from the library/`node_modules` directly, but I don't think that's possible with the multi-file output `near-lake-primitives` currently has. I'll see if I can get that working.

morgsmccauley:
I've updated the `primitives` (`near-lake-primitives`) types for Frontend - but this won't include the types for exported dependant packages, i.e. `borsher`. This exports are still usable, but won't be visible via frontend intellisense. Dependant packages are referenced via `import`s, which therefore assume that the library exists within `node_modules`, this is tricky to achieve with Monaco, so for now, all dependant packages will be left as `any`. We may revisit this again in future.

For reference, [borsher](https://github.com/nameskyteam/borsher) is exported under the `borsh` module within `near-lake-primitives`. So the entire library is accessible via `primitives.borsh.{borsher API}`, e.g. `primitives.borsh.borshSerialize()`

FYI @tguntenaar

PR #615, opened 2024-03-21
feat: Add pgBouncer to QueryApi

darunrs:
QueryApi has experienced issues with Postgres connections since the introduction of * indexers due to how QueryApi creates these connections through Application level connection pools. Since we can't use one pool for all workers, I've introduced PgBouncer as a Middleware to serve as an additional connection pooler in front of the DB.


init-scripts/postgres/init.sql
```diff
@@ -0,0 +1,12 @@
+CREATE ROLE pgbouncer LOGIN;
+ALTER ROLE pgbouncer WITH PASSWORD 'pgbouncer';
```
darunrs, line 2: Password should match DB_PASSWORD in docker container.
Note that this SQL needs to also be run in the psotgres instance. It's re-defined in this NEAR Ops PR: https://github.com/near/near-ops/pull/1652

PR #624, opened 2024-03-27
feat: Support Case Sensitive Schemas

### LINKED ISSUE #447: Indexers with quoted case-sensitive column names will have context.db calls fail against them

darunrs:
Indexer schemas can have quoted or unquoted table & column names. However, QueryApi always quotes table names and does not quote column names during SQL query construction for context.db. This is because the AST generated form parsing the schema does not include if the identifier was quoted or not. However, recent updates tot he parsing library has added this functionality in. 

I've updated QueryApi to quote both the table name and the column names if they were quoted originally, and leave them unquoted otherwise. 

In addition, I've replaced kevin-node-sql-parser back with the original package now that the 5.0 update has released. I've also added a typscript examples folder for convenience, as well as a script to clear the local postgres database.


scripts/wipe-database.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
darunrs, line 1: This is a helpful file for resetting the local PG database if it's ever out of sync with local hasura at any point.

darunrs:
I tested with the following schema and verified upsert works. Previously, we would get "relation does not exist" errors. 
```
CREATE TABLE
  Posts (
    "id" SERIAL NOT NULL,
    "AccountId" VARCHAR NOT NULL,
    BlockHeight DECIMAL(58, 0) NOT NULL,
    "receiptId" VARCHAR NOT NULL,
    content TEXT NOT NULL,
    block_Timestamp DECIMAL(20, 0) NOT NULL,
    "Accounts_Liked" JSONB NOT NULL DEFAULT '[]',
    "LastCommentTimestamp" DECIMAL(20, 0),
    CONSTRAINT "posts_pkey" PRIMARY KEY ("id")
  );

CREATE TABLE
  "CommentsTable" (
    "id" SERIAL NOT NULL,
    PostId SERIAL NOT NULL,
    "accountId" VARCHAR NOT NULL,
    blockHeight DECIMAL(58, 0) NOT NULL,
    CONSTRAINT "comments_pkey" PRIMARY KEY ("id")
  );
```

runner/src/indexer/indexer.ts
```diff
@@ -120,7 +125,7 @@ export default class Indexer {
         // TODO: Prevent unnecesary reruns of set status
         const resourceCreationSpan = this.tracer.startSpan('prepare vm and context to run indexer code');
         simultaneousPromises.push(this.setStatus(functionName, blockHeight, 'RUNNING'));
-        const vm = new VM({ timeout: 20000, allowAsync: true });
+        const vm = new VM({ allowAsync: true });
```
darunrs, line 128: I've removed the timeout as it doesn't even work with async code anyway, right now. I'm also hoping removing the timeout resolves the issues where Pavel's bitmap_v2 indexer is not being traced. 

I'll monitor this change and revert it if I need to.

runner/src/indexer/indexer.ts
```diff
@@ -215,31 +220,60 @@ export default class Indexer {
...
+        const columnNameDef = columnDef.column.column.expr;
+        const actualColumnName = columnNameDef.type === 'double_quote_string' ? `"${columnNameDef.value as string}"` : columnNameDef.value;
+        columnDefinitionNames.set(columnNameDef.value, actualColumnName);
+      }
+    }
+    return columnDefinitionNames;
+  }
+
+  private retainOriginalQuoting (schema: string, tableName: string): string {
+    const createTableQuotedRegex = `\\b(create|CREATE)\\s+(table|TABLE)\\s+"${tableName}"\\s*`;
```
darunrs, line 236: This regex just looks for a create table statement which uses the quoted tableName. If we find a match, the tableName was quoted.

runner/examples/list-executors.ts
```diff
@@ -0,0 +1,13 @@
+// Run with 'npx ts-node src/test-client.ts'
```
darunrs, line 1: I added typescript examples as well since they're tried to the proto file in the runner folder, and I find messing with them a bit easier.

PR #625, opened 2024-03-27
feat: Schedule log partition jobs during provisioning

### LINKED ISSUE #499: Setup Logs Retention to 2 weeks
Setup a cron job for every indexer to run every two weeks to delete old logs. 

https://hasura.io/docs/latest/api-reference/metadata-api/scheduled-triggers/#metadata-create-scheduled-event

morgsmccauley:
This PR expands provisioning to also schedule the cron jobs for adding/deleting log partitions. It assumes:
1. The `cron` database exists and has `pg_cron` enabled (https://github.com/near/near-ops/pull/1665)
2. The `__logs` table exists and has the partition functions defined (https://github.com/near/queryapi/pull/608)

In relation to this flow, the high-level steps are:
1. Use an admin connection to the `cron` database to grant the required access to the user
2. Use a user connection to the `cron` database to schedule the jobs

The cron job is executed under the user which schedules the job, therefore the user _must_ schedule the job as they are the only ones who have access to their schemas. If the admin were to schedule the job the job itself would fail as it doesn't have the required access.

Merging this before 2. is fine, the jobs will just fail, but should start to succeed after it has been implemented.

### DIFF
--- a/runner/src/provisioner/provisioner.ts
+++ b/runner/src/provisioner/provisioner.ts
@@ -161,12 +161,6 @@ export default class Provisioner {
             await this.addDatasource(userName, password, databaseName);
           }
 
-          // Untrack tables from old schema to prevent conflicts with new DB
-          if (await this.hasuraClient.doesSchemaExist(HasuraClient.DEFAULT_DATABASE, schemaName)) {
-            const tableNames = await this.getTableNames(schemaName, HasuraClient.DEFAULT_DATABASE);
-            await this.hasuraClient.untrackTables(HasuraClient.DEFAULT_DATABASE, schemaName, tableNames);
-          }
-
           await this.createSchema(databaseName, schemaName);
           await this.runMigrations(databaseName, schemaName, databaseSchema);
 
--- a/runner/src/provisioner/provisioner.ts
+++ b/runner/src/provisioner/provisioner.ts
@@ -1,4 +1,5 @@
 import { type Tracer, trace } from '@opentelemetry/api';
+import pgFormatLib from 'pg-format';
 
 import { wrapError } from '../utility';
 import cryptoModule from 'crypto';
@@ -8,7 +9,7 @@ import PgClient from '../pg-client';
 const DEFAULT_PASSWORD_LENGTH = 16;
 const CRON_DATABASE = 'cron';
 
-const _adminPgClient = new PgClient({
+const adminDefaultPgClientGlobal = new PgClient({
   user: process.env.PGUSER,
   password: process.env.PGPASSWORD,
   database: process.env.PGDATABASE,
@@ -16,7 +17,7 @@ const _adminPgClient = new PgClient({
   port: Number(process.env.PGPORT),
 });
 
-const _cronPgClient = new PgClient({
+const adminCronPgClientGlobal = new PgClient({
   user: process.env.PGUSER,
   password: process.env.PGPASSWORD,
   database: CRON_DATABASE,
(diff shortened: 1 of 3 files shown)


runner/src/provisioner/provisioner.ts
```diff
@@ -161,15 +200,12 @@ export default class Provisioner {
             await this.addDatasource(userName, password, databaseName);
           }
 
-          // Untrack tables from old schema to prevent conflicts with new DB
```
morgsmccauley, line 164: This was for when we moved from a single DB to separate user DBs and is no longer needed

runner/src/provisioner/provisioner.ts
```diff
@@ -1,19 +1,29 @@
...
 import cryptoModule from 'crypto';
 import HasuraClient from '../hasura-client';
 import PgClient from '../pg-client';
-import { type Tracer, trace } from '@opentelemetry/api';
 
 const DEFAULT_PASSWORD_LENGTH = 16;
+const CRON_DATABASE = 'cron';
 
-const sharedPgClient = new PgClient({
+const _adminPgClient = new PgClient({
```
morgsmccauley, line 11: Underscore to avoid conflicts with the class defined property

runner/src/provisioner/provisioner.ts
```diff
@@ -1,19 +1,29 @@
...
   user: process.env.PGUSER,
   password: process.env.PGPASSWORD,
   database: process.env.PGDATABASE,
   host: process.env.PGHOST,
   port: Number(process.env.PGPORT),
 });
 
+const _cronPgClient = new PgClient({
```
morgsmccauley, line 19: Admin user connected to the `cron` database

runner/src/provisioner/provisioner.ts
```diff
@@ -57,16 +64,48 @@ export default class Provisioner {
...
+        await this.cronPgClient.query(this.cronPgClient.format('GRANT EXECUTE ON FUNCTION cron.schedule_in_database TO %I;', userName));
+      },
+      'Failed to grant cron access'
+    );
+  }
+
+  async scheduleLogPartitionJobs (databaseName: string, schemaName: string): Promise<void> {
+    await wrapError(
+      async () => {
+        await this.cronPgClient.query(
```
morgsmccauley, line 92: TODO: This needs to be scheduled by the user and not admin

runner/src/provisioner/provisioner.ts
```diff
@@ -57,16 +64,48 @@ export default class Provisioner {
...
+          this.cronPgClient.format(
+            "SELECT cron.schedule_in_database('%1$I_logs_create_partition', '0 1 * * *', $$SELECT fn_create_partition('%1$I.__logs', CURRENT_DATE, '1 day', '2 day')$$, %2$I);",
+            schemaName,
+            databaseName
+          )
+        );
+        await this.cronPgClient.query(
+          this.cronPgClient.format(
+            "SELECT cron.schedule_in_database('%1$I_logs_delete_partition', '0 2 * * *', $$SELECT fn_delete_partition('%1$I.__logs', CURRENT_DATE, '-15 day', '-14 day')$$, %2$I);",
```
morgsmccauley, line 101: I may need to prefix the function call with the schema as this probably won't exist in the `public` schema

runner/src/provisioner/provisioner.ts
gabehamilton: Are you planning to make this user change in this PR or a separate PR?

morgsmccauley: This - I realised it was wrong only after raising the PR 😅

morgsmccauley: Done

eduohe (approved):

PR #627, opened 2024-03-29
test: Add integration tests for `Indexer`

### LINKED ISSUE #609: Implement Runner integration testing
Be able to execute a integration test suite through a single command, that relies on a running redis, postgres and hasura. Update README

morgsmccauley:
This PR adds a very basic integration test for `Indexer`. It uses `testcontainers` to stand up both `postgres` and `hasura` so that `Indexer` can talk to real components rather than mocks. The test uses `Indexer` directly, which means S3/Redis are still somewhat mocked/ignored. We can add those in later if need be.

This is essentially just the scaffolding for integration testing which can be expanded over time. The suite includes only 1 very basic test, which if successful should provide a fair amount of confidence that things are working as expected. The flow includes: provisioning, writing data to Postgres, and then asserting its existence via GraphQL. All errors bubble up from `Indexer` so this test should catch most problems.

This PR points to https://github.com/near/queryapi/pull/625, as so I could test the `pg_cron` flow via this integration test :)


hasura/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM hasura/graphql-engine:latest.cli-migrations-v3
```
morgsmccauley, line 1: Automatically applies locally stored metadata/migrations (i.e. logs/state tables, `append` role etc.) prior to standing up graphql server - probably only useful in tests right now

postgres.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM postgres:14
```
morgsmccauley, line 1: postgres image which enables `pg_cron`

runner/tsconfig.build.json
```diff
@@ -0,0 +1,5 @@
+{
```
morgsmccauley, line 1: Separate tsconfig to use when building, the default `tsconfig.json` should only be used locally now since it includes tests files etc.

runner/src/provisioner/provisioner.ts
```diff
@@ -96,9 +108,9 @@ export default class Provisioner {
         const userCronPgClient = new this.PgClient({
```
morgsmccauley, line 108: the connection parameters returned from hasura will include the `host`/`port` used within the docker network, therefore we need an override

runner/src/pg-client.ts
```diff
@@ -17,16 +17,20 @@ export default class PgClient {
...
       user: connectionParams.user,
       password: connectionParams.password,
       host: connectionParams.host,
-      port: Number(process.env.PGPORT),
+      port: Number(connectionParams.port),
       database: connectionParams.database,
       ...poolConfig,
     });
+
+    this.pgPool.on('error', onError);
```
morgsmccauley, line 32: `pgPool` will throw an error if the connection is closed from the server side - which will be the case during integration tests as the container will shut down first. Catching it here to prevent runtime errors.

runner/src/indexer/indexer.ts
```diff
@@ -49,6 +59,7 @@ export default class Indexer {
     deps?: Partial<Dependencies>,
     databaseConnectionParameters = undefined,
     dmlHandler = undefined,
+    private readonly config: Config = defaultConfig,
```
morgsmccauley, line 62: Need to parameterise these so they are configurable from test

runner/src/hasura-client/hasura-client.ts
```diff
@@ -22,13 +22,31 @@ type MetadataRequestArgs = Record<string, any>;
...
+};
+
 export default class HasuraClient {
   static DEFAULT_DATABASE = 'default';
   static DEFAULT_SCHEMA = 'public';
 
   private readonly deps: Dependencies;
 
-  constructor (deps?: Partial<Dependencies>) {
+  constructor (deps?: Partial<Dependencies>, private readonly config: Config = defaultConfig) {
```
morgsmccauley, line 49: parameterising so they are configurable from tests

gabehamilton (approved):
Cool!

postgres.Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN apt-get update && apt-get install -y postgresql-14-cron
+
+RUN echo "shared_preload_libraries = 'pg_cron'" >> /usr/share/postgresql/postgresql.conf.sample
+
+RUN echo "CREATE EXTENSION pg_cron;" > /docker-entrypoint-initdb.d/init-pg-cron.sql
```
darunrs, line 7: This probably works. Though, I actually have more SQL that's necessary to set up pgBouncer. I created [an sql file](https://github.com/near/queryapi/blob/main/init-scripts/postgres/init.sql) in queryApi which is copied into the docker entry point inside the docker compose. Since you're adding a dockerfile for postgres now, maybe we can merge the two approaches together?

runner/src/hasura-client/hasura-client.ts
```diff
@@ -10,17 +10,43 @@ interface SqlOptions {
   source?: string
 }
 
+interface DatabaseConnectionParameters {
```
darunrs, line 13: I actually export this interface in Indexer. You can import it here.

runner/src/provisioner/provisioner.ts
```diff
@@ -57,16 +79,68 @@ export default class Provisioner {
...
+          user: userDbConnectionParameters.username,
+          password: userDbConnectionParameters.password,
+          database: this.config.cronDatabase,
+          host: this.config.postgresHost ?? userDbConnectionParameters.host,
+          port: this.config.postgresPort ?? userDbConnectionParameters.port,
+        });
+
+        await userCronPgClient.query(
+          this.pgFormat(
+            "SELECT cron.schedule_in_database('%1$I_logs_create_partition', '0 1 * * *', $$SELECT fn_create_partition('%1$I.__logs', CURRENT_DATE, '1 day', '2 day')$$, %2$L);",
```
darunrs, line 118: Perhaps we can convert these into constants and label them something like CreatePartitionCronSql or something.

runner/tests/integration.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      {
+        hasuraAdminSecret: hasuraContainer.getAdminSecret(),
+        hasuraEndpoint: hasuraContainer.getEndpoint(),
+      }
+    );
+
+    await indexer.runFunctions(
+      Block.fromStreamerMessage(block1 as any as StreamerMessage),
+      {
+        'morgs.near/test': {
```
darunrs, line 86: This works fine for now. How are you thinking of organizing the integ tests once we have more test cases and test code? Maybe a new folder called integ/ and label the code according to the test? Same for sample block data. Maybe we can also leverage the block-server to download actual blocks too?

darunrs (approved):
Neat work! This will be fun to build on top of to get a good quite of integ tests going. 

Just some small comments.

runner/src/provisioner/provisioner.ts
morgsmccauley: This changes are from https://github.com/near/queryapi/pull/625 which this PR originally pointed to but has since been merged 😄

runner/tests/integration.test.ts
morgsmccauley: I hadn't put a lot of though in to that because I'm not entirely sure how these will evolve over time - we can tackle that when needed

runner/src/hasura-client/hasura-client.ts
morgsmccauley: Again related to https://github.com/near/queryapi/pull/625, I'll fix in a follow up PR

postgres.Dockerfile
morgsmccauley: Maybe we can create a top level `postgres/` directory which houses both this Dockerfile and the init script. I'll look at that in a follow up :)

PR #628, opened 2024-04-02
fix: Configure cron db from environment

morgsmccauley:
The admin/cron connection is currently hard-coded to the `cron` database, but this needs to be configurable so that we can use the default DB (`postgres`) locally.

Additionally, this PR combines the `pgbouncer` / `pg_cron` init scripts and uses the combined output in both docker compose and integration tests.

### DIFF
--- a/runner/src/provisioner/provisioner.ts
+++ b/runner/src/provisioner/provisioner.ts
@@ -34,13 +34,15 @@ export interface DatabaseConnectionParameters {
 
 interface Config {
   cronDatabase: string
-  // Allow overriding the default values for testing
-  postgresHost?: string
-  postgresPort?: number
+  // Override the host/port values returned by Hasura during testing/local development
+  hasuraHostOverride?: string
+  hasuraPortOverride?: number
 }
 
 const defaultConfig: Config = {
   cronDatabase: process.env.CRON_DATABASE,
+  hasuraHostOverride: process.env.HASURA_HOST_OVERRIDE,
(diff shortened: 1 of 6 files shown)


runner/src/provisioner/provisioner.ts
```diff
@@ -34,13 +34,15 @@ export interface DatabaseConnectionParameters {
...
-  postgresPort?: number
+  // Override the host/port values returned by Hasura during testing/local development
+  hasuraHostOverride?: string
+  hasuraPortOverride?: number
 }
 
 const defaultConfig: Config = {
   cronDatabase: process.env.CRON_DATABASE,
+  hasuraHostOverride: process.env.HASURA_HOST_OVERRIDE,
+  hasuraPortOverride: process.env.HASURA_POST_OVERRIDE ? Number(process.env.HASURA_POST_OVERRIDE) : undefined
```
darunrs, line 45: nit: `HASURA_PORT_OVERRIDE` PORT vs POST

darunrs (approved):
Nice! While you're here, you might want to also update the docker compose to specify your dockerfile instead of the postgres:14 image. I think the image tag is overriding the usage of your new postgres Dockerfile. 

Think just need a small update to the integ tests too.

runner/src/provisioner/provisioner.ts
morgsmccauley: Good catch, that's not a nit, that's a genuine error :D

PR #634, opened 2024-04-03
feat: Code for Set Status and Blockheight through Postgres

### LINKED ISSUE #588: Prepare code for setting Status and Last Processed BlockHeights through Postgres
- [x] Move indexer status into its own database from the `default` database
- [x] Code cleanup

Will do migrations in [this Issue](https://github.com/near/queryapi/issues/631) and implement functionality in this Issue.

darunrs:
I introduce the code necessary to perform status and last processed block height writes through Postgres. 

I also refactored DmlHandler and its usage in Indexer as caching of the database credentials allows for a simplification of its constructor.


runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -37,14 +39,14 @@ describe('IndexerLogger', () => {
 
       await indexerLogger.writeLogs(logEntry);
 
-      const expectedQueryStructure = `INSERT INTO "${functionName}".__logs (block_height, date, timestamp, type, level, message) VALUES`;
+      const expectedQueryStructure = `INSERT INTO ${schemaName}.__logs (block_height, date, timestamp, type, level, message) VALUES`;
```
darunrs, line 42: We create the schema using the name as is. In other words, without quotes. So, we are actually creating non case sensitive schema names. If a schema had a capital letter in it, and we used quotes, I believe there would be a problem. I need to do more testing on this, but for now, unquoted is better to match how we create the schema itself. 

I believe the refactoring that is planned will resolve this issue by uniformly providing schema names to all functions in Runner.

runner/src/indexer/indexer.test.ts
```diff
@@ -212,17 +213,14 @@ CREATE TABLE
...
-    format: jest.fn().mockReturnValue('mock')
-  } as unknown as PgClient;
-
-  const indexerLoggerInstance: any = new IndexerLogger(
-    INDEXER_NAME,
-    LogLevel.INFO,
-    genericDbCredentials,
-    mockPgClient
-  );
+  const indexerLoggerInstance: IndexerLogger = {
```
darunrs, line 216: These changes are ones I expect Kevin to do in his PR. Mainly, creating a mock class instead of mock dependencies for an actual Logger class. I'll rebase the changes he makes into this one when those are ready.

runner/src/indexer/indexer.ts
```diff
@@ -70,7 +70,7 @@ export default class Indexer {
     deps?: Partial<Dependencies>,
     databaseConnectionParameters = undefined,
     dmlHandler = undefined,
-    indexerLogger = undefined,
+    indexerLogger: IndexerLogger | undefined = undefined,
```
darunrs, line 73: It seems like TypeScript was thinking this was undefined type ONLY. I added a type here to allow for passing in the mock.

runner/src/provisioner/schemas/metadata-table.ts
```diff
@@ -0,0 +1,8 @@
+export const metadataTableDDL = (): string => `
+CREATE TABLE IF NOT EXISTS __metadata (
+    function_name TEXT NOT NULL,
```
pkudinov, line 3: @darunrs this is stored inside of a specific indexer schema, right? Do we need `function_name` in this case?

darunrs: Ah yes you're correct. We don't need that anymore. I'll work with Eduardo again to finalize this schema.

gabehamilton (approved):

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -127,5 +128,21 @@ describe('IndexerLogger', () => {
 
       expect(query).not.toHaveBeenCalled();
     });
+
+    it('log status for indexer', async () => {
+      const indexerLogger = new IndexerLogger(functionName, 5, mockDatabaseConnectionParameters, genericMockPgClient);
+      await indexerLogger.setIndexerStatus(IndexerStatus.RUNNING);
+      expect(query.mock.calls[0][0]).toEqual(
```
morgsmccauley, line 135:
```suggestion
      expect(query.mock).toBeCalledWith(
```
I believe this works too. Access the actual `calls` themselves is only really necessary if you need to check sequence or specific calls.

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -127,5 +128,21 @@ describe('IndexerLogger', () => {
 
       expect(query).not.toHaveBeenCalled();
     });
+
+    it('log status for indexer', async () => {
```
morgsmccauley, line 132:
```suggestion
    it('writes status for indexer', async () => {
```
`log` confused me here since we also log 😅

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -80,4 +92,28 @@ export default class IndexerLogger {
...
+
+    try {
+      await wrapError(async () => await this.pgClient.query(query), `Failed to update status for ${this.schemaName}`);
+    } finally {
+      setStatusSpan.end();
+    }
+  }
+
+  async updateIndexerBlockheight (blockHeight: number): Promise<void> {
+    const setLastProcessedBlockSpan = this.tracer.startSpan(`set last processed block to ${blockHeight} through postgres`);
```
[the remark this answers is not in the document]
darunrs: Good idea. I'll go ahead and give that a go.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -80,4 +92,28 @@ export default class IndexerLogger {
         writeLogSpan.end();
       });
   }
+
+  async setIndexerStatus (status: IndexerStatus): Promise<void> {
```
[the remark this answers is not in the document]
darunrs: That's true. I'll update!

runner/src/provisioner/schemas/metadata-table.ts
```diff
@@ -0,0 +1,8 @@
+export const metadataTableDDL = (): string => `
+CREATE TABLE IF NOT EXISTS __metadata (
+    instance TEXT NOT NULL,
```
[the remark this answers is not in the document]
darunrs: It's a column in the event that we intend to potentially have multiple instances of the same Indexer running. I'm still not sure how we want to implement that, but I added the column just to provide us something to work with when we get there.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -23,6 +30,11 @@ export enum LogType {
   SYSTEM = 'system',
   USER = 'user',
 }
+
+const METADATA_TABLE_UPSERT = 'INSERT INTO %I.__metadata (instance, attribute, value) VALUES %L ON CONFLICT (instance, attribute) DO UPDATE SET value = EXCLUDED.value RETURNING *';
+const STATUS_ATTRIBUTE = 'STATUS';
+const LAST_PROCESSED_BLOCK_HEIGHT_ATTRIBUTE = 'LAST_PROCESSED_BLOCK_HEIGHT';
+
 export default class IndexerLogger {
```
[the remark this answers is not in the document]
darunrs: I had a thought of naming it something like that previously. I agree it seems to fit a bit better. I'll do the rename then!

runner/src/indexer-logger/indexer-logger.test.ts
darunrs: Oops sorry haha. Will re-title both tests.

darunrs:
After refactoring DmlHandler a little to introduce tracing, I tested a local indexer calling most of the DB operations, and they all worked still. I did not need to change the unit tests at all either.

runner/src/indexer-logger/indexer-logger.ts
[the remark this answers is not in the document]
darunrs: Implemented and used! Ran some indexers locally and don't see any problems with the usage.

darunrs:
The wrapSpan stuff (The latest two commits) are a little bit heavy/out of focus. I think I'll remove those and put them in a followup PR.

runner/src/provisioner/schemas/metadata-table.ts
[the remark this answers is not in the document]
morgsmccauley: I see. Maybe this should be `instance_id` then? I think ultimately we don't really know what this is going to look like so I would just remove it. It would be trivial to add later so can just do it when we've got more certainty around what it looks like.

runner/src/provisioner/schemas/metadata-table.ts
```diff
@@ -0,0 +1,8 @@
...
+    attribute TEXT NOT NULL,
```
morgsmccauley, line 4: Do we want concrete fields (`current_block_height` etc.) rather than string values? 🤔

darunrs: I had a thought of doing concrete fields, but I wanted to avoid the scenario where we need to replace this table again to add any other metadata fields that we need. Or run some migration. I felt that just having an attribute value and a value value would be pretty extensible to any future needs. All we have to do is ensure we use the right attribute names which shouldn't be too hard I think.

runner/src/provisioner/schemas/metadata-table.ts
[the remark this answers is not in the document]
darunrs: Oh my impression was it wouldn't be all that trivial to add more columns down the line, to be honest. If it is trivial, then yeah I can definitely remove it. 

What would the process be for modifying this table down the line for any future needs look like? Our current process of migrating logs tables is a bit painful right now despite having few customers. Would adding new columns be a lot easier process wise?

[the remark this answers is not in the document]
morgsmccauley: Well I guess a bit more complicated than trivial - we would need a migration hook like we do for the creation of this table.

By creating this now, we run the risk of it being wrong, and having to change it later. But since I've labelled migrations as trivial that wouldn't be the end of the world haha.

Regardless, I'd opt to remove given we don't know how this will be used, and may just create confusion since it is unused right now. You're call though.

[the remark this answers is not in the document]
darunrs: Ok, I think that makes sense. Really it was a bit of a Hail Mary but you're right that we really don't know how it will go and I am worried of wanting to shoehorn a solution into using instance_id unnecessarily. I'll remove it, as I expect horizontal scaling will be hard no matter what we do!

morgsmccauley (approved):

PR #636, opened 2024-04-04
feat: Provision logs for existing users

### LINKED ISSUE #631: Migrate existing user schemas to contain logs and metadata tables
Conditionally provision logs and metadata table for all existing users.

morgsmccauley:
The provisioning flow will not be run for existing Indexers, this PR adds a separate provisioning check/step which sets up the partitioned logs table for existing users.

I've opted for a in-code approach as a "manual" migration script requires specific timing, i.e. we'd need to deploy the logs change, ensuring all new Indexers are provisioned correct, and then migrate all existing users to ensure that no Indexers are missed. But since the logs provisioning change is coupled with the logging itself, existing Indexers would fail to log until the migration is complete.

My only concern for this approach is a "thundering herd". After this is deployed, all Indexers will attempt to provision there logs table at the same time - I'll monitor this in Dev.

As this code is temporary, I didn't bother adding instrumentation/unit-tests, nor worry about the performance impact. It will be removed promptly.

This is dependant on #608 and should be merged after.


runner/tests/integration.test.ts
```diff
@@ -81,7 +82,37 @@ describe('Indexer integration', () => {
     );
 
     await indexer.runFunctions(
-      Block.fromStreamerMessage(block1 as any as StreamerMessage),
+      Block.fromStreamerMessage(block115185108 as any as StreamerMessage),
```
morgsmccauley, line 85: Many things _don't_ happen on the second execution, i.e. provisioning. Adding this as a sanity check for those.

runner/src/provisioner/provisioner.ts
```diff
@@ -239,7 +267,7 @@ export default class Provisioner {
           await this.createSchema(databaseName, schemaName);
           await this.runMigrations(databaseName, schemaName, databaseSchema);
 
-          // TODO re-enable once logs table is created
+          // TODO enable once new logs implementation is ready
           // await this.setupPartitionedLogsTable(userName, databaseName, schemaName);
```
morgsmccauley, line 271: This needs to happen **after** the existing provisioning step - the schema must be present for this to succeed

runner/src/provisioner/provisioner.ts
```diff
@@ -218,6 +218,34 @@ export default class Provisioner {
...
+    const sanitizedFunctionName = this.replaceSpecialChars(functionName);
+
+    const databaseName = sanitizedAccountId;
+    const userName = sanitizedAccountId;
+    const schemaName = `${sanitizedAccountId}_${sanitizedFunctionName}`;
+    const logsTable = '__logs';
+
+    await wrapError(
+      async () => {
+        const tableNames = await this.getTableNames(schemaName, databaseName);
```
darunrs, line 237: It depends on the plan for this code, but I can imagine it will stay in for a while (We need it to release it into prod). So, it's probably a good idea to cache the result of this in the class in the short term to prevent repeated hasura calls per invocation of runFunction.

runner/tests/integration.test.ts
darunrs: Now we have two tests which are dependent on each other order wise. I suppose the next test that goes in should be a separate file in that case?

morgsmccauley: This is still a single test, we just execute the Indexer twice.

There is so much setup involved in integration tests, a separate file would mean duplicating that, or extracting it. A single file avoids that, which we can probably live with for a while, until this becomes too unwieldy.

runner/src/provisioner/provisioner.ts
morgsmccauley: My intention was to remove this as soon as possible, but you're right, caching is a good idea. I'll do that.

PR #637, opened 2024-04-04
feat: Avoid unnecessary status updates

### LINKED ISSUE #604: Remove Unnecessary Status Writes

morgsmccauley:
Quick & dirty PR which short-circuits updating the status via GraphQL when it is unchanged. We don't need to check whether block height has changed, as that is updated in a separate call later on. I've also updated the integration tests to assert the output of status/logs.

### DIFF
--- a/runner/package-lock.json
+++ b/runner/package-lock.json
@@ -51,6 +51,7 @@
         "eslint-plugin-n": "^16.0.1",
         "eslint-plugin-prettier": "^5.0.0",
         "eslint-plugin-promise": "^6.1.1",
+        "graphql-request": "^6.1.0",
         "jest": "^29.6.2",
         "prettier": "^3.0.0",
         "testcontainers": "^10.7.2",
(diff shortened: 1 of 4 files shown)


darunrs:
I don't believe this would work for FAILING indexers. Since we write RUNNING, run the code, and when it fails, write FAILING. I suppose that's still fine since they're failing anyway, but I did want to call that out, if you have an idea on how to quickly cover that case too.

darunrs (approved):
Quick comment on failing indexers. Otherwise, should do the job!

morgsmccauley:
> I don't believe this would work for FAILING indexers. Since we write RUNNING, run the code, and when it fails, write FAILING. I suppose that's still fine since they're failing anyway, but I did want to call that out, if you have an idea on how to quickly cover that case too.

Good point, this only solves the happy path for when Indexers are continuously succeeding. In the failing case, we actually want to update: from `RUNNING` -> `FAILING`, and then `FAILING` -> `RUNNING` again. I guess the change would be to avoid setting it back to `RUNNING` until it has actually succeeded. But like you said, given they are failing we wouldn't be benefiting much.

I'd opt for moving forward with this, and waiting for your change to mitigate the above :)

PR #640, opened 2024-04-05
fix: Unresolved comments in #608

### LINKED ISSUE #639: Fix unresolved comments in original PR (https://github.com/near/queryapi/pull/608)
```[tasklist]
### Tasks
- [x] Prefix log indexes with __
- [x] Move creation of logs table to abstracted function
- [x] Update LogEntry to be a class -https://github.com/near/queryapi/pull/608#discussion_r1548710538
```

Kevin101Zhang:
Feat: created logEntry class and test cases
Chore:  relocated createLogs to abstracted func
Chore: renamed schema idx to prefix with '__'


runner/src/log-entry/log-entry.test.ts
```diff
@@ -0,0 +1,22 @@
+import LogEntry from './log-entry';
+import { LogType, LogLevel } from '../indexer-logger/indexer-logger';
+
+describe('LogEntry', () => {
+  test('should create a LogEntry instance with current timestamp', () => {
+    const currentTime = new Date();
+    const logEntry = new LogEntry('Test message', LogLevel.INFO, LogType.SYSTEM);
```
darunrs, line 7: Let's check to verify that the log entry also has the same message log level, and log type. Right now it's not hard to tell. We may complicate the constructor in the future, so we should ensure no side effects that are unexpected.

runner/src/log-entry/log-entry.ts
```diff
@@ -0,0 +1,22 @@
...
+  constructor (
+    public readonly message: string,
+    public readonly level: LogLevel,
+    public readonly type: LogType,
+    public readonly blockHeight?: number
+  ) {
+    this.timestamp = new Date();
+  }
+
+  static systemInfo (message: string, blockHeight?: number): LogEntry {
```
darunrs, line 15: We have plenty of system error logs that we need to write (provisioning, failing to get DB connections, failing on block, worker crashed, etc.). Let's add that.

runner/src/provisioner/provisioner.ts
```diff
@@ -137,7 +137,7 @@ export default class Provisioner {
   async setupPartitionedLogsTable (userName: string, databaseName: string, schemaName: string): Promise<void> {
     await wrapError(
       async () => {
-        // TODO: Create logs table
+        // await this.runLogsSql(databaseName, schemaName);
```
darunrs, line 140: Just an FYI, moving this call impacts Provisioner's unit tests. The tests rely on specifying responses from hasuraClient.executeSqlOnSchema in a certain order.

runner/src/log-entry/log-entry.ts
```diff
@@ -0,0 +1,22 @@
+import { LogType, LogLevel } from '../indexer-logger/indexer-logger';
```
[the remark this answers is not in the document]
Kevin101Zhang: Im under the impression after my last sync with @morgsmccauley that because this is going to role out in a specific order this is strictly for resolving the issues #1.

1. Fix all unresolved comments in original PR
Prefix log indexes with __
Move creation of logs table to abstracted function
Update LogEntry to be a class
2. Uncomment provisioning of logs, add unit tests in Provisioner, and ensure integration tests pass
3. Ensure auto-provisioning of logs for existing users works, and merge https://github.com/near/queryapi/pull/636
4. Uncomment logging calls, update unit tests in Indexer, and ensure integration tests pass

Because its rolling out in a specific order I would have to modify the indexer no matter what. Is there a reason we have to do it in step 1 as oppose to step 4. I would be modifying the index file regardless of it being commented out that would pollute the PR.

runner/src/log-entry/log-entry.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    expect(logEntry.timestamp).toBeInstanceOf(Date);
+    const timestampDifference = Math.abs(currentTime.getTime() - logEntry.timestamp.getTime());
+    expect(timestampDifference).toBeLessThanOrEqual(1000);
```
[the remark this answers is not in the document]
Kevin101Zhang: got it I can update this

runner/src/log-entry/log-entry.test.ts
Kevin101Zhang: ah ok. I initially had this but I removed it as I thought it wasnt necessary since it was provided through a constructor

runner/src/log-entry/log-entry.ts
Kevin101Zhang: that makes sense so we want additional static methods such as
systemError
systemWarn
systemDebug
I can do this but wondering why at that point would it be redundant since this can all be solved in one static method where the indexer can provide the LogLevel?

runner/src/log-entry/log-entry.ts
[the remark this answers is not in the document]
darunrs: You don't have to actually implement test and release that code. It was more like, we are adding this class, was it actually used locally as a test? To make sure it does everything we need? You can just re-comment out indexer again afterward. 

You mention below you used it so that's all good.

runner/src/log-entry/log-entry.ts
darunrs: Either option is fine really. Either you add the semantics to the function call or you add the semantics as a function parameter. Pick your poison I suppose. 

Just systemInfo on its own is insufficient.

runner/src/provisioner/provisioner.ts
darunrs: Hmm I also don't see a problem with keeping them separate. We shouldn't uncomment Provisioner until we're ready, and commented out code can't be tested.

runner/src/log-entry/log-entry.ts
morgsmccauley, line 1: This code is tightly coupled with `IndexerLogger` and is small enough that I think it could just exist within `IndexerLogger` - we will never use one without the other. Can we please move it there?

runner/src/log-entry/log-entry.test.ts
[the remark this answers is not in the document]
morgsmccauley: If you really want, you can use [Fake Timers](https://jestjs.io/docs/jest-object#fake-timers) to mock the `Date` object, and the assert the `LogEntry` was constructed with the mock date. But it's probably overkill.

runner/src/log-entry/log-entry.ts
```diff
@@ -0,0 +1,22 @@
...
+    public readonly blockHeight?: number
+  ) {
+    this.timestamp = new Date();
+  }
+
+  static systemInfo (message: string, blockHeight?: number): LogEntry {
+    return new LogEntry(message, LogLevel.INFO, LogType.SYSTEM, blockHeight);
+  }
+
+  static userLogs (message: string, level: LogLevel, blockHeight?: number): LogEntry {
```
morgsmccauley, line 19: Nit: The plural use here is a little misleading. I understand the intent, but it also makes me assume there are multiple elements.

I would probably just go ahead and create different user methods for all levels: `userInfo`, `userError`, `userWarn`. I think this read better and provide more convenience.

runner/src/log-entry/log-entry.ts
morgsmccauley: > I can do this but wondering why at that point would it be redundant since this can all be solved in one static method where the indexer can provide the LogLevel?

You're right that it seems somewhat redundant, but it's very likely that the code using this class would be larger than the code within this class itself, I think it will reduce the noise overall.

Having `LogLevel` as an argument means that _all_ callers need to import both `LogLevel` and `LogEntry`, and use both every time. Having a dedicated method for it means you only need `LogEntry`, and can use any of the convenience method defined.

runner/src/log-entry/log-entry.ts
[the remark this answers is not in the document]
morgsmccauley: I would go ahead and start using this class, even if the final result is commented out. We can worry about tests etc later, and just use TypeScript to guide/provide confidence. This will reduce the size of future PRs making things a bit easier.

morgsmccauley:
Can you please update the PR title to something like `fix: Unresolved comments in #608`

morgsmccauley:
@darunrs were there any other comments from you on #608 that should be resolved here?

runner/src/log-entry/log-entry.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor (
+    public readonly message: string,
+    public readonly level: LogLevel,
+    public readonly type: LogType,
+    public readonly blockHeight?: number
+  ) {
+    this.timestamp = new Date();
+  }
+
+  static createLog (message: string, level: LogLevel, type: LogType, blockHeight?: number): LogEntry {
```
Kevin101Zhang, line 15: Introduced a static create method that takes all the necessary parameters for creating a LogEntry instance. This method eliminates the redundancy. Each static method (systemDebug, systemInfo, systemWarn, etc) now just calls the create method with appropriate params, reducing duplication and making the code more DRY.

runner/src/log-entry/log-entry.ts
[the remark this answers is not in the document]
Kevin101Zhang: I integrated the LogEntry class and validated its functionality within the indexer-logger test suites. The test ran but do let me know if there are some additional things I should cover.

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -115,15 +82,9 @@ describe('IndexerLogger', () => {
...
-      const logEntry: LogEntry = {
-        blockHeight: 123,
-        logTimestamp: new Date(),
-        logType: LogType.SYSTEM,
-        logLevel: LogLevel.INFO,
-        message: 'Test log message'
-      };
-
-      await indexerLogger.writeLogs(logEntry);
+      const debugEntry = LogEntry.systemDebug('Debug message');
```
Kevin101Zhang, line 85: The IndexerLogger class is created with a specified logging level, and a log entry with a lower level is generated using the newly added LogEntry class

runner/src/log-entry/log-entry.ts
Kevin101Zhang: Yeah I agree it is tightly if not directly dependent and unusable without `IndexerLogger`. But I think until we have the full picture of LogEntry keeping it separate allows for easier management and potential future expansions. Towards the end of the Epic if there is still not purpose of keeping them separate I think we can move it back with few issues.

darunrs (commented):
Pretty close! I don't have any comments from 608 which needs to be addressed anymore here. Just some comments on data organization and such. Can approve after remaining comments are resolved.

runner/src/log-entry/log-entry.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    const logEntry = LogEntry.userWarn('User warn message', blockHeight);
+    expect(logEntry.message).toBe('User warn message');
+    expect(logEntry.level).toBe(LogLevel.WARN);
+    expect(logEntry.type).toBe(LogType.USER);
+    expect(logEntry.timestamp).toBeInstanceOf(Date);
+    expect(logEntry.blockHeight).toBe(blockHeight);
+  });
+
+  test('create a user error log entry', () => {
+    const blockHeight = 100;
```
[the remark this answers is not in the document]
Kevin101Zhang: I tested this actually and after they passed I manually added in the bh ( I reasoned that it would be more comprehensive to test it with all parameters rather than demonstrate a test with fewer parameters). My question then is should I add 8 more test to just have them without bh. Its odd to add a few since theyre all different static methods technically.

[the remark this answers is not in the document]
darunrs: Absolutely not. Add one more test and check that the value is correct. These are just wrapper functions over some underlying private function, which itself is a simple call to a constructor. We don't need to go hard on the unit tests here.

runner/src/indexer/indexer.test.ts
```diff
@@ -1263,4 +1263,4 @@ CREATE TABLE
             }
     ]);
   });
-});
\ No newline at end of file
+});
```
darunrs, line 1266: Add an empty at the end here.

runner/src/indexer-logger/indexer-logger.test.ts
```diff
@@ -30,8 +30,23 @@ describe('IndexerLogger', () => {
       const infoEntry = LogEntry.systemInfo('Info message');
       await indexerLogger.writeLogs(infoEntry);
 
-      const expectedQueryStructure = `INSERT INTO "${functionName}".__logs (block_height, date, timestamp, type, level, message) VALUES`;
-      expect(query.mock.calls[0][0]).toContain(expectedQueryStructure);
+      const timestampPattern = '\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\.\\d+\\+\\d{2}';
```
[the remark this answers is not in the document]
Kevin101Zhang: Mocking the timestamp for LogEntry instances may be difficult without modifying the constructor to accept an optional timestamp parameter. Since the timestamp is initialized within the constructor with the current system time by default.
```
  constructor (
    public readonly message: string,
    public readonly level: LogLevel,
    public readonly type: LogType,
    public readonly blockHeight?: number
  ) {
    this.timestamp = new Date();
  }
  
```

unless we make timestamp not a readonly or something like this
```
constructor(
    public readonly message: string,
    public readonly level: LogLevel,
    public readonly type: LogType,
    public readonly blockHeight?: number,
    timestamp?: Date // Optional parameter for specifying the timestamp
  ) {
    if (timestamp) {
      this.timestamp = timestamp; // Use the provided timestamp if available
    } else {
      this.timestamp = new Date(); // Otherwise, use the current system time
    }
```

runner/src/log-entry/log-entry.ts
morgsmccauley: We can still keep a separate file, but house it under `indexer-logger/`. It's only purpose is to be used within `indexer-logger` so it makes sense to have there.

I want to reduce the noise in the root directory. As this code base starts to grow we should be creating higher level abstractions, housing related functionality, to make navigation easier.

runner/src/indexer-logger/indexer-logger.test.ts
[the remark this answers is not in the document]
morgsmccauley: I mentioned in a previous [comment](https://github.com/near/queryapi/pull/640#discussion_r1555089601) we can use [Fake Timers](https://jestjs.io/docs/jest-object#fake-timers), looks like you resolved it but didn't actually implement? Can you make sure to only resolve if you have followed up?

We don't need to inject the `Date`/timestamp, we can mock the date like so:

```js
it('should insert a single log entry into the database', async () => {
  const date = new Date();

  jest.useFakeTimers({ now: date.getTime() });

  const indexerLogger = new IndexerLogger(functionName, LogLevel.INFO, mockDatabaseConnectionParameters, pgClient);
  await indexerLogger.writeLogs(LogEntry.systemInfo('Info message'));

  const expectedDate = date.toISOString().replace('T', ' ').replace('Z', '+00');

  expect(query.mock.calls[0][0])
    .toEqual(`INSERT INTO "${functionName}".__logs (block_height, date, timestamp, type, level, message) VALUES (NULL, '${expectedDate}', '${expectedDate}', 'system', 'INFO', 'Info message')`);
});
```

You may need to mock the `Date` at the top level, and pass the mocked date value around to avoid mock conflicts across tests.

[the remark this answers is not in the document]
darunrs: Adding an optional field to the parameter for the purpose of testing isn't outlandish. We do it quite often in fact. Honestly, you can also just say .contains( THE PART AFTER THE TIMESTAMP). The point I was making was we don't validate the query contains all of the expected values. As long as you expect on it, it's fine. 

Validating the timestamp in a unit test is a separate problem, and I'm ok not mixing them.

runner/src/indexer-logger/indexer-logger.ts
```diff
@@ -57,7 +40,7 @@ export default class IndexerLogger {
   async writeLogs (
     logEntries: LogEntry | LogEntry[],
   ): Promise<void> {
-    const entriesArray = (Array.isArray(logEntries) ? logEntries : [logEntries]).filter(entry => this.shouldLog(entry.logLevel)); ;
+    const entriesArray = (Array.isArray(logEntries) ? logEntries : [logEntries]).filter(entry => this.shouldLog(entry.level)); ;
```
morgsmccauley, line 43: On the previous PR I mentioned changing this `writeLogs` method to only take `LogEntry[]` so we don't need this logic. Can we go ahead and make that change please?

runner/src/indexer-logger/indexer-logger.test.ts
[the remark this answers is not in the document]
Kevin101Zhang: @morgsmccauley Ah I must have tunneled there and resolved it after I reading the initial comment. I didn't realize that implementation need jest to mock. Ill go ahead and mock it with FakeTimers

darunrs (commented):
LGTM! It does require a large rebase, so go ahead and do that and I
l'll approve after skimming it over!

morgsmccauley (approved):

darunrs (approved):

PR #646, opened 2024-04-08
refactor: Convert IndexerConfig to Class

### LINKED ISSUE #644: Refactor IndexerConfig into Class
Migrating any data related to the Indexer into a common class to simplify data interactions with things like AccountId, which are common.

darunrs:
Migrating any data related to the Indexer into a common class to simplify data interactions with things like AccountId, which are common.

I've also added an integ test for context DB.

### DIFF
--- a/coordinator/Cargo.lock
+++ b/coordinator/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "actix"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -921,6 +931,7 @@ dependencies = [
  "tonic 0.10.2",
  "tonic-build 0.10.2",
  "tracing",
+ "tracing-stackdriver",
  "tracing-subscriber",
  "wildmatch",
 ]
@@ -4359,6 +4370,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "tracing-serde"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bc6b213177105856957181934e4920de57730fc69bf42c37ee5bb664d406d9e1"
+dependencies = [
+ "serde",
+ "tracing-core",
+]
+
+[[package]]
+name = "tracing-stackdriver"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "80048836e000e1f058562f01d69cc46f476955bf389c0dc2d2d7edb98ca63ac1"
+dependencies = [
+ "Inflector",
+ "serde",
+ "serde_json",
+ "thiserror",
+ "time",
+ "tracing-core",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "tracing-subscriber"
 version = "0.3.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4368,12 +4404,15 @@ dependencies = [
  "nu-ansi-term",
  "once_cell",
  "regex",
+ "serde",
+ "serde_json",
  "sharded-slab",
  "smallvec",
  "thread_local",
  "tracing",
  "tracing-core",
  "tracing-log 0.2.0",
+ "tracing-serde",
 ]
 
 [[package]]
--- a/coordinator/Cargo.lock
+++ b/coordinator/Cargo.lock
@@ -3,16 +3,6 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
-dependencies = [
- "lazy_static",
- "regex",
-]
-
-[[package]]
 name = "actix"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -931,7 +921,6 @@ dependencies = [
  "tonic 0.10.2",
  "tonic-build 0.10.2",
  "tracing",
- "tracing-stackdriver",
  "tracing-subscriber",
  "wildmatch",
 ]
@@ -4370,31 +4359,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "tracing-serde"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bc6b213177105856957181934e4920de57730fc69bf42c37ee5bb664d406d9e1"
-dependencies = [
- "serde",
- "tracing-core",
-]
-
-[[package]]
-name = "tracing-stackdriver"
-version = "0.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "80048836e000e1f058562f01d69cc46f476955bf389c0dc2d2d7edb98ca63ac1"
-dependencies = [
- "Inflector",
- "serde",
- "serde_json",
- "thiserror",
- "time",
- "tracing-core",
- "tracing-subscriber",
-]
-
-[[package]]
 name = "tracing-subscriber"
 version = "0.3.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4404,15 +4368,12 @@ dependencies = [
  "nu-ansi-term",
  "once_cell",
  "regex",
- "serde",
- "serde_json",
  "sharded-slab",
  "smallvec",
  "thread_local",
  "tracing",
  "tracing-core",
  "tracing-log 0.2.0",
- "tracing-serde",
 ]
 
 [[package]]
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -10,7 +10,7 @@ interface SqlOptions {
   source?: string
 }
 
-export interface DatabaseConnectionParameters {
+export interface HasuraDatabaseConnectionParameters {
   password: string
   database: string
   username: string
@@ -125,7 +125,7 @@ export default class HasuraClient {
     return metadata;
   }
 
-  async getDbConnectionParameters (account: string): Promise<DatabaseConnectionParameters> {
+  async getDbConnectionParameters (account: string): Promise<HasuraDatabaseConnectionParameters> {
     const metadata = await this.exportMetadata();
     const source = metadata.sources.find((source: { name: any, configuration: any }) => source.name === account);
     if (source === undefined) {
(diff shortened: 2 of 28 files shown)


runner/src/indexer-config/indexer-config.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  fullName (): string {
+    return `${this.accountId}/${this.functionName}`;
+  }
+
+  hasuraRoleName (): string {
+    return this.sanitizeNameForDatabase(this.accountId);
+  }
+
+  hasuraFunctionName (): string {
```
darunrs, line 62: I made these all different functions to add some semantic meaning since we seem to be treating these values between hasura and postgres differently. It's probably *possible* for them to diverge. 

What I was thinking was to simply have say sanitizedAccount, sanitizedIndexerName, sanitizedFullName and use those instead. What do you think?

runner/src/indexer/indexer.test.ts
```diff
@@ -363,20 +348,6 @@ CREATE TABLE
     expect(mockFetch.mock.calls).toMatchSnapshot();
   });
 
-  test('Indexer.buildContext() throws when a GraphQL response contains errors', async () => {
```
darunrs, line 366: Moved to indexer config

runner/src/indexer/indexer.test.ts
```diff
@@ -1164,41 +1096,6 @@ CREATE TABLE
     expect(mockFetchError).toHaveBeenCalledTimes(3); // 1 log + 2 graphql call
   });
 
-  test('does not attach the hasura admin secret header when no role specified', async () => {
```
darunrs, line 1167: I don't understand the point of this test. It seems like it supports a use case where we create Context without a role. We never do this, as far as I can tell. Please do let me know if there's something here I should be supporting and how.

runner/src/indexer/indexer.ts
```diff
@@ -95,108 +83,85 @@ export default class Indexer {
...
-          }
-        }
+    try {
+      const runningMessage = `Running function ${this.indexerConfig.fullName()} on block ${blockHeight}, lag is: ${lag?.toString()}ms from block timestamp`;
+      simultaneousPromises.push(this.writeLog(LogLevel.INFO, blockHeight, runningMessage));
 
-        // logEntries.push({ blockHeight, logTimestamp: new Date(), logType: LogType.SYSTEM, logLevel: LogLevel.INFO, message: runningMessage });
-        // Cache database credentials after provisioning
-        const credentialsFetchSpan = this.tracer.startSpan('fetch database connection parameters');
+      if (options.provision) {
```
darunrs, line 90: We originally were also checking if the passed in config was true/false as well. We do not currently use this feature. So, I did not add it to IndexerConfig. We can add it in at a later date when Runner is being potentially told by Coordinator if provisioning is necessary.

runner/src/indexer-config/indexer-config.ts
```diff
@@ -0,0 +1,115 @@
...
+      startExecutorRequest.accountId,
+      startExecutorRequest.functionName,
+      parseInt(startExecutorRequest.version),
+      startExecutorRequest.code,
+      startExecutorRequest.schema,
+      LogLevel.INFO
+    );
+  }
+
+  static fromObject (data: IndexerConfigData): IndexerConfig {
```
darunrs, line 44: Necessary for transferring class from StreamHandler into Worker as it gets serialized into an object no matter what.

runner/src/indexer/indexer.test.ts
morgsmccauley: This case is a protection, rather than testing actual functionality. While we aren't actively doing this, if someone were to (perhaps accidentally) make that change, this test would catch it and complain. Think of it as protecting against potential bugs.

In regards to the test itself, this is a security concern. If the secret is attached without the role then the request will be assumed as admin, granting the Indexer access to _all_ data.

runner/src/indexer/indexer.ts
```diff
@@ -84,8 +74,6 @@ export default class Indexer {
 
   async runFunctions (
```
morgsmccauley, line 75: This name is inaccurate now, how about `execute`? :)

morgsmccauley (approved):
This is amazing, great job on this! Just a few small comments.

In terms of conventional commits, this isn't really a `feat:` it's a `refactor:`, could you update the PR title? :)

runner/src/indexer-config/indexer-config.ts
```diff
@@ -0,0 +1,115 @@
...
+    return {
+      redisStreamKey: this.redisStreamKey,
+      accountId: this.accountId,
+      functionName: this.functionName,
+      version: this.version,
+      code: this.code,
+      schema: this.schema,
+      logLevel: this.logLevel
+    };
+  }
```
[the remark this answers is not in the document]
darunrs: The problem is that when I get it back in worker, worker continues to treat it as a class object, but it no longer has any functions. I was getting errors when I was trying to do things like indexerConfig.fullName().

runner/src/indexer/indexer.ts
darunrs: Will do!

runner/src/indexer-config/indexer-config.ts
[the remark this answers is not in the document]
morgsmccauley: Oh yeah that makes sense - I guess it is serialised to an object when it's passed across the thread bounday.

morgsmccauley:
> Should I add that indexer into the integration tests?

@darunrs Yup, definitely. We don't test DML handler at all via integration tests atm.

runner/src/indexer/indexer.test.ts
darunrs: Ohh ok gotcha. Right now it isn't possible for the Context object to *not* pass this value in anymore as the only parameter passed into buildContext() is blockHeight. We could theoretically remove this check outright. 

What I can do is set an optional parameter in buildContext for hasura role which can be set to null. But I also don't see a use case for that right now. 

For now, I'll add the test back but directly call runGraphQL instead to confirm this fallback works. But we're no longer testing context building anymore.

runner/src/hasura-client/hasura-client.ts
```diff
@@ -10,7 +10,7 @@ interface SqlOptions {
   source?: string
 }
 
-interface DatabaseConnectionParameters {
+export interface DatabaseConnectionParameters {
```
darunrs, line 13: I've removed This same interface from Provisioner. Now the workflow is this:
1. Hasura gets DatabaseConnectionParameters.
2. Provisioner's wrapper function over the hasura call converts it to PostgresConnectionParameters which is an exported interface from PgClient. 
3. The PgClient params are what get passed around, allowing for a simple pass in, into the constructor.

darunrs:
I've changed the usage of env variables a bit now. We now have three pairs of host/port evn variables to be used in QueryApi:
Running Runner locally
- PGHOST/PORT: localhost/5432
- PGHOST/PORT_PGBOUNCER: localhost/6432
- PGHOST/PORT_HASURA: postgres/5432

Running Runner in Docker Compose
- PGHOST/PORT: postgres/5432
- PGHOST/PORT_PGBOUNCER: pgbouncer/6432
- PGHOST/PORT_HASURA: postgres/5432

Running Runner IntegTest
- PGHOST/PORT: localhost/host machine postgres container port
- PGHOST/PORT_PGBOUNCER: localhost/host machine postgres container port
- PGHOST/PORT_HASURA: postgres/testcontainer's network port

Running QueryApi in dev/prod
- PGHOST/PORT: docker public ip/port
- PGHOST/PORT_PGBOUNCER: 0.0.0.0/6432
- PGHOST/PORT_HASURA: docker public ip/port

runner/src/indexer/indexer.test.ts
morgsmccauley: While it might not be possible now, the code can always change, making it possible in future. I'd opt to leave this in, it doesn't add much overhead, but still protects us against this security issue. Testing `runGraphQL` directly like you said makes most sense though!

coordinator/Cargo.lock
```diff
@@ -2,6 +2,16 @@
 # It is not intended for manual editing.
```
morgsmccauley, line 2: Do you mean to commit this?

runner/src/provisioner/provisioner.ts
```diff
@@ -108,14 +105,10 @@ export default class Provisioner {
...
-        const userDbConnectionParameters = await this.hasuraClient.getDbConnectionParameters(userName);
-        const userCronPgClient = new this.PgClient({
-          user: userDbConnectionParameters.username,
-          password: userDbConnectionParameters.password,
-          database: this.config.cronDatabase,
-          host: this.config.hasuraHostOverride ?? userDbConnectionParameters.host,
-          port: this.config.hasuraPortOverride ?? userDbConnectionParameters.port,
-        });
+        const userDbConnectionParameters = await this.getPostgresConnectionParameters(userName);
+        userDbConnectionParameters.database = this.config.cronDatabase;
```
morgsmccauley, line 109:
```suggestion
        const userDbConnectionParameters = {
          ...(await this.getPostgresConnectionParameters(userName)),
          database: this.config.cronDatabase
        };
```

Nit: In this case it's probably safe to mutate, but it's generally a better idea to avoid it and create new objects.

runner/src/utility.ts
```diff
@@ -4,6 +4,7 @@ export async function wrapError<T> (fn: () => Promise<T>, errorMessage: string):
   try {
     return await fn();
   } catch (error) {
+    console.log('ERROR', error);
```
morgsmccauley, line 7: 🔪

runner/tests/integration.test.ts
```diff
@@ -68,92 +69,53 @@ describe('Indexer integration', () => {
...
-        }
-      },
-      false,
-      {
-        provision: true
-      }
-    );
-
-    await indexer.runFunctions(
-      Block.fromStreamerMessage(block115185109 as any as StreamerMessage),
```
morgsmccauley, line 133: Running 2 blocks was intentional, because we have a lot of logic which does not run the second time round. This ensures that works.

coordinator/Cargo.lock
darunrs: Ah no. let me remove that.

runner/tests/integration.test.ts
```diff
@@ -189,6 +151,143 @@ describe('Indexer integration', () => {
...
+      indexerConfig,
+      {
+        provisioner
+      },
+      undefined,
+      {
+        hasuraAdminSecret: hasuraContainer.getAdminSecret(),
+        hasuraEndpoint: hasuraContainer.getEndpoint(),
+        hasuraHostOverride: postgresContainer.getIpAddress(),
+        hasuraPortOverride: Number(postgresContainer.getPort())
```
morgsmccauley, line 254: These are unused/unneeded now right? We can remove them from `Indexer`

runner/src/indexer/indexer.ts
morgsmccauley: Should we just remove this option too?

runner/tests/integration.test.ts
darunrs: Gotcha. I'll add it back!

runner/tests/integration.test.ts
darunrs: Gotcha! Will remove.

runner/src/indexer/indexer.ts
darunrs: Hmm yeah we don't use this either. I'll remove it!

morgsmccauley (approved):

PR #648, opened 2024-04-08
fix: type generation on load

### LINKED ISSUE #629: Types are not generated onLoad

Kevin101Zhang:
The object post astify() we receive from node-sql-parser changed. We can think about Version Control in the future.
Quick fix here: returning the object to how it was and adding an additional layer on Editor to ensure mounting of types


frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -159,17 +162,21 @@ export class PgSchemaTypeGen {
...
+			Object.prototype.hasOwnProperty.call(columnDef, "nullable")
+		columnDef.nullable &&
+			columnDef.nullable.value === "not null";
 		return isPrimaryKey || isNullable ? false : true;
 	}
 
 	getRequiredStatus(columnDef, nullable) {
 		const hasDefaultValue =
-			columnDef.hasOwnProperty("default_val") && columnDef.default_val != null;
+			Object.prototype.hasOwnProperty.call(columnDef, "default_val") &&
```
Kevin101Zhang, line 177: Modified access method to avoid 'shadowing'

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -83,32 +83,34 @@ export class PgSchemaTypeGen {
 	generateTypes(sqlSchema) {
 		const schemaSyntaxTree = this.parser.astify(sqlSchema, { database: "Postgresql" });
 		const dbSchema = {};
-
+		console.log(schemaSyntaxTree)
```
morgsmccauley, line 86: 🔪 ?

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -141,11 +143,12 @@ export class PgSchemaTypeGen {
 	}
 
 	addColumn(columnDef, columns) {
-		const columnName = columnDef.column.column.expr.value;
+		const columnName = columnDef.column.column;
+		console.log('found the issue ', columnName)
```
morgsmccauley, line 147: 🔪

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -159,17 +162,21 @@ export class PgSchemaTypeGen {
 
 	getNullableStatus(columnDef) {
 		const isPrimaryKey =
-			columnDef.hasOwnProperty("unique_or_primary") &&
-			columnDef.unique_or_primary == "primary key";
+			Object.prototype.hasOwnProperty.call(columnDef, "unique_or_primary") &&
```
morgsmccauley, line 165: Why do we have to call the `prototype` directly?

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -159,17 +162,21 @@ export class PgSchemaTypeGen {
...
+			columnDef.unique_or_primary &&
```
morgsmccauley, line 166: Why do we have this check when we already check if the property exists above?

frontend/src/utils/pgSchemaTypeGen.js
Kevin101Zhang: To avoid shadowing the code
```

let obj = {
  hasOwnProperty: function() {
    return false; // Shadowing the built-in hasOwnProperty method
  },
  foo: 'bar'
};

console.log(obj.hasOwnProperty('foo')); // false

```

frontend/src/utils/pgSchemaTypeGen.js
Kevin101Zhang: Im not sure I understand. Are you asking why we have both 
```
Object.prototype.hasOwnProperty.call(columnDef, "unique_or_primary") &&
'columnDef.unique_or_primary'
```
The first portion is checking if the object has a property named "unique_or_primary" on itself, rather than inherited prototype chain. The second is just verifying that it holds a value that is considered truthy.

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -83,32 +83,33 @@ export class PgSchemaTypeGen {
...
 						for (const foreignKeyDef of columnSpec.definition) {
-							columns[foreignKeyDef.column.expr.value].nullable = false;
+							columns[foreignKeyDef.column].nullable = false;
 						}
 					}
 				}
 				dbSchema[tableName] = columns;
 			} else if (statement.type === "alter") {
 				// Process ALTER TABLE statements
+				let newConstraint = {};
```
darunrs, line 112: What's the point in moving the contraint variable outside the for loop? It's scope is to apply to the alterSpec in question.

frontend/src/components/Editor/Editor.jsx
```diff
@@ -180,6 +180,7 @@ const Editor = ({ actionButtonText }) => {
   useEffect(() => {
     localStorage.setItem(SCHEMA_TYPES_STORAGE_KEY, schemaTypes);
     attachTypesToMonaco();
+    handleCodeGen();
```
darunrs, line 183: handleCodeGen already attaches types to monaco. Let's remove the above call if we're doing this call here.

frontend/src/utils/pgSchemaTypeGen.js
Kevin101Zhang: This was just due to a stricter eslint configuration where there was a lexical declaration inside a case block. To fix the issue, I moved the lexical declaration outside of the case block.

frontend/src/utils/pgSchemaTypeGen.js
morgsmccauley: You're right, I read that wrong 😅

frontend/src/components/Editor/Editor.jsx
Kevin101Zhang: thats a good catch. removed

Kevin101Zhang:
Sorry, let this sit and forgot about rerequesting are review here. Are we good to merge here?

frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -141,11 +142,11 @@ export class PgSchemaTypeGen {
 	}
 
 	addColumn(columnDef, columns) {
-		const columnName = columnDef.column.column.expr.value;
+		const columnName = columnDef.column.column;
```
darunrs, line 145: I this correct? I thought the location of the column name was changed after the update? Can you verify this is the right way to get the name?

darunrs (approved):
LGTM! Just one callout to verify something before merging.

frontend/src/utils/pgSchemaTypeGen.js
Kevin101Zhang: Yep! This should be it! I verified locally with the latest version we have of node-sql-parser and the object we from astify.

Kevin101Zhang: confirmed by pulling and it works for me

PR #649, opened 2024-04-08
feat: Introduce wrapSpan utility and Refactor DmlHandler tracing

darunrs:
Many areas in Runner involve adding a try finally block to support tracing it. I added a wrapSpan utility function to simplify the code a little bit. I also refactored DmlHandler's tracing. I was previously tracing it in Indexer which adds to clutter and isn't the right place for it. I instead added a private function which does the tracing for DmlHandler functions. I also add the SQL query as an attribute for the span.

### DIFF
--- a/runner/src/dml-handler/dml-handler.ts
+++ b/runner/src/dml-handler/dml-handler.ts
@@ -1,10 +1,10 @@
 import { wrapError } from '../utility';
 import PgClient, { type PostgresConnectionParams } from '../pg-client';
 import { type TableDefinitionNames } from '../indexer';
+import type IndexerConfig from '../indexer-config/indexer-config';
 
 import { type Tracer, trace, type Span } from '@opentelemetry/api';
 import { type QueryResult } from 'pg';
-import type IndexerConfig from '../indexer-config/indexer-config';
 
 type WhereClauseMulti = Record<string, (string | number | Array<string | number>)>;
 type WhereClauseSingle = Record<string, (string | number)>;
--- a/runner/src/dml-handler/dml-handler.ts
+++ b/runner/src/dml-handler/dml-handler.ts
@@ -2,7 +2,9 @@ import { wrapError } from '../utility';
 import PgClient, { type PostgresConnectionParams } from '../pg-client';
 import { type TableDefinitionNames } from '../indexer';
 
-import { type Tracer, trace } from '@opentelemetry/api';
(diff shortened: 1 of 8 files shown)


runner/src/dml-handler/dml-handler.ts
```diff
@@ -1,14 +1,17 @@
 import { wrapError } from '../utility';
+import { trace, type Span, type Tracer } from '@opentelemetry/api';
 import PgClient from '../pg-client';
 import { type DatabaseConnectionParameters } from '../provisioner/provisioner';
 import { type TableDefinitionNames } from '../indexer';
+import { type QueryResult } from 'pg';
```
morgsmccauley, line 6: Nit: Can we keep absolute (dependencies) and relative (local code) imports separate? I think there is an eslint rule for it :D

runner/src/indexer/indexer.ts
```diff
@@ -121,19 +123,18 @@ export default class Indexer {
...
-          this.deps.dmlHandler ??= new DmlHandler(this.database_connection_parameters);
-        } catch (e) {
-          const error = e as Error;
-          await this.writeLog(LogLevel.ERROR, functionName, blockHeight, 'Failed to get database connection parameters', error.message);
-          // logEntries.push({ blockHeight, logTimestamp: new Date(), logType: LogType.SYSTEM, logLevel: LogLevel.ERROR, message: `Failed to get database connection parameters ${error.message}` });
-          throw error;
-        } finally {
-          credentialsFetchSpan.end();
-        }
+        await wrapSpan(this.tracer, 'fetch database connection parameters', async () => {
```
morgsmccauley, line 126: We could take this a step further, and introduce a `Tracer` wrapper class, has the `wrapSpan()` method, that way we wouldn't need to pass `tracer` every time?

morgsmccauley: Definitely a big nitpick, but I prefer the function as first parameter, just to be consistent with `wrapError` 😅

darunrs: The general practice for tracers is to create them in whatever context needs them, rather than trying to explicitly manage their instances. This is help let the tracer's context propagation do the heavy lifting. I'm not sure how creating a shared class will impact that. I'd probably rather just continue to create tracers where necessary, with any local context attached (Like what class is using it) whenever I do so. 

And sure, I can make the function the first parameter lol.

morgsmccauley (approved):

PR #643, opened 2024-04-05
feat: Introduce provisioning of Logs Table for new and existing users

Kevin101Zhang:
First few commits is from this branch https://github.com/near/queryapi/pull/640. Created this branch based off the initial branch.

This PR intends to introduce the creation of the logs table by provisioning the logsSchema and the follow CRON jobs for new Users but does not use or writeLogs to the new logsTable itself.

If this is merged by itself new users will have unused log table but the provisioning will occur.
To provision existing users - https://github.com/near/queryapi/pull/636

PR #657, opened 2024-04-10
feat: Enable Logging functionality to both new and old Log Tables

### LINKED ISSUE #650: Release Logs to Production (both old and new way)

Kevin101Zhang:
Uncommented functionality so we actually start writing logs to new the Tables that have been provisioned in #643.
Old logging implementation remains untouched as still functions (although it has been renamed from writeLog -> writeLogOld). We are writing to both log tables.

### 1. Provisioning and Logging (to both tables) for a new Indexer
https://www.loom.com/share/3ad6d6ea3368412e8896340a74759ffb?sid=4d5379e8-5401-41bf-9e38-d0f8e8c4eca5

### 2. Logging (to both tables) for a existing Indexer
https://www.loom.com/share/4ba411f2bcb740e1842650f695ffb347?sid=253ced68-9d4c-459f-871b-b0a3ee00cd91

### Provisioning and Logging new logs table for a existing Indexer (that does not have logs table)
https://www.loom.com/share/2aa7c0cc882f4dbdb9e51fc2a9e9b7b9?sid=1aa511fe-3054-4d27-9996-2b9fddc44ed8


runner/src/indexer/indexer.ts
```diff
@@ -490,15 +475,18 @@ export default class Indexer {
...
+    if (blockHeight) {
+      return await this.writeLogOld(level, functionName, blockHeight, message);
+    }
+  }
 
-  // async callWriteLog (logEntry: LogEntry): Promise<any> {
-  //   await (this.deps.indexerMeta as IndexerMeta).writeLogs([logEntry]);
-  // }
+  // onetime use method to allow stream-handler to writeLog into new log table in case of failure
+  async callWriteLog (logEntry: LogEntry): Promise<any> {
```
Kevin101Zhang, line 487: streamHandler is exclusively using this method if it to directly call writeLogs in case of an error processing the streams. Since streamhandler does not currently have access to an instance of IndexerMeta (we can refactor in the future), we use the indexer that it does have access to and call a helper function to writeLog directly.

runner/src/indexer/indexer.test.ts
```diff
@@ -1017,7 +1017,7 @@ CREATE TABLE
     await indexer.runFunctions(mockBlock, functions, false, { provision: true });
 
     expect(provisioner.provisionUserApi).not.toHaveBeenCalled();
-    expect(provisioner.getDatabaseConnectionParameters).toHaveBeenCalledTimes(1);
+    expect(provisioner.getDatabaseConnectionParameters).toHaveBeenCalledTimes(3); // todo: temporary fix for the test to ensure new logging works
```
Kevin101Zhang, line 1020: // todo: temporary fix for the test to ensure new logging functionality all works

this.database_connection_parameters is no longer cached due to getDatabaseConnection is no longer cached

Kevin101Zhang:
Tackel: https://github.com/near/queryapi/pull/643#discussion_r1560017523

runner/src/indexer/indexer.ts
```diff
@@ -96,54 +96,53 @@ export default class Indexer {
...
-        // logEntries.push(runningLogEntry);
+        const runningLogEntry = LogEntry.systemInfo(runningMessage, blockHeight);
+        logEntries.push(runningLogEntry);
         // Cache database credentials after provisioning
         const credentialsFetchSpan = this.tracer.startSpan('fetch database connection parameters');
         try {
-          this.database_connection_parameters ??= await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName) as DatabaseConnectionParameters;
-          // this.database_connection_parameters = await this.getDatabaseConnectionParams(hasuraRoleName);
-          // this.deps.indexerMeta ??= new IndexerMeta(functionName, this.indexer_behavior.log_level, this.database_connection_parameters);
+          // this.database_connection_parameters ??= await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName) as DatabaseConnectionParameters;
```
morgsmccauley, line 137: We can remove this now right?

runner/src/indexer/indexer.ts
```diff
@@ -490,15 +475,18 @@ export default class Indexer {
...
 
-  // async writeLog (logEntry: LogEntry, logEntries: LogEntry[], functionName: string): Promise<any> {
-  //   logEntries.push(logEntry);
-  //   const { logLevel, blockHeight, message } = logEntry;
-  //   return await this.writeLogOld(logLevel, functionName, blockHeight, message);
-  // }
+  async writeLog (logEntry: LogEntry, logEntries: LogEntry[], functionName: string): Promise<any> {
+    logEntries.push(logEntry);
+    const { level, blockHeight, message } = logEntry;
+    if (blockHeight) {
```
morgsmccauley, line 481: Why do we only call the old logging method if there is a `blockHeight`? Shouldn't we always call it?

morgsmccauley (changes requested):

runner/src/indexer/indexer.ts
Kevin101Zhang: Sorry I should have made a comment here. I only did this if statement as a form of assertion due to the new log table having an optional bh but the old method having a required bh param.

runner/src/indexer/indexer.test.ts
```diff
@@ -1248,4 +1248,53 @@ CREATE TABLE
...
+    const indexerDebug = new Indexer(
+      { log_level: LogLevel.DEBUG },
+      { fetch: mockFetchDebug as unknown as typeof fetch, provisioner: genericProvisioner, dmlHandler: genericMockDmlHandler, indexerMeta },
+      undefined,
+      config
+    );
+
+    await indexerDebug.runFunctions(mockBlock, functions, false);
+    expect(indexerMeta.writeLogs).toHaveBeenCalledTimes(1);
+    expect(indexerMeta.writeLogs.mock.calls[0][0]).toHaveLength(5);
```
Kevin101Zhang, line 1298: I aimed to verify the the method calls, but had trouble due to slight timestamp variations. This would be a more thorough test.
```
 expect(jest.fn()).toBeCalledWith(...expected)

    - Expected
    + Received

    @@ -8,30 +8,30 @@
        },
        LogEntry {
          "blockHeight": 456,
          "level": 2,
          "message": "debug log",
    -     "timestamp": 2024-04-14T18:54:21.734Z,
    +     "timestamp": 2024-04-14T18:54:21.735Z,
          "type": "system",
        },
        LogEntry {
          "blockHeight": 456,
          "level": 5,
          "message": "info log",
    -     "timestamp": 2024-04-14T18:54:21.734Z,
    +     "timestamp": 2024-04-14T18:54:21.735Z,
          "type": "system",
        },
        LogEntry {
          "blockHeight": 456,
          "level": 8,
          "message": "error log",
    -     "timestamp": 2024-04-14T18:54:21.734Z,
    +     "timestamp": 2024-04-14T18:54:21.735Z,
          "type": "system",
        },
        LogEntry {
          "blockHeight": 456,
          "level": 2,
          "message": "Selecting objects in table posts with values {\"account_id\":\"morgs_near\",\"receipt_id\":\"abc\"} with no limit",
    -     "timestamp": 2024-04-14T18:54:21.734Z,
    +     "timestamp": 2024-04-14T18:54:21.735Z,
          "type": "system",
        },
      ],
```

runner/tests/integration.test.ts
```diff
@@ -190,5 +179,15 @@ describe('Indexer integration', () => {
...
+
+    const { morgs_near_test___logs: _logs }: any = await graphqlClient.request(gql`
+    query {
+      morgs_near_test___logs {
+        message
+      }
+    }
+  `);
+
+  expect(_logs.length).toEqual(4);
```
Kevin101Zhang, line 191: the length of logs from indexer_log_entries and the new logs table (morgs_near_test___logs) should be identical.

runner/src/indexer/indexer.ts
```diff
@@ -181,22 +179,22 @@ export default class Indexer {
...
-  //   const { username, password, database, host, port } = await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName) as DatabaseConnectionParameters;
-  //   return {
-  //     username,
-  //     password,
-  //     database,
-  //     host: this.config.hasuraHostOverride ?? host,
-  //     port: this.config.hasuraPortOverride ?? port
-  //   }
-  // }
+  async getDatabaseConnectionParams (hasuraRoleName: string): Promise<DatabaseConnectionParameters> {
```
darunrs, line 188: This is not necessary with the changes I've made.

runner/src/indexer/indexer.ts
```diff
@@ -96,54 +96,52 @@ export default class Indexer {
...
-        // logEntries.push(runningLogEntry);
+        const runningLogEntry = LogEntry.systemInfo(runningMessage, blockHeight);
+        logEntries.push(runningLogEntry);
         // Cache database credentials after provisioning
         const credentialsFetchSpan = this.tracer.startSpan('fetch database connection parameters');
         try {
-          this.database_connection_parameters ??= await this.deps.provisioner.getDatabaseConnectionParameters(hasuraRoleName) as DatabaseConnectionParameters;
-          // this.database_connection_parameters = await this.getDatabaseConnectionParams(hasuraRoleName);
-          // this.deps.indexerMeta ??= new IndexerMeta(functionName, this.indexer_behavior.log_level, this.database_connection_parameters);
+          this.database_connection_parameters ??= await this.getDatabaseConnectionParams(hasuraRoleName);
```
darunrs, line 137: Not necessary, due to some updates made to Runner. A successful rebase will allow you to use provisioner's function.

runner/src/indexer/indexer.ts
darunrs: There's some confusion in your implementation about writeLog and writeLogOld. We ahve two separate functions. Are you trying to have writeLog also call writeLogOld? If so, writeLog should add to the log entries array and then call writeLogOld. And then we call writeLogs function on IndexerMeta at the end. 

Or, if they're supposed to be separate, writeLog should not call writeLogOld. What's your intended implementation? The end result should be that ALL logs are added to the log entries array, and also calling writeLogOld. And then at the end, we run indexerMeta.writeLogs(THE LOGS ARRAY). 

So, pick one:
A. writeLog adds to array, then calls writeLogOld
B. We only have writeLogOld and adding to logEntries.

runner/tests/integration.test.ts
darunrs: Let's also check the result to see if we see all expected logs. Things like provisioning logs, running function on ... logs. We can ignore the date.

runner/tests/integration.test.ts
```diff
@@ -149,6 +148,27 @@ describe('Indexer integration', () => {
...
+    
+    const { morgs_near_test___logs: provisioning_endpoints }: any = await graphqlClient.request(gql`
+      query MyQuery($searchString: [String!]) {
+        morgs_near_test___logs(where: {message: {_ilike: "%Provisioning endpoint%"}}) {
+          message
+        }
+      }
+    `);
+    
+    expect(provisioning_endpoints.length).toEqual(2);
```
Kevin101Zhang, line 170: {
        "message": "Provisioning endpoint: starting"
      },
      {
        "message": "Provisioning endpoint: successful"
      }

runner/tests/integration.test.ts
```diff
@@ -149,6 +148,37 @@ describe('Indexer integration', () => {
...
+
+    const { morgs_near_test___logs: running_function_enpoint }: any = await graphqlClient.request(gql`
+      query {
+        morgs_near_test___logs(where: {message: {_ilike: "%Running function%"}}) {
+          message
+        }
+      }
+    `);
+    
+    expect(running_function_enpoint.length).toEqual(2);
```
Kevin101Zhang, line 180: We run on 2 blocks
    await indexer.execute(Block.fromStreamerMessage(block115185108 as any as StreamerMessage));
    await indexer.execute(Block.fromStreamerMessage(block115185109 as any as StreamerMessage));

runner/tests/integration.test.ts
Kevin101Zhang: done I also checked locally if specific columns are null or not. I can add it to the test as well but im not sure if this is necessary as its not done with indexer_log_entries.

runner/src/indexer/indexer.ts
Kevin101Zhang: Im confused about this comment here -.

The method writeLog's goal is to append a logEntry into logEntries[] and then call writeLogOld.

within direct block scope of runFunction we have access to both simultaneous promise array and logEntries array as they are declared locally.

Here we just push any log entries to an array for both to later be called in the finally so we do not use the writeLog method at all.


We start using writeLog in the other functional context - buildContext and buildDatabaseContext, originally they were just calling the old writeLog method-> now renamed to writeLogOld. This function has not changed besides the naming and its purpose is to run a graphQLQuery. 

In the current code now we just added a param logEntries to both buildContext and buildDatabaseContext every time we want to write a log as the previous code want we just run it through the new writeLog method which does two things:
1. appends a log entry to the logEntries array to later be resolved in the finally block of runFunctions
2. call the writeLogOld to execute the graphQL query.

Kevin101Zhang: I believe its already doing option A well here unless im missing something

darunrs: It looks odd as we explicitly call writeLogOld and then do logEntries.push earlier up (e.g. in Provisioning), and do writeLog() itself later on (e.g. context logs). I suppose this is fine? 

I would like better would be to have writeLog take in a logEntry, a blockheight, logEntries array, and simultaneous promises array. You can add logEntry to the logEntries array (It may or may not have a BH). Then, you can add to simultaneous promises the call to writeLog Old which takes attributes of logEntry along with the passed in block height. This would result in writeLog fully encapsulating all our behavior, allowing us to keep writeLog calls exactly where they were originally while keeping the behavior. We also don't need to worry as much about optional blockheights. This is what I would consider to be a fully implemented option A.

Maybe @morgsmccauley can chime in here if I'm going off the track.

Kevin101Zhang: I understand where you are coming from. If this is permanent code I also believe that it would be great to pass in simultaneousPromises as a Param into buildcontext and builddatabasecontext methods and just push to the respective logEntries/simultaneous promises arrays. This way the functionality is strictly appending entries to be later resolved in the finally block in run functions. But then I thought since this is all temporary and the next PR on sprint would involve removing the hasura writes it did not make sense to continue.

morgsmccauley: Agree that this is a bit convoluted as we have two implementations going on. I would say let's defer tidying/optimising while this is in flux, and we can clean up once we remove the old implementation.

runner/tests/integration.test.ts
```diff
@@ -149,6 +148,37 @@ describe('Indexer integration', () => {
     `);
 
     expect(logs.length).toEqual(4);
+
+    const { morgs_near_test___logs: _logs }: any = await graphqlClient.request(gql`
```
morgsmccauley, line 152:
```suggestion
    const { morgs_near_test___logs: logs }: any = await graphqlClient.request(gql`
```
The prefix here is really necessary, it's only really important at the DB level

morgsmccauley (approved):
1 minor comment. Pre-emptively approving on the basis that mine/@darunrs comments are resolved before merging :).

runner/tests/integration.test.ts
Kevin101Zhang: I didn't use this naming due to the above code 
`const { indexer_log_entries: logs }: any = await graphqlClient.request(gql`
what I will do is rename this to `old_logs`. so this can be `logs` as suggested and remove old_logs as apart of the removal PR

runner/src/indexer-meta/indexer-meta.test.ts
```diff
@@ -23,16 +24,17 @@ describe('IndexerMeta', () => {
     port: 5432,
     database: 'test_database'
   };
-  const functionName = 'some_account/some_indexer';
-  const schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
+
+  const indexerConfig = new IndexerConfig('', '', 'some_account/some_indexer', 0, '', '', LogLevel.INFO);
```
darunrs, line 28: This is the wrong way to use IndexerConfig. You should be passing in the accountID, and then functionName. Not pass in schema name. That is now a generated value based on accountID and function name. Also, you do not need to use underscore sin the names anymore as that's cleaned up as part of the internal behavior. You can do some-account and some-indexer and it will still return some_account/some_indexer as the schema.

runner/src/indexer-meta/indexer-meta.ts
```diff
@@ -25,16 +26,15 @@ export default class IndexerMeta {
...
     databaseConnectionParameters: PostgresConnectionParams,
     pgClientInstance: PgClient | undefined = undefined
   ) {
     const pgClient = pgClientInstance ?? new PgClient(databaseConnectionParameters);
 
     this.pgClient = pgClient;
-    this.schemaName = functionName.replace(/[^a-zA-Z0-9]/g, '_');
-    this.loggingLevel = loggingLevel;
+    this.schemaName = indexerConfig.schemaName();
+    this.loggingLevel = indexerConfig.logLevel;
```
darunrs, line 37: This is not necessary. Set indexerConfig as private readonly in the constructor and you can access it using this.indexerConfig. This also allows you to use all of its methods. Just do inline function calls below.

runner/src/indexer/indexer.ts
```diff
@@ -156,12 +159,12 @@ export default class Indexer {
       await this.setStatus(blockHeight, IndexerStatus.FAILING);
       throw e;
     } finally {
-      await Promise.all([...simultaneousPromises]);
+      await Promise.all([...simultaneousPromises, (this.deps.indexerMeta as IndexerMeta).writeLogs(logEntries)]);
```
darunrs, line 162: So we have an issue here. Basically, writeLogs could potentially fail because provisioning failed. If that happens, would the error from this Promise.all() lead to the writeLogs writing error overwrite provisioning errors? It's worth having a unit test checking for this behavior as we want the provisioning error to be logged at least to the console, if it's overwritten. If it isn't overwritten, this is fine.

runner/src/indexer/indexer.ts
```diff
@@ -316,10 +315,8 @@ export default class Indexer {
               return await this.tracer.startActiveSpan('Call context db insert', async (insertSpan: Span) => {
                 try {
                   // Write log before calling insert
-                  await this.writeLog(LogLevel.DEBUG, blockHeight,
-                    `Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`);
-                  // const insertLogEntry = LogEntry.systemDebug(`Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`, blockHeight);
-                  // await this.writeLog(insertLogEntry, logEntries, functionName);
+                  const insertLogEntry = LogEntry.systemDebug(`Inserting object ${JSON.stringify(objectsToInsert)} into table ${tableName}`, blockHeight);
```
darunrs, line 318: There are not really system logs. They're logs describing stuff that the user is doing. Set them to user logs instead.

darunrs (approved):
Mostly good. Some comments to address. Can be merged after resolving the below comments.

runner/src/indexer/indexer.ts
Kevin101Zhang: Yeah thats a good thought - In this case they are not overwritten 

if the provisioning process fails, the error is appropriately handled within a catch on 105-109.
the code structure has provisioning errors immediately logged and handled within that try-catch block there, this is distinct from subsequent logging operations. There shouldnt be a risk of overwriting provisioning errors.

runner/src/indexer-meta/indexer-meta.ts
```diff
@@ -33,12 +32,11 @@ export default class IndexerMeta {
     const pgClient = pgClientInstance ?? new PgClient(databaseConnectionParameters);
 
     this.pgClient = pgClient;
-    this.schemaName = indexerConfig.schemaName();
-    this.loggingLevel = indexerConfig.logLevel;
+    this.indexerConfig = indexerConfig;
```
darunrs, line 35: This is not necessary. Using private readonly will automatically expose indexerConfig to the class as a class variable.

PR #658, opened 2024-04-11
feat: Conditionally provision metadata table

### LINKED ISSUE #651: Provision metadata tables
Provision metadata table for new indexers and conditional re-provisioning for existing ones

darunrs:
Enable conditional provisioning of metadata table.

### DIFF
--- a/coordinator/Cargo.lock
+++ b/coordinator/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "actix"
 version = "0.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


runner/src/indexer/indexer.ts
```diff
@@ -50,6 +50,7 @@ const defaultConfig: Config = {
 
 export default class Indexer {
   DEFAULT_HASURA_ROLE: string;
+  LOGGGED_CONTEXT_DB_WARNING: boolean = false;
```
darunrs, line 53: We keep printing the warn log which says context db failed to build. This is helpful to know once in the event where an indexer somehow gets through publishing with context.db issues. We don't need it repeatedly playing. 

This can be removed later when we refactor Indexer to build the context object once.

coordinator/Cargo.lock
```diff
@@ -2,6 +2,16 @@
 # It is not intended for manual editing.
 version = 3
 
+[[package]]
```
darunrs, line 5: Yeah I don't know what this change is but it keeps showing up in VS Code for some reason. It seems like its related to the new tracing stuff. I think my lock file is being auto generated without prompt and it keeps bugging me about the change.

darunrs:
Tested provisioning locally on partially provisioned and new indexers. Observed logs and metadata table were provisioned and tracked.

PR #662, opened 2024-04-15
feat: Count S3 get requests made by `near-lake-framework`

### LINKED ISSUE #419: Reduce duplicate Lake requests across dedicated streams

morgsmccauley:
Depends on https://github.com/near/near-lake-framework-rs/pull/102

This PR exposes a new metrics which counts the number of Get requests made to S3 by `near-lake-framework`. I wanted to start tracking this metric _before_ I merge the change which reduces them, so I can measure the impact of that change. The easiest way to track these requests was to pass a custom `S3Client` to `near-lake-framework`, so we can hook in to the actual requests made.

The custom `S3Client` (`LakeS3Client`) is exactly the same as the default implementation in `near-lake-framework` itself, but with the added metric. This is essentially part 1 for #419, as the "reduction" in requests will build on this custom client, adding caching/de-duplication.

### DIFF
--- a/block-streamer/Cargo.toml
+++ b/block-streamer/Cargo.toml
@@ -7,8 +7,8 @@ edition = "2021"
 actix-web = "4.5.1"
 anyhow = "1.0.57"
 async-trait = "0.1.74"
-aws-config = { version = "1.0.0", features = ["behavior-version-latest"]}
-aws-sdk-s3 = "0.39.1"
+aws-config = { version = "1.1.3", features = ["behavior-version-latest"] }
+aws-sdk-s3 = "1.13.0"
 borsh = "0.10.2"
 chrono = "0.4.25"
 futures = "0.3.5"
@@ -30,7 +30,7 @@ wildmatch = "2.1.1"
 
 registry-types = { path = "../registry/types", features = ["near-primitives"] }
 
-near-lake-framework = "=0.7.4"
+near-lake-framework = { git = "https://github.com/near/near-lake-framework-rs", branch = "feat/custom-s3-client" }
 
 [build-dependencies]
 tonic-build = "0.10"
--- a/block-streamer/Cargo.toml
+++ b/block-streamer/Cargo.toml
@@ -30,7 +30,7 @@ wildmatch = "2.1.1"
 
 registry-types = { path = "../registry/types", features = ["near-primitives"] }
 
-near-lake-framework = { git = "https://github.com/near/near-lake-framework-rs", branch = "feat/custom-s3-client" }
+near-lake-framework = "0.7.8"
 
 [build-dependencies]
 tonic-build = "0.10"
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -116,6 +116,7 @@ impl BlockStream {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
 #[tracing::instrument(
     skip_all,
     fields(
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -47,7 +47,7 @@ impl BlockStream {
         start_block_height: near_indexer_primitives::types::BlockHeight,
         redis_client: std::sync::Arc<crate::redis::RedisClient>,
         delta_lake_client: std::sync::Arc<crate::delta_lake_client::DeltaLakeClient>,
-        lake_s3_config: aws_sdk_s3::Config,
+        lake_s3_client: crate::lake_s3_client::LakeS3Client,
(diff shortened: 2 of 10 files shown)


block-streamer/Cargo.toml
```diff
@@ -30,7 +30,7 @@ wildmatch = "2.1.1"
 
 registry-types = { path = "../registry/types", features = ["near-primitives"] }
 
-near-lake-framework = "=0.7.4"
+near-lake-framework = { git = "https://github.com/near/near-lake-framework-rs", branch = "feat/custom-s3-client" }
```
morgsmccauley, line 33: Will update to actual Cargo version once this PR is merged

block-streamer/src/block_stream.rs
```diff
@@ -313,10 +314,31 @@ async fn process_near_lake_blocks(
 mod tests {
     use super::*;
 
+    use std::sync::Arc;
+
     use mockall::predicate;
+    use near_lake_framework::s3_client::GetObjectBytesError;
 
+    // FIX: near lake framework now infinitely retires - we need a way to stop it to allow the test
```
morgsmccauley, line 322: Will try to address this in a future PR, will probably need to add a config option to Near Lake which prevents infinite retries.

block-streamer/src/lake_s3_client.rs
```diff
@@ -0,0 +1,108 @@
...
+                .filter_map(|prefix_string| prefix_string.split('/').next().map(String::from))
+                .collect(),
+        };
+
+        Ok(prefixes)
+    }
+}
+
+#[cfg(test)]
+mockall::mock! {
```
morgsmccauley, line 85: Unable to use usual `#[automock]` due to the multiple `impl` blocks and trait implementation, so have to use this custom `mock!` block. The end result is essentially the same.

block-streamer/src/rules/matcher.rs
```diff
@@ -119,9 +119,10 @@ fn match_account(
             .split(',')
             .any(|sub_account_id| match_account(sub_account_id.trim(), outcome_with_receipt)),
         _ => {
-            wildmatch::WildMatch::new(account_id).matches(&outcome_with_receipt.receipt.receiver_id)
+            wildmatch::WildMatch::new(account_id)
+                .matches(outcome_with_receipt.receipt.receiver_id.as_str())
                 || wildmatch::WildMatch::new(account_id)
-                    .matches(&outcome_with_receipt.receipt.predecessor_id)
+                    .matches(outcome_with_receipt.receipt.predecessor_id.as_str())
```
morgsmccauley, line 125: Latest `near-lake-framework` changed these to `AccountId` from `String`

block-streamer/Cargo.toml
morgsmccauley: Done

darunrs (approved):
Sweet work! Excited to see the next PR with the caching!

block-streamer/src/block_stream.rs
```diff
@@ -116,6 +116,7 @@ impl BlockStream {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
morgsmccauley: Yeah - this suppresses the error which I was getting sick of. `BlockStreamContext` is a good idea.

block-streamer/src/lake_s3_client.rs
```diff
@@ -0,0 +1,108 @@
...
+            .request_payer(aws_sdk_s3::types::RequestPayer::Requester)
+            .send()
+            .await?;
+
+        let bytes = object.body.collect().await?.into_bytes().to_vec();
+
+        Ok(bytes)
+    }
+
+    async fn list_common_prefixes(
```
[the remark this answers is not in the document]
morgsmccauley: This is part of the `S3Client` trait exposed by `near_lake_framework`. Under the hood it's used to list block heights, and was originally added to mock those heights returned. We don't do anything with it now, but still need to supply it.

We may want to add caching here too, but it's a bit more complicated because the arguments will likely vary widely across all `near_lake_framework` instances.

block-streamer/src/lake_s3_client.rs
```diff
@@ -0,0 +1,108 @@
...
+        &self,
+        bucket: &str,
+        start_after_prefix: &str,
+    ) -> Result<Vec<String>, ListCommonPrefixesError> {
+        let response = self
+            .s3_client
+            .list_objects_v2()
+            .max_keys(1000)
```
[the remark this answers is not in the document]
morgsmccauley: This is the default from the API, essentially just a copy paste from `near_lake_framework` code.

PR #663, opened 2024-04-15
feat: Retry Tracking and Permissions for Tables in Hasura

### LINKED ISSUE #660: Retry Tracking and Permissions for New Logs/Metadata tables
Logs and Metadata tables get provisioned but tracking of the tables or adding permissions seems to fail from time to time. The tables need to be provisioned or else basic functionality (writing logs, setting status, etc.) will fail. 

We should retry tracking and permissions each loop until it succeeds. It's not necessary to block execution since writes to the table are done through Postgres and will still succeed. It's just that the table won't be queryable.

darunrs:
Logs Table and Metadata Table are necessary for important functions of QueryApi. Hasura often invisibly fails to track tables and add permissions. These operations need to be successful, so I added a check which verifies tracking and permissions are correct, and reattempts them if not. When successful, the result is cached. In a successful case, the expensive hasura calls (getTableNames and getTrackedTablesWithPermissions) are done at most twice. 

I also combined the conditional provisioning functions since they are verified to work already.

### DIFF
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -181,10 +181,7 @@ export default class HasuraClient {
     source: string,
     schemaName: string,
   ): Promise<any> {
-    const metadata = await this.executeMetadataRequest(
-      'export_metadata',
-      {},
-    );
+    const metadata = await this.exportMetadata();
     const tablesForSchema = metadata.sources
       .find((database: { name: string }) => database.name === source)
       .tables
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -10,12 +10,72 @@ interface SqlOptions {
   source?: string
 }
 
+export type HasuraPermission = 'select' | 'insert' | 'update' | 'delete';
+
+interface TableDefinition {
+  name: string
+  schema: string
+}
+interface HasuraRolePermission {
+  role: string
+  permission: {
+    check?: Record<string, any>
+    columns?: string[]
+    backend_only?: boolean
+    filter?: Record<string, any>
+    allow_aggregations?: boolean
+  }
+}
+type HasuraPermissions = HasuraRolePermission[];
+export interface HasuraTableMetadata {
+  table: TableDefinition
+  insert_permissions?: HasuraPermissions
+  select_permissions?: HasuraPermissions
+  update_permissions?: HasuraPermissions
+  delete_permissions?: HasuraPermissions
+}
+
 export interface HasuraDatabaseConnectionParameters {
   password: string
   database: string
   username: string
   host: string
   port: number
+}
+
+interface HasuraDatabaseUrl {
+  connection_parameters: HasuraDatabaseConnectionParameters
+}
+
+interface DefaultHasuraDatabaseUrl {
+  from_env: string
+}
+
+function isDefaultDatabaseUrl (
+  object: DefaultHasuraDatabaseUrl | HasuraDatabaseUrl
+): object is DefaultHasuraDatabaseUrl {
+  return 'from_env' in object;
+}
+
+export interface HasuraConfiguration {
+  connection_info: {
+    database_url: DefaultHasuraDatabaseUrl | HasuraDatabaseUrl
+    isolation_level?: string
+    pool_settings?: Record<string, any>
+    use_prepared_statements?: boolean
+  }
+}
+
+export interface HasuraSource {
+  name: string
+  kind: string
+  tables: HasuraTableMetadata[]
+  configuration: HasuraConfiguration
+}
+
+export interface HasuraMetadata {
+  version: number
+  sources: HasuraSource[]
 }
 
 type MetadataRequestArgs = Record<string, any>;
@@ -116,7 +176,7 @@ export default class HasuraClient {
     return await this.executeMetadataRequest('bulk', metadataRequests);
   }
 
-  async exportMetadata (): Promise<any> {
+  async exportMetadata (): Promise<HasuraMetadata> {
     const { metadata } = await this.executeMetadataRequest(
       'export_metadata',
       {},
@@ -126,10 +186,12 @@ export default class HasuraClient {
   }
 
   async getDbConnectionParameters (account: string): Promise<HasuraDatabaseConnectionParameters> {
-    const metadata = await this.exportMetadata();
-    const source = metadata.sources.find((source: { name: any, configuration: any }) => source.name === account);
+    const metadata: HasuraMetadata = await this.exportMetadata();
+    const source = metadata.sources.find((source: HasuraSource) => source.name === account);
     if (source === undefined) {
       throw new Error(`Could not find connection parameters for user ${account} on respective database.`);
+    } else if (isDefaultDatabaseUrl(source.configuration.connection_info.database_url)) {
+      throw new Error('Default connection parameters are not supported.');
     }
     return source.configuration.connection_info.database_url.connection_parameters;
   }
@@ -177,15 +239,13 @@ export default class HasuraClient {
       .map(({ name }: { name: string }) => name);
   }
 
-  async getTrackedTablesWithPermissions (
-    source: string,
+  async getTrackedTablePermissions (
+    databaseName: string,
     schemaName: string,
   ): Promise<any> {
-    const metadata = await this.exportMetadata();
-    const tablesForSchema = metadata.sources
-      .find((database: { name: string }) => database.name === source)
-      .tables
-      .filter((table: { table: { schema: string } }) => table.table.schema === schemaName);
+    const metadata: HasuraMetadata = await this.exportMetadata();
+    const hasuraSource = metadata.sources.find((source: HasuraSource) => source.name === databaseName);
+    const tablesForSchema = hasuraSource?.tables.filter((tableMetadata: HasuraTableMetadata) => tableMetadata.table.schema === schemaName);
 
     return tablesForSchema;
   }
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -172,6 +172,7 @@ export default class HasuraClient {
         source,
       }
     );
+    console.log(tablesInSource);
     return tablesInSource
       .filter(({ schema }: { schema: string }) => schema === schemaName)
       .map(({ name }: { name: string }) => name);
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -241,12 +241,12 @@ export default class HasuraClient {
   async getTrackedTablePermissions (
     databaseName: string,
     schemaName: string,
-  ): Promise<any> {
+  ): Promise<HasuraTableMetadata[]> {
     const metadata: HasuraMetadata = await this.exportMetadata();
     const hasuraSource = metadata.sources.find((source: HasuraSource) => source.name === databaseName);
     const tablesForSchema = hasuraSource?.tables.filter((tableMetadata: HasuraTableMetadata) => tableMetadata.table.schema === schemaName);
 
-    return tablesForSchema;
+    return tablesForSchema ?? [];
   }
 
   async trackTables (
--- a/runner/src/hasura-client/hasura-client.ts
+++ b/runner/src/hasura-client/hasura-client.ts
@@ -16,7 +16,7 @@ interface TableDefinition {
   name: string
   schema: string
 }
-interface HasuraRolePermission {
+export interface HasuraRolePermission {
   role: string
   permission: {
     check?: Record<string, any>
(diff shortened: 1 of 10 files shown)


darunrs:
Manual testing performed: 
1. Logs and metadata created and tracked, but metadata lacking permissions.
2. Neither created
3. metadata not tracked
4. logs not tracked
5. everything provisioned, tracked, and with permissions

I ran all with multiple blocks to confirm caching behavior works correctly

runner/src/provisioner/provisioner.ts
```diff
@@ -201,12 +220,12 @@ export default class Provisioner {
     return await wrapError(async () => await this.hasuraClient.trackTables(schemaName, tableNames, databaseName), 'Failed to track tables');
   }
 
-  async addPermissionsToTables (schemaName: string, databaseName: string, tableNames: string[], roleName: string, permissions: string[]): Promise<void> {
+  async addPermissionsToTables (indexerSchema: IndexerConfig, tableNames: string[], permissions: string[]): Promise<void> {
```
darunrs, line 223: I am thinking of refactoring Provisioner to use IndexerConfig as a class variable (passed in during construction) instead of having it passed around all over. Will do that in another PR though, instead of this one. But I did so for this function just for the sake of it.

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +243,95 @@ export default class Provisioner {
...
+    shouldHavePermissionsTables: string[],
+    tableMetadata: Map<string, HasuraTableMetadata>,
+    permissionsToCheck: Permission[]
+  ): string[] {
+    return shouldHavePermissionsTables.filter((tableName: string) => {
+      const tablePermissions = tableMetadata.get(tableName);
+      if (tablePermissions) {
+        return permissionsToCheck.some((permission: string) => {
+          const permisionAttribute = `${permission}_permissions` as keyof HasuraTableMetadata;
+          const usersWithPermission = tablePermissions[permisionAttribute] as (HasuraPermissions | undefined);
```
darunrs, line 327: Not terribly happy that I have to twist TypeScript's arm by asserting the types here so that my forEach loop below doesn't complain about types. Mainly the problem is that the `HasuraTableMetadata` interface has one attribute of type `TableDefinitions`, which causes the `.some()` below get mad since it can't convert `TableDefinitions` into `{ role: string }` format. 

Open to suggestions for how I could potentially clean that up.

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
 
   /**
-    * Provision logs table for existing Indexers which have already had all
+    * Provision logs and metadata table for existing Indexers which have already had all
     * other resources provisioned.
     *
     * */
-  async provisionLogsIfNeeded (indexerConfig: IndexerConfig): Promise<void> {
-    if (this.#hasLogsBeenProvisioned[indexerConfig.accountId]?.[indexerConfig.functionName]) {
+  async provisionLogsAndMetadataIfNeeded (indexerConfig: IndexerConfig): Promise<void> {
```
morgsmccauley, line 248: So I guess the intention here is that this method will run after the usual `provisionUserApi()`, therefore fixing any Hasura issues if any. But what about existing users who have already been provisioned? This will run standalone, and again, Hasura could have issues leaving us in a bad state.

I guess that is unlikely though since we are only tracking 2 tables: metadata and logs.

morgsmccauley: Ah, I understand now, this method will continue to run until both all tables are tracked and have roles.

It's a little confusing for this to do both provisioning of logs/metadata _and_ ensure Hasura is consistent. Maybe we can create a new method to do the latter: `ensureConsistentHasuraState()`?

I think that's clearer, but also, when we remove this function, that method can remain.

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
+        );
+
+        if (needsTrackingTables.length === 0 && needsPermissionsTables.length === 0) {
+          provisioningComplete = true;
+        } else {
+          if (needsTrackingTables.length > 0) {
+            await this.trackTables(indexerConfig.schemaName(), needsTrackingTables, indexerConfig.databaseName());
+          }
+          if (needsPermissionsTables.length > 0) {
+            await this.addPermissionsToTables(indexerConfig, needsPermissionsTables, permissionsToAdd);
```
morgsmccauley, line 286: Will this fail if a table has some of the permissions, i.e. just `select`/`insert`, and not the others? This will attempt to set them again?

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
+    permissionsToCheck: Permission[]
+  ): string[] {
+    return shouldHavePermissionsTables.filter((tableName: string) => {
+      const tablePermissions = tableMetadata.get(tableName);
+      if (tablePermissions) {
+        return permissionsToCheck.some((permission: string) => {
+          const permisionAttribute = `${permission}_permissions` as keyof HasuraTableMetadata;
+          const usersWithPermission = tablePermissions[permisionAttribute] as (HasuraPermissions | undefined);
+          // Returns true if the table does not have the permission or the user doesn't have the permission
+          const userIsLackingPermission = !usersWithPermission?.some((role: { role: string }) => role.role === userName);
```
morgsmccauley, line 327: I think this would make more sense as `role` rather than `user`?

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
+          return userIsLackingPermission;
+        });
+      }
+      return true;
+    });
```
morgsmccauley, line 332: It took me a while to grok this code, so I had a play around with it. This structure makes a lot more sense to me, especially the naming. Can we iterate here please? Feel free to adapt the below to what makes sense to you.

```ts
const trackedTablePermissions = await this.getTrackedTablePermissions(indexerConfig);
const untrackedTables = allTableNames.filter((tableName: string) => !trackedTablePermissions.has(tableName));
const tablesWithoutRole = this.getTablesWithoutRole(
  indexerConfig.hasuraRoleName(),
  allTableNames,
  trackedTablePermissions,
  permissionsToAdd
);

if (untrackedTables.length === 0 && tablesWithoutRole.length === 0) {
  provisioningComplete = true;
  return;
}

if (untrackedTables.length > 0) {
  await this.trackTables(indexerConfig.schemaName(), untrackedTables, indexerConfig.databaseName());
}

if (tablesWithoutRole.length > 0) {
  await this.addPermissionsToTables(indexerConfig, tablesWithoutRole, permissionsToAdd);
}
```

```ts
private getTablesWithoutRole (
    roleName: string,
    allTableNames: string[],
    tableMetadata: Map<string, HasuraTableMetadata>,
    permissionsToCheck: Permission[]
  ): string[] {
    return allTableNames.filter((tableName: string) => {
      const tablePermissionMetadata = tableMetadata.get(tableName);
      if (!tablePermissionMetadata) {
        return false;
      }

      return permissionsToCheck.some((permission: string) => {
        const permisionAttribute = `${permission}_permissions` as keyof Omit<HasuraTableMetadata, 'table'>;
        const tablePermissions = tablePermissionMetadata[permisionAttribute];
        // Returns true if the table does not have the permission or the user doesn't have the permission
        const roleExistsForTable = !tablePermissions?.some(({ role }) => role === roleName);
        return roleExistsForTable;
      });
    });
  }
```

morgsmccauley (changes requested):

runner/src/provisioner/provisioner.ts
darunrs: Hmm that's a good point. I'll do that!

runner/src/hasura-client/hasura-client.ts
```diff
@@ -177,6 +177,19 @@ export default class HasuraClient {
       .map(({ name }: { name: string }) => name);
   }
 
+  async getTrackedTablesWithPermissions (
```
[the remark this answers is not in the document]
darunrs: Renamed!

runner/src/hasura-client/hasura-client.ts
```diff
@@ -177,6 +177,19 @@ export default class HasuraClient {
...
+    source: string,
+    schemaName: string,
+  ): Promise<any> {
```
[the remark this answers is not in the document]
darunrs: Added strong types to Hasura. It should fully cover the returned Metadata information.

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
-          await this.trackTables(indexerConfig.schemaName(), [metadataTable], indexerConfig.databaseName());
-          await this.addPermissionsToTables(indexerConfig.schemaName(), indexerConfig.databaseName(), [metadataTable], indexerConfig.userName(), ['select', 'insert', 'update', 'delete']);
-        }
-      },
-      'Failed standalone metadata provisioning'
-    );
+    return trackedTablePermissions;
+  }
+
+  private getTablesMissingTracking (shouldBeTrackedTables: string[], tableMetadata: Map<string, any>): string[] {
```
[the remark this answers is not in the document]
darunrs: Renamed!

runner/src/provisioner/provisioner.ts
```diff
@@ -224,59 +241,95 @@ export default class Provisioner {
...
+    return shouldBeTrackedTables.filter((tableName: string) => !tableMetadata.has(tableName));
+  }
 
-    this.#hasMetadataBeenProvisioned[indexerConfig.accountId] ??= {};
-    this.#hasMetadataBeenProvisioned[indexerConfig.accountId][indexerConfig.functionName] = true;
+  private getTablesMissingPermissions (
```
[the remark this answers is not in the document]
darunrs: Renamed!

runner/src/provisioner/provisioner.ts
```diff
@@ -36,6 +36,24 @@ interface Config {
...
 
+type TableName = string;
+type TrackedTablePermissions = Map<TableName, HasuraTableMetadata>;
+
+interface TableDefinition {
+  name: string
+  columns: string
+}
+
+type Permission = 'select' | 'insert' | 'update' | 'delete';
```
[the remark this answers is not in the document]
darunrs: Moved and renamed to HasuraPermissions.

runner/src/provisioner/provisioner.ts
darunrs: Went ahead with Omit for now. Thanks for the suggestion! I believe I fixed the spelling mistakes too haha.

runner/src/provisioner/provisioner.ts
darunrs: Yeah I suppose that's true. I'll rename it.

runner/src/provisioner/provisioner.ts
darunrs: I did not cover partial failures of adding permissions. The current code just expects it to be a all or none. I did configure the code to accept different lists of permissions (even though we currently set all of them). I felt like that was something we can potentially cover in the future rather than now.

runner/src/provisioner/provisioner.ts
darunrs: Refactored the code to be easier to understand. Thanks for the call out!

runner/src/provisioner/provisioner.test.ts
```diff
@@ -316,3 +342,19 @@ describe('Provisioner', () => {
     });
   });
 });
+
+function generateTableConfig (schemaName: string, tableName: string, role: string, permissionsToAdd: HasuraPermission[]): HasuraTableMetadata {
```
darunrs, line 346: How weird would it be to export a function from a test class? Both of these benefit from a function which generates metadata .

runner/src/provisioner/provisioner.ts
```diff
@@ -234,60 +247,113 @@ export default class Provisioner {
...
+      }
+
+      return this.roleLacksPermissionsForTable(roleName, tablePermissionsMetadata, permissionsToCheck);
+    });
+  }
+
+  private roleLacksPermissionsForTable (roleName: string, tablePermissionsMetadata: HasuraTableMetadata, permissionsToCheck: HasuraPermission[]): boolean {
+    return permissionsToCheck.some((permission: string) => {
+      const permissionAttribute = `${permission}_permissions` as keyof Omit<HasuraTableMetadata, 'table'>;
+      return this.roleLacksPermission(roleName, tablePermissionsMetadata[permissionAttribute]);
```
morgsmccauley, line 351: Isn't this really `tablePermissionsLackRole`?

runner/src/provisioner/provisioner.test.ts
morgsmccauley: Kinda weird, personally I'd prefer it went in a `test-utils` file or something. I think the duplication is fine though 🤔

runner/src/indexer/indexer.ts
```diff
@@ -99,8 +99,8 @@ export default class Indexer {
           const provisionSuccessLogEntry = LogEntry.systemInfo('Provisioning endpoint: successful', blockHeight);
           logEntries.push(provisionSuccessLogEntry);
         }
-        await this.deps.provisioner.provisionLogsIfNeeded(this.indexerConfig);
-        await this.deps.provisioner.provisionMetadataIfNeeded(this.indexerConfig);
+        await this.deps.provisioner.provisionLogsAndMetadataIfNeeded(this.indexerConfig);
+        await this.deps.provisioner.ensureConsistentHasuraState(this.indexerConfig);
```
morgsmccauley, line 103: Eventually I think we should put this within `provisionUserApi()` and have it retry, but is probably fine living here for now

morgsmccauley (approved):

runner/src/provisioner/provisioner.test.ts
darunrs: Fair enough. Maybe if we end up with more I'll make a shared utils file.

runner/src/provisioner/provisioner.ts
darunrs: True. That's a better name.

PR #664, opened 2024-04-15
Point logs graphql query to the new logs schema

Kevin101Zhang:
Pointing the fetching of logs away from using hasura but instead to be using the new logs table from pg.


Kevin101Zhang:
Closes In favor of smaller PRs

PR #665, opened 2024-04-15
fix: Reduce requests made to Near Lake S3

### LINKED ISSUE #419: Reduce duplicate Lake requests across dedicated streams

morgsmccauley:
Each `BlockStream` uses its own dedicated `near-lake-framework` instance, and hence manages its own connection with S3. This leads to many duplicate S3 requests, particularly across the large majority of Indexers which follow the network tip, which request the same block data at the same time. 

This PR introduces a shared S3 client to be used across all `near-lake-framework` instances. `SharedLakeS3Client` ensures that duplicate requests made within a short time-frame, including those made in parallel, result in only a single request to S3.

## Cache Strategy

This implementation will mostly impact `BlockStream`s following the network tip, i.e. `From Latest`. These streams will wait for new data in Near Lake S3, and request it as soon as it is available, at the same time. Therefore, it would be enough to cache the result alone, by the time we actually prime the cache, all other requests would have missed it and fired a request of their own. Locking while the request is in-flight also is not feasible, as this would force _every_ request to execute in sequence.

Instead of caching the result of the request, we cache its computation. The first request initiates the request and stores its `Future`, then all subsequent requests retrieve that `Future` from cache and `await` its result, ensuring only one underlying request at most.

## Performance Impact

My main concern with this implementation is the impact it will have on performance. Each request made must block to check the cache, introducing contention/delays. The lock is only held while checking the cache, and not while the request is being made, so my hope is that it does not impact too much. This may be something that needs to be iterated over time.

From local testing the impact seemed to be negligible, but that was with 5 Indexers, it may be worse with many. I've added a metric to measure lock wait time, to determine whether this contention is becoming a problem.

### DIFF
--- a/block-streamer/src/lake_s3_client.rs
+++ b/block-streamer/src/lake_s3_client.rs
@@ -64,6 +64,7 @@ impl near_lake_framework::s3_client::S3Client for SharedLakeS3ClientImpl {
 #[derive(Debug)]
 pub struct LakeS3Client {
     s3_client: crate::s3_client::S3Client,
+    // TODO use a more efficient cache
     futures_cache: RwLock<HashMap<String, SharedGetObjectBytesFuture>>,
 }
 
@@ -100,17 +101,17 @@ impl LakeS3Client {
 
     async fn get_object_bytes_cached(&self, bucket: &str, prefix: &str) -> GetObjectBytesResult {
         let existing_future = {
-            let read_lock = self.futures_cache.read().await;
+            let futures_cache = self.futures_cache.read().await;
 
-            read_lock.get(prefix).cloned()
+            futures_cache.get(prefix).cloned()
         };
 
         let get_object_bytes_future = if let Some(future) = existing_future {
             future
         } else {
-            let mut write_lock = self.futures_cache.write().await;
+            let mut futures_cache = self.futures_cache.write().await;
 
-            write_lock
+            futures_cache
                 .entry(prefix.to_string())
                 .or_insert_with(|| self.get_object_bytes(bucket, prefix).shared())
                 .clone()
@@ -119,9 +120,9 @@ impl LakeS3Client {
         let get_object_bytes_result = get_object_bytes_future.await;
 
         if get_object_bytes_result.is_err() {
-            let mut write_lock = self.futures_cache.write().await;
+            let mut futures_cache = self.futures_cache.write().await;
 
-            write_lock.remove(prefix);
+            futures_cache.remove(prefix);
         }
 
         get_object_bytes_result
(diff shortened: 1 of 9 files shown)


block-streamer/src/lake_s3_client.rs
```diff
@@ -1,69 +1,174 @@
...
+        }
+    }
+
+    pub fn from_conf(config: aws_sdk_s3::config::Config) -> Self {
+        let s3_client = crate::s3_client::S3Client::new(config);
+
+        Self::new(s3_client)
+    }
+
+    fn get_object_bytes_shared(&self, bucket: &str, prefix: &str) -> SharedGetObjectBytesFuture {
```
morgsmccauley, line 129: This is essentially an `async fn`, and can be `await`ed as such, but, this `Future` can be cloned. To achieve this, the values referenced must be cloned to ensure they live long enough.

block-streamer/src/lake_s3_client.rs
```diff
@@ -1,69 +1,174 @@
...
+            Ok(bytes)
+        }
+        .boxed()
+        .shared()
+    }
+
+    async fn get_object_bytes_cached(&self, bucket: &str, prefix: &str) -> GetObjectBytesResult {
+        let get_object_bytes_future = self
+            .futures_cache
+            .get_or_set_with(prefix.to_string(), || {
```
morgsmccauley, line 150: Get/set must be done in one operation at the cache level. Doing them in sequence leaves room for multiple cache writes and therefore multiple requests.

block-streamer/src/lake_s3_client.rs
```diff
@@ -83,26 +188,146 @@ impl near_lake_framework::s3_client::S3Client for LakeS3ClientImpl {
...
+    use near_lake_framework::s3_client::S3Client;
+
+    #[tokio::test]
+    async fn deduplicates_parallel_requests() {
+        let s3_get_call_count = Arc::new(AtomicUsize::new(0));
+
+        let call_count_clone = s3_get_call_count.clone();
+
+        let mut mock_s3_client = crate::s3_client::S3Client::default();
+        mock_s3_client.expect_clone().returning(move || {
```
morgsmccauley, line 237: We can't truely clone this mock instance, so therefore can't simply assert that handler was called `once()`. Each clone gets a completely new instance.

To work around this, I use an atomic counter to count the actual number of requests made.

block-streamer/src/lake_s3_client.rs
```diff
@@ -83,26 +188,146 @@ impl near_lake_framework::s3_client::S3Client for LakeS3ClientImpl {
...
+
+                Ok(GetObjectOutput::builder().build())
+            });
+
+            mock_s3_client
+        });
+
+        let shared_lake_s3_client = SharedLakeS3ClientImpl::new(LakeS3Client::new(mock_s3_client));
+
+        let barrier = Arc::new(Barrier::new(10));
```
morgsmccauley, line 252: Blocks execution until it has been `wait`ed the specified number of times (`10`), making the requests fire in parallel.

block-streamer/src/s3_client.rs
```diff
@@ -52,7 +51,7 @@ impl S3ClientImpl {
             .list_objects_v2()
             .delimiter("/")
             .bucket(bucket)
-            .prefix(prefix);
```
morgsmccauley, line 55: This limits the keys to only those which begin with `prefix`, so for listing block heights, only 1 value is ever returned.

This may have been intentional for `DeltaLakeClient`, still need to confirm this doesn't break anything there.

block-streamer/src/s3_client.rs
```diff
@@ -120,3 +119,42 @@ impl S3ClientImpl {
         Ok(results)
     }
 }
+
+#[cfg(test)]
+mockall::mock! {
```
morgsmccauley, line 124: Changed to "manual" mock implementation so that `.clone()` can also be mocked. The default mock doesn't implement `Clone`.

block-streamer/src/lake_s3_client.rs
```diff
@@ -1,69 +1,174 @@
...
+use futures::{Future, FutureExt};
 use near_lake_framework::s3_client::{GetObjectBytesError, ListCommonPrefixesError};
+use tokio::sync::Mutex;
 
 use crate::metrics;
 
-#[cfg(not(test))]
-pub use LakeS3ClientImpl as LakeS3Client;
+/// Number of files added to Near Lake S3 per hour
+const CACHE_SIZE: usize = 18_000;
```
morgsmccauley, line 16: With 1 block produced per second - 60 seconds x 60 minutes x 5 files per block.

So roughly caching 1 hours worth of blocks?

darunrs: I'm trying to think how backfills would impact what constitutes the cache size. Backfill block requests might force out some of the queue size to be usable for catchups, but I think a lot would need to occur simultaneously for that to happen. I think this should work in the main scenarios I would expect us to gain the most benefits from:

1. When Block Streamer starts up after a pause.
2. When many streams are up to date.

block-streamer/src/lake_s3_client.rs
darunrs: Yeah we covered this in the meeting. We can probably try bumping the parallel threads/requests to something like 50 to more consistently test this behavior.

darunrs (approved):
Super cool change!

block-streamer/src/s3_client.rs
morgsmccauley: This was actually a breaking change - I reverted so existing functionality uses the original implementation, and this PR uses a new method.

block-streamer/src/lake_s3_client.rs
morgsmccauley: Backfills do use Delta Lake rather than Near Lake, so won't actually use this cache. But after that, you're right, they will start to flood it.

We can definitely adjust this value, or whole caching strategy, as we go :)

PR #669, opened 2024-04-16
fix: fix cron in provisioning

### LINKED ISSUE #668: fix cron statement in provisioning

Kevin101Zhang:
CRON statement functions was attempting to access a non-existent scoped property. Added syntax for dynamic sql generation to properly traverse.

Tested by setting cron fn_create_partition to trigger every 30 seconds. 
Previously we would not see the Non Trackable functions and we would get the original error message below. 
Now we are able to view the 2 Non Trackable functions and the row succeeds.

Screenshot 2024-04-16 at 7 39 10 PM

Screenshot 2024-04-16 at 8 02 24 PM

Original error
`ERROR: function fn_delete_partition(unknown, date, unknown, unknown) does not exist LINE 1: SELECT fn_delete_partition('kevin33_near_component_01.__logs... ^ HINT: No function matches the given name and argument types. You might need to add explicit type casts.`

### DIFF
--- a/runner/src/provisioner/provisioner.test.ts
+++ b/runner/src/provisioner/provisioner.test.ts
@@ -112,8 +112,8 @@ describe('Provisioner', () => {
         ['GRANT EXECUTE ON FUNCTION cron.schedule_in_database TO morgs_near;'],
       ]);
       expect(userPgClientQuery.mock.calls).toEqual([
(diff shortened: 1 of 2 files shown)


darunrs (approved):
Unit test error but otherwise good to go!

morgsmccauley (approved):

PR #674, opened 2024-04-18
fix: Pass startblock regardless of DDL Validation Error

### LINKED ISSUE #673: start_block is missing from certain transactions
Image

Image

Kevin101Zhang:
Passes startblock regardless if DDL Validation Error or not.

The culprit seems to be when we bypass the ddl Validation Error.
` There was an error while generating types for your SQL schema. Please ensure your schema is valid SQL DDL.`

If the SQL schema is valid we generate the start_block. If its not, we use a modal with obj and did not pass in the startblock k/v pair. The modal never had the start_block param.

### DIFF
--- a/frontend/src/components/Editor/Editor.jsx
+++ b/frontend/src/components/Editor/Editor.jsx
@@ -235,6 +235,13 @@ const Editor = ({ actionButtonText }) => {
     )[1];
     indexerName = indexerName.replaceAll(" ", "_");
 
+    const startBlock =
+    indexerConfig.startBlock === "startBlockHeight"
+      ? { HEIGHT: indexerConfig.height }
+      : indexerConfig.startBlock === "startBlockLatest"
+      ? "LATEST"
+      : "CONTINUE";  
+
     if (schemaValidationError?.type === FORMATTING_ERROR_TYPE) {
       setError(SCHEMA_FORMATTING_ERROR_MESSAGE);
       return;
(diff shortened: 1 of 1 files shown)


frontend/src/components/Editor/Editor.jsx
```diff
@@ -243,23 +250,12 @@ const Editor = ({ actionButtonText }) => {
...
-      startBlock = {
-        HEIGHT: indexerConfig.height
-      };
-    } else if (indexerConfig.startBlock === "startBlockLatest") {
-      startBlock = "LATEST";
-    } else {
-      startBlock = "CONTINUE"
-    }
-
     request("register-function", {
```
darunrs, line 259: I see we do request on the register function below. Is this what calls register? In the above if statement, we return if there is an error. How does showModal fit into continuing publishing and why does it not reach return, and this not reach request, which I presume calls register. I'm generally confused by the layout of execution that can happen here.

Kevin101Zhang: yeah thats a fair question - the request is nothing more than a string param thats passed in that we use in the `QueryAPI.Editor.jsx` file where it goes through request handler. Here are the relevant code snippets.

```

/**
 * Request Handlers here
 */
const requestHandler = (request, response) => {
  switch (request.type) {
    case "register-function":
      registerFunctionHandler(request, response);
      break;
    case "delete-indexer":
      deleteIndexer(request, response);
      break;
    case "default":
      console.log("default case");
  }
};
```
```

const registerFunctionHandler = (request, response) => {
  const gas = 200000000000000;
  const { indexerName, code, schema, startBlock, contractFilter } =
    request.payload;

  const jsonFilter = `{"indexer_rule_kind":"Action","matching_rule":{"rule":"ACTION_ANY","affected_account_id":"${contractFilter || "social.near"}","status":"SUCCESS"}}`

  Near.call(
    `${REPL_REGISTRY_CONTRACT_ID}`,
    "register",
    {
      function_name: indexerName,
      code,
      schema,
      start_block: startBlock,
      rule: {
        kind: "ACTION_ANY",
        affected_account_id: contractFilter,
        status: "SUCCESS"
      } 
    },
    gas
  );
};
```

darunrs: Ok gotcha. Yeah it's probably worth investigating to be sure this workflow works correctly in the future too.

darunrs:
Pointed local to use prod contract. Verified this fixes the issue.

darunrs (approved):

PR #675, opened 2024-04-18
fix: Continue Replacement of Logs/Metadata tables after untracking failures

darunrs:
Errors in provisioning should be logged to the machine as they can potentially be overwritten by errors in the finally block of the parent try catch. 

We ideally want to move the provisioning out to its own try catch but this is a simple fix for the time being. 

In addition the PRs to replace the logs/metadata tables failed due to untracking being partially successful. This PR allows untracking errors.

### DIFF
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -48,7 +48,7 @@ const defaultConfig: Config = {
 
 export default class Indexer {
   DEFAULT_HASURA_ROLE: string;
-  IS_FIRST_EXECUTION: boolean = false;
+  IS_FIRST_EXECUTION: boolean = true;
   tracer = trace.getTracer('queryapi-runner-indexer');
 
   private readonly deps: Dependencies;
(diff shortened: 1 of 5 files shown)


runner/src/server/runner-service.ts
```diff
@@ -34,7 +34,7 @@ function getRunnerService (executors: Map<string, StreamHandler>, StreamHandlerT
         return;
       }
 
-      console.log('Starting executor: ', indexerConfig);
+      console.log('Starting executor: ', JSON.stringify(indexerConfig));
```
pkudinov, line 37: We use multiple different names for the same thing. "Executor" seems to be only used in this code – perhaps worth replacing it with "StreamHandler" in this PR?

darunrs: We refer to instances in Runner as executor in Coordinator, so I retained that naming convention here too. The API is called StartExecutor.

runner/src/indexer/indexer.ts
```diff
@@ -158,7 +161,13 @@ export default class Indexer {
       await this.setStatus(blockHeight, IndexerStatus.FAILING);
       throw e;
     } finally {
-      await Promise.all([...simultaneousPromises, (this.deps.indexerMeta as IndexerMeta).writeLogs(logEntries)]);
+      this.IS_FIRST_EXECUTION = false;
+      try {
```
darunrs, line 165: It is possible for errors here to overwrite the caught error, which does not help with fixing problems. The better long term solution is to have a try and log approach to provisioning IndexerMeta. For now, we can just print the log so that the original error continues through.

darunrs:
Merging this as it's necessary to resolve some of the problems in dev currently for replacing logs and metadata tables.

PR #676, opened 2024-04-19
Use new Metadata Table for status and block height

darunrs:
Frontend now queries the new metadata table instead of the old one. 
image

Indexers which have not successfully run since the introduction of the new logs table will not have a last processed block height since they never processed a block successfully. So, I opted to set it as N/A and leave a tooltip on hover which says why its N/A. 
image


darunrs:
Query needs to be renamed after creating new metadata table. Existing metadata and logs tables are prepended with double underscores, which are reserved by graphQL. The tables need to be renamed.

frontend/src/components/Logs/Status.jsx
```diff
@@ -8,81 +8,87 @@ import {
...
-const Status = ({ functionName, latestHeight }) => {
-  const GET_STATUS = gql`
-    query GetState($_functionName: String!) {
-      indexer_state(where: { function_name: { _eq: $_functionName } }) {
-        status
-        function_name
-        current_block_height
-        current_historical_block_height
+const Status = ({ accountId, functionName, latestHeight }) => {
+  const hasuraRole = accountId.replace(/[^a-zA-Z0-9]/g, '_'); // TODO: Support accounts that start with a number
```
morgsmccauley, line 12: Can we support this TODO now? Implicit accounts are likely to start with numbers

morgsmccauley (approved):
Just one comment

frontend/README.md
```diff
@@ -12,9 +12,9 @@ First, download the bos-loader cli by following this guide [here](https://docs.n
 From the root of QueryAPI Frontend repo, run the following command
 
 ```bash
-yarn serve:widgets:local // for running local enviornment
```
darunrs, line 15: Since we removed the yarn file from Runner, we should do the same for frontend too. I checked and we do not use any yarn commands in the deployment of frontend in Terraform. Deleting yarn.lock for the same reason.

runner/package.json
```diff
@@ -54,11 +54,12 @@
     "@opentelemetry/api": "^1.8.0",
     "@opentelemetry/exporter-zipkin": "^1.22.0",
     "@opentelemetry/resources": "^1.22.0",
-    "@opentelemetry/sdk-node": "^0.49.1",
+    "@opentelemetry/sdk-node": "^0.50.0",
     "@opentelemetry/sdk-trace-base": "^1.22.0",
     "@opentelemetry/sdk-trace-node": "^1.22.0",
     "@opentelemetry/semantic-conventions": "^1.22.0",
     "express": "^4.18.2",
+    "graphql": "^16.8.1",
```
darunrs, line 62: Apparently we were using this package but it wasn't automatically added through install.

.gitignore
```diff
@@ -4,3 +4,5 @@ redis/
 *.log
 /indexer/blocks/
 node_modules/
+.vscode/
+*/yarn.lock
```
darunrs, line 8: There are mainly for convenience. .vscode folder is necessary for the IDE to work right, and I'm ignoring yarn.lock files to prevent publishing of them again in the future.

frontend/src/components/Logs/Status.jsx
```diff
@@ -8,81 +8,87 @@ import {
...
-                <strong>{item.current_historical_block_height}</strong>
-              </ListGroup.Item>
-              <ListGroup.Item>
                 <OverlayTrigger
                   placement="bottom"
-                  overlay={<Tooltip>{item.status === "RUNNING" ? "Indexer is operating normally" : "Indexer stopped due to errors. Check Logs for more details."}</Tooltip>}
+                  overlay={<Tooltip> {
+                    attributeMap.get("LAST_PROCESSED_BLOCK_HEIGHT")
+                    ? `Current Block Height of Near is ${latestHeight}. Your indexer has a gap of ${latestHeight - attributeMap.get("LAST_PROCESSED_BLOCK_HEIGHT")} Blocks`
+                    : 'Indexer needs to run successfully to update block height' } </Tooltip>}
```
darunrs, line 62: It is possible now for failing indexers to not have a last processed block as they haven't successfully processed a block since the new table was rolled out. 

We only set the height after a successful run. I have a custom message for this case.

PR #684, opened 2024-04-22
feat: Expose log count metric from Runner

morgsmccauley:
Creates a new `winston` transport method to count logs by level, and exposes a new prometheus metric to record this value.

Additionally, metrics recorded on the main thread were not captured. This was not an issue as the majority of metrics were done within the worker. But since we also log on main thread, this PR updates metrics aggregation to expose main thread metrics.

### DIFF
--- a/runner/src/logger.ts
+++ b/runner/src/logger.ts
@@ -1,7 +1,18 @@
 import winston from 'winston';
 import { LoggingWinston } from '@google-cloud/logging-winston';
+import Transport from 'winston-transport';
+
+import { METRICS } from './metrics';
 
 const { format, transports } = winston;
+
+class LogCounter extends Transport {
+  log (info: { level: string }, callback: () => void): void {
+    METRICS.LOGS_COUNT.labels({ level: info.level }).inc();
+
+    callback();
+  }
+}
 
 const logger = winston.createLogger({
   level: 'info',
@@ -9,6 +20,7 @@ const logger = winston.createLogger({
     format.timestamp(),
     format.errors({ stack: true }),
   ),
+  transports: [new LogCounter()],
 });
 
 if (process.env.GCP_LOGGING_ENABLED) {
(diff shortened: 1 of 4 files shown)


runner/src/logger.ts
```diff
@@ -1,14 +1,26 @@
 import winston from 'winston';
 import { LoggingWinston } from '@google-cloud/logging-winston';
+import Transport from 'winston-transport';
+
+import { METRICS } from './metrics';
 
 const { format, transports } = winston;
 
+class LogCounter extends Transport {
+  log (info: { level: string }, callback: () => void): void {
```
darunrs, line 10: Is this going to count only .log calls? Or is this applied to all log calls (.warn, .error, etc.)? Another thing, it would be nice if we could similarly differentiate between error logs too for metrics. For example, it would be nice to distinguish between a single "Encountering error processing stream" error log and many "Error writing..." error calls. The former indicating the worker shut down whereas the latter could be any of these three: Hasura is down, our system queries are malformed, or an indexer is trying to run some faulty query.

darunrs (approved):

runner/src/metrics.ts
```diff
@@ -87,7 +94,9 @@ export const startServer = async (): Promise<void> => {
   app.get('/metrics', async (_req, res) => {
     res.set('Content-Type', aggregatorRegistry.contentType);
 
-    const metrics = await AggregatorRegistry.aggregate(Array.from(workerMetrics.values())).metrics();
+    const mainThreadMetrics = await register.getMetricsAsJSON();
```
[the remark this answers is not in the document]
morgsmccauley: I outlined this in the description - mainly just the log counts :)

runner/src/logger.ts
morgsmccauley: We could definitely add that, but it may be expensive given all the different log messages. My motive behind this was to be able to quickly see _if_ we are erroring out from Grafana, alert from it, and then go in to GCP to see the actual issue/log messages.

PR #685, opened 2024-04-22
fix: Add back yarn frontend

darunrs:
My due diligence was insufficient. As it turns out, yarn.lock IS used by the frontend during development. Adding it back.


darunrs:
Frontend still runs locally.

PR #692, opened 2024-04-23
fix: Remove all hasura write calls to logs/state tables

### LINKED ISSUE #618: Remove writing logs through Hasura

darunrs:
Removing writes to old logs and state table now that it is confirmed that the new tables work as expected. We expect to also see a reduction in the number of DB connections used by QueryApi since fewer write calls are being made by Hasura. 

I've retained the whitelist so that we can still control provisioning of new tables.


runner/src/indexer/indexer.ts
```diff
@@ -205,21 +188,21 @@ export default class Indexer {
           setSpan.end();
         }
       },
-      debug: async (...log) => {
-        const debugLogEntry = LogEntry.systemDebug(log.join(' '), blockHeight);
-        return await this.writeLog(debugLogEntry, logEntries);
+      debug: (...log) => {
+        const debugLogEntry = LogEntry.systemDebug(log.join(' : '), blockHeight);
```
morgsmccauley, line 192:
```suggestion
        const debugLogEntry = LogEntry.userDebug(log.join(' : '), blockHeight);
```
Should we be using `user` here and below?

darunrs: Yep!

morgsmccauley (approved):

PR #702, opened 2024-04-24
fix: Stable hasura state fix

darunrs:
Not intended for actual merge. I will instead grab the commit hash and use it in prod runner. This is to side step having to force push commit history in the prod release but still have this fix in place.


darunrs:
No longer needed as migration to prod is complete.

PR #721, opened 2024-04-30
fix: Reduce lock contention in Lake Cache

### LINKED ISSUE #419: Reduce duplicate Lake requests across dedicated streams

morgsmccauley:
This PR addresses significant lock contention issues within the Lake Cache by replacing the `tokio::sync::Mutex` with `std::sync::Mutex`. The use of an asynchronous mutex exacerbated contention in our highly concurrent environment. Specifically, `tokio::sync::Mutex` yielding when the lock is unavailable, leads to all tasks requiring the lock to enter a repetitive cycle of acquisition attempts, failures, and subsequent yields. When the lock becomes available, all waiting tasks wake simultaneously, attempt to re-acquire the lock, and yield again if unsuccessful. This behaviour results in substantial scheduling overhead.

Switching to `std::sync::Mutex` alters this dynamic by blocking the thread attempting to acquire the lock, rather than merely yielding control. This change simplifies the lock acquisition process by eliminating the need to wake multiple tasks. Instead, the OS scheduler directly allocates the lock to one of the waiting threads when it becomes available. This approach reduces the complexity and overhead introduced by the task scheduler, enhancing performance and reliability under high contention.

### DIFF
--- a/block-streamer/src/lake_s3_client.rs
+++ b/block-streamer/src/lake_s3_client.rs
@@ -2,13 +2,13 @@
 
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::{Mutex, MutexGuard};
 
 use async_trait::async_trait;
 use cached::{Cached, SizedCache};
 use futures::future::Shared;
 use futures::{Future, FutureExt};
 use near_lake_framework::s3_client::{GetObjectBytesError, ListCommonPrefixesError};
-use tokio::sync::Mutex;
 
 use crate::metrics;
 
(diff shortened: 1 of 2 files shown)


block-streamer/src/lake_s3_client.rs
```diff
@@ -77,12 +77,19 @@ impl FuturesCache {
...
+    fn lock(&self) -> MutexGuard<'_, SizedCache<String, SharedGetObjectBytesFuture>> {
         let timer = metrics::LAKE_CACHE_LOCK_WAIT_SECONDS.start_timer();
 
-        let lock = self.cache.lock().await;
+        let lock = match self.cache.lock() {
+            Ok(lock) => lock,
+            Err(poisoned) => {
+                let lock = poisoned.into_inner();
+
+                tracing::error!("Lake Cache Mutex was poisoned, recovering...");
```
darunrs, line 88: Just curious, what sort of scenario would lead to a lock throwing an error like this?

block-streamer/src/lake_s3_client.rs
```diff
@@ -77,12 +77,19 @@ impl FuturesCache {
...
+
+                lock
```
darunrs, line 90: So we still return the lock? Is the intention to simply log the error but proceed? Why does the lock still work?

darunrs (approved):
Nice solution!

block-streamer/src/lake_s3_client.rs
morgsmccauley: > If another user of this mutex panicked while holding the mutex, then
this call will return an error once the mutex is acquired.

from the docs

block-streamer/src/lake_s3_client.rs
morgsmccauley: Yeah, another thread panicing while still holding the lock isn't really an issue. The cache wouldn't necessarily be in a bad state so we can still operate on it.

PR #724, opened 2024-05-02
feat: Store Forked Indexer when Publishing

### LINKED ISSUE #698: Track forks of indexers
Similar to how React components are tracked on forks, we need to introduce `forked_from` field when someone forks an indexer. Based on that we can build a Databricks query that would track the number of times an indexer was forked and whether the source indexers are being used.

darunrs:
Maintaining what Indexer was forked during development is useful for various usage statistics. I've updated the frontend to now store what indexer was forked from and store that information in the contract after publishing. We specifically store the indexer which was directly forked. 

With this information, we can construct a graph using the "forked_from" as edge information, and calculate various other statistics such as how many indexers used some other indexer as a base by doing path counts. This can be done by creating an indexer which indexes the QueryApi contract. 

There are some bugs in the workflow currently. Specifically, refreshing while in the forked page will cause the loss of whatever code was written. You need to re-fork the Indexer. While this benefits to ensuring the forked from field is correct, if we end up fixing this behavior, we need to then ensure forked from is stored persistently. 

As part of updating the contract, I've removed various old functions which are no longer necessary.

### DIFF
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -260,6 +260,8 @@ impl Contract {
         &mut self,
         account_id: Option<String>,
         function_name: String,
+        forked_from_account_id: Option<String>,
+        forked_from_function_name: Option<String>,
         code: String,
         schema: String,
         rule: Rule,
@@ -277,6 +279,20 @@ impl Contract {
                 env::signer_account_id()
             }
         };
+
+        let forked_from: match (forked_from_account_id, forked_from_function_name) {
+            (Some(account_id), Some(function_name)) => {
+                const account_id_parsed = account_id.parse::<AccountId>().unwrap_or_else(|_| {
+                    env::panic_str(&format!("Account ID {} is invalid", account_id));
+                });
+                IndexerIdentity {
+                    account_id_parsed,
+                    function_name
+                }
+            },
+            (None, None) => Ok(None),
+            _ => env::panic_str(&format!("Either the account {} or name {} of the forked indexer is missing", forked_from_account_id, forked_from_function_name),
+        }
 
         log!(
             "Registering function {} for account {}",
@@ -317,6 +333,7 @@ impl Contract {
                     start_block,
                     updated_at_block_height: Some(env::block_height()),
                     created_at_block_height: indexer.created_at_block_height,
+                    forked_from,
                 });
             }
             near_sdk::store::unordered_map::Entry::Vacant(entry) => {
@@ -327,6 +344,7 @@ impl Contract {
                     start_block,
                     updated_at_block_height: None,
                     created_at_block_height: env::block_height(),
+                    forked_from,
                 });
             }
         }
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -260,8 +260,7 @@ impl Contract {
         &mut self,
         account_id: Option<String>,
         function_name: String,
-        forked_from_account_id: Option<String>,
-        forked_from_function_name: Option<String>,
+        forked_from: Option<IndexerIdentity>,
         code: String,
         schema: String,
         rule: Rule,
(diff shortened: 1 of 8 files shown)


registry/contract/src/lib.rs
```diff
@@ -455,44 +361,16 @@ impl Contract {
         }
     }
 
-    pub fn list_indexer_functions(&self, account_id: Option<String>) -> OldAccountOrAllIndexers {
```
morgsmccauley, line 458: Can we just double check these aren't used anywhere in the codebase (probably just frontend) to ensure this isn't a breaking change.

registry/contract/src/lib.rs
```diff
@@ -281,7 +280,7 @@ impl Contract {
                 affected_account_id,
                 ..
             } => {
-                if affected_account_id == "*" {
+                if affected_account_id.contains("*") {
```
morgsmccauley, line 283: This would prevent `*.something.near` which isn't what we want? I think we still want `== "*"`?

registry/contract/src/lib.rs
```diff
@@ -259,12 +245,25 @@ impl Contract {
     pub fn register(
         &mut self,
         function_name: String,
+        forked_from: Option<IndexerIdentity>,
         code: String,
         schema: String,
         rule: Rule,
         start_block: StartBlock,
+        account_id: Option<String>,
```
morgsmccauley, line 253:
```suggestion
        account_id: Option<AccountId>,
```
If I recall correctly, inputs will be parsed to their defined types when the method is called. So setting `AccountId` makes the parsing below obsolete.

registry/contract/src/lib.rs
```diff
@@ -259,12 +245,25 @@ impl Contract {
...
     ) {
-        let account_id = env::signer_account_id();
+        let account_id = match account_id {
+            Some(account_id) => {
+                self.assert_roles(vec![Role::Owner]);
```
morgsmccauley, line 257: Maybe we should only make this assertion if `env::signer_account_id` is different from what's passed 🤔

registry/types/src/lib.rs
```diff
@@ -14,123 +14,6 @@ use serde::{Deserialize, Serialize};
 
 type FunctionName = String;
 
-#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
```
morgsmccauley, line 17: These types are shared across `block-streamer` and `coordinator`, can you try compile those just to ensure they still work. I'm pretty sure there will be no impact, just want to sanity check.

registry/contract/src/lib.rs
darunrs: I think we usually don't provide it at all. At least, the frontend omits it. But I can add a condition over the assert to only assert if it's different. Not too hard to do.

registry/contract/src/lib.rs
darunrs: I see. Yeah in Block Streamer, we actually have a hardcoded check where if the filter contains any of `["*", "*.near", "*.kaiching", "*.tg"]` then we skip lake filtering. But the contract only checks Owner role for *. Even though *.near is basically just *. 

I guess I can for now I can make them match. We'll just need to remember to evolve the contract if we make changes to the restrictions in the future.

registry/contract/src/lib.rs
darunrs: So we do not use it anywhere. But, Gabe's account status component does use it. However, his component is only interested in the 2D map structure. He picks the Map corresponding to the account, and then gets the keys from the inner map. So, that's the reason I still kept this API. I just modified it to use our new structures, and the unit tests interestingly all passed with minimal changes.

registry/types/src/lib.rs
darunrs: Thanks for the call out. I did a cargo build on both and they succeeded building. I don't believe I need to do any runtime checks right?

morgsmccauley: Yup, compile check is enough.

morgsmccauley (approved):

PR #725, opened 2024-05-08
fix: Catch failed Lake S3 request to prevent unhandled rejections

### LINKED ISSUE #715: TimeoutError: socket hang up in Runner
**Describe the bug**
Some indexers in DEV stop with "TimeoutError: socket hang up" error in Cloud Logs. Users see this in logs: `Encountered error processing stream: nearpavel.near/bitmap_v2:block_stream, terminating thread TimeoutError: socket hang up`

**To Reproduce**
It is hard to reproduce. Bitmap_v2 indexer keeps running for a few hours and then shuts down with this error.

**Screenshots**
image
image

morgsmccauley:
Rejected promises must be handled via `await` or `.catch()`, without either of these Node will exit due to the "unhandled rejection". Currently, Lake S3 requests are created in advance, and then handled later when the block is ready to be executed. Usually, this is not an issue, as the delay between promise creation/handling is not too large. But for slow indexers (i.e. `nearpavel_near/bitmap_v2`), this delay is long enough for Node to consider it "unhandled".

This PR attaches a rejection handler to block requests, so that failures can be handled gracefully. This does not affect the "pre-fetch" behaviour, these requests will still be executed ahead of time, but failed requests will be handled within our code. Explicit handling binds the error to our execution context providing a meaningful call-stack, as opposed to a lonesome error which seemingly comes from no-where.

To mitigate the problem itself - failed S3 requests, I have bumped `maxAttempts`, and changed the `retryMode` in the hopes that the transient error can be overcome.

### DIFF
--- a/runner/src/indexer/indexer.ts
+++ b/runner/src/indexer/indexer.ts
@@ -79,6 +79,8 @@ export default class Indexer {
   async execute (
     block: lakePrimitives.Block,
   ): Promise<string[]> {
+    this.logger.debug('Executing block', { blockHeight: block.blockHeight });
+
     const blockHeight: number = block.blockHeight;
 
     const lag = Date.now() - Math.floor(Number(block.header().timestampNanosec) / 1000000);
--- a/runner/src/lake-client/lake-client.ts
+++ b/runner/src/lake-client/lake-client.ts
@@ -6,7 +6,10 @@ import RedisClient from '../redis-client';
 export default class LakeClient {
   constructor (
     private readonly network: string = 'mainnet',
-    private readonly s3Client: S3Client = new S3Client(),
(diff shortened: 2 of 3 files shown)


darunrs (approved):
Nice! I'm definitely intrigued about this error though. If an indexer fails to fetch a block, it would stop the entire Indexer. Yet, we nearly never see this error. worker does not attempt retries outside of S3's own retry mechanism. I wonder why bitmap_v2 is specifically so prone to this error.

morgsmccauley:
@darunrs Yeah, I wondered the same. My best guess is that the long synchronous code causes the socket to be closed abruptly on either client/server side. I'll continue investigating.

PR #727, opened 2024-05-09
refactor: Introduce struct to manager persistent indexer state

### LINKED ISSUE #667: Ability to manually suspend indexers
A new secure API that implements indexer blacklist in the coordinator. We should be able to Create, Delete and List all blacklisted indexer. The API should be secure and not available externally.

@darunrs list down all APIs and how will they work.
```[tasklist]
### Tasks
- [ ] https://github.com/near/queryapi/issues/719
- [ ] https://github.com/near/queryapi/issues/720
```

morgsmccauley:
This PR centralises persistent Indexer State within the `IndexerStateManager` struct. Currently, we only persist the "stream version", but this will soon grow to "enabled/disabled", which will be implemented in my next PR. This is just a tidy up to make the next step a bit easier.

Indexer state will be stored as stringified JSON under `{account_id}/{function_name}:state`, currently this only includes when the block stream was last synced. I've included a migration step in to move from the old key/structure to the new.

### DIFF
--- a/coordinator/src/main.rs
+++ b/coordinator/src/main.rs
@@ -55,7 +55,9 @@ async fn main() -> anyhow::Result<()> {
     loop {
         let indexer_registry = registry.fetch().await?;
 
-        // migrate state
+        indexer_manager
+            .migrate_state_if_needed(&indexer_registry)
+            .await?;
 
         tokio::try_join!(
             synchronise_executors(&indexer_registry, &executors_handler),
@@ -70,7 +72,5 @@ async fn main() -> anyhow::Result<()> {
                 Ok(())
             }
         )?;
-
-        // flush state
     }
 }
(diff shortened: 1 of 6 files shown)


coordinator/src/main.rs
```diff
@@ -52,9 +55,18 @@ async fn main() -> anyhow::Result<()> {
     loop {
         let indexer_registry = registry.fetch().await?;
 
+        indexer_state_manager
+            .migrate_state_if_needed(&indexer_registry)
```
morgsmccauley, line 59: Migrating "stream versions" to new indexer state objects

coordinator/src/redis.rs
```diff
@@ -93,11 +93,61 @@ impl RedisClientImpl {
...
+        ) -> anyhow::Result<Option<u64>>;
+
+        pub async fn clear_block_stream(&self, indexer_config: &IndexerConfig) -> anyhow::Result<()>;
+
+        pub async fn set_migration_complete(&self) -> anyhow::Result<()>;
+
+        pub async fn is_migration_complete(&self) -> anyhow::Result<Option<bool>>;
+    }
+
+    impl Clone for RedisClientImpl {
```
morgsmccauley, line 150: Need to manually mock `RedisClient` so `Clone` is available on the mock object.

gabehamilton (approved):

PR #728, opened 2024-05-10
Introduce Logs Table v1

### LINKED ISSUE #224: Improve UX for viewing logs and status on the Indexer Status page
Description

- [x] Show the number of logs per each severity status and log type. This should be a clickable filter displayed on the left to the logs (similar to GCP Logs)
- [x] Improve table styling to look more like GCP logs (fit more logs on one page)
- [x] Refresh button should also refresh the status of the indexer
- [x] Status section should be made smaller by reducing unnecessary info and vertical space
- [x] Filter by block should still work
- [ ] Optional: add auto-refresh button that would fetch new logs every 5 seconds

### LINKED ISSUE #726: Make GridJS query from GraphQL

Kevin101Zhang:
Introduce an improved logs table with new fields and the ability to search and filter based of radio inputs.


frontend/src/components/Logs/LogsViewContainer/IndexerLogsContainer.tsx
```diff
@@ -0,0 +1,192 @@
...
+    server: any;
+    style: any;
+    sort: boolean;
+}
+
+interface InitialPayload {
+    currentUserAccountId: string;
+}
+
+const DEV_ENV: string = 'https://queryapi-hasura-graphql-mainnet-vcqilefdcq-ew.a.run.app/v1/graphql';
```
Kevin101Zhang, line 24: TODO: use env

Kevin101Zhang:
All the Logs Table logic begins by following the container file 
**frontend/src/components/Logs/LogsViewContainer/IndexerLogsContainer.tsx**

here we defined a series of configs 
```
getIndexerLogConfig
getSearchConfig
getPaginationConfig
getGridStyle
getLanguageConfig

**That is used by gridjs and rendered/reloaded out. with a basic useEffect**

const getGridConfig: any = () => {
        return {
            columns: [...],
            search: getSearchConfig(), **Any Interaction of the search bar**
            pagination: getPaginationConfig(), // static JSON
            style: getGridStyle(), //just style
            server: getIndexerLogsConfig(),  **Any interaction with filters**
            language: getLanguageConfig(), // just placeholder style 
            // sort: true,
        };
    };
 
From there on each file is broken down in small files and helper functions. 

```

Kevin101Zhang:
Requesting a review outside of the exporting DEV_ENV/PROD_ENV

frontend/src/components/Logs/GraphQL/GetSearchLogsQueryDefinition.ts
```diff
@@ -0,0 +1,46 @@
...
+    getBlockHeightAndMessageSearchQueryWithTimeRange,
+    getMessageSearchQueryWithSeverityAndTimeRange,
+    getBlockHeightAndMessageSearchQueryWithSeverityAndTimeRange,
+    getMessageSearchQueryWithLogTypeAndTimeRange,
+    getBlockHeightAndMessageSearchQueryWithLogTypeAndTimeRange,
+    getBlockHeightAndMessageSearchQueryWithSeverityAndLogTypeAndTimeRange,
+    getMessageSearchQueryWithSeverityAndLogTypeAndTimeRange,
+} from "./Queries";
+
+export const getSearchLogsQueryDefinition = (tableName: string, keyword: string, severity: string, logType: string, startTime: string): void => {
```
morgsmccauley, line 21: I really think we should use the single query that I suggested, this is way more complicated than it needs to be.

I would prefer to do it now rather than later, deferring it just creates more work. If we need to introduce a new variable _all_ of this needs to change, as opposed to just changing the query itself.

morgsmccauley (changes requested):

frontend/src/components/Logs/GraphQL/GetSearchLogsQueryDefinition.ts
Kevin101Zhang: I'm in favor of changing it to (default) parameterized queries for file organization and neatness. But to be honest I'm also in favor of creating a query builder class to build out the queries themselves and instantiate the class with parameters to build the query since its technically the cleanest code approach (which isn't a good idea due to readability and debugging).

In terms of debugging, the code currently isWET (Write Every Time). Its has been really easy to debug since the query names are static and I can see which variation of filters (in the UI) cause conflicts as each combination is a new query(with the exception of timestamp)

With the use of default variables there may be a con with debugging since the query name and title itself will now be dynamically generated and if one variation does not work I will be affecting multiple queries if I make any changes. (It essentially either all work or all does not). I guess this is pretty dependent on many queries we end up with really. 

If I'm understanding the link correctly We can the entire file down to a minimum of 2 queries.
One for searching and one for Indexer. I'm fine with either approach in the end.

darunrs: I'd personally reduce the number of queries. It doesn't necessarily have to be down 1 but definitely fewer. Regarding testing, I don't think it would be too hard to just templatize the query, and print it as is. You can just copy it from the console into Hasura to work with it. Going back and replacing it with templates might be a small hassle but we'd only really do this rarely during debugging or adding new variables. 

The way I see it, we have the following outcomes:

- A query doesn't work anymore
- We add a new variable to a query

Debugging an existing query is, as I mentioned, printing the failing query to the console. And since we don't use variables, it's a direct copy paste. I don't think re-templatizing it is hard and you can use any number of string diff checkers to aid in that. 

For that second case, when we add a new variable, we need to add new combinations for that new variable, which I think is a much bigger hassle. We also may end up expanding the bitmap you are using below. It's definitely a lot of additional dev overhead. And it's way harder as well for a new person to work with because they need to maintain all these queries in their head as additional context along with the bitmap case switch. 

I think the better solution is to wrap the query in a nice interface which takes in well defined variable names. And we ensure whenever we call the function, we properly name any variables we pass it. That way, all the context you need is right there. Internally, you can create some helper functions to do things like `setReturnType` or `aggregateOn` which take in variables and generates a string using them. You can outright omit null variables instead of adding a default type. Basically, you're doing a StringBuilder approach by stringing in helper methods to fill out some base template query. This might be a good alternate approach if adding default values doesn't seem to work. 

Regarding a query builder, I think it would be a nice optimization to have if we were making queries to a LOT of different schemas. But we only query one or two types of tables. I feel it isn't necessary at this stage.

frontend/src/components/Logs/GraphQL/GetSearchLogsQueryDefinition.ts
```diff
@@ -0,0 +1,46 @@
...
+    const isKeywordNumber = !isNaN(Number(keyword));
+    const ISOString: string = startTime ? calculateTimestamp(startTime) : startTime;
+
+    const conditions: { [key: string]: () => void } = {
```
darunrs, line 25: This is quite scary to look at to be honest. Debugging this would be frustrating.

frontend/src/components/Logs/LogsViewContainer/DateSelectorContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+    LAST_45_MINUTES,
+    LAST_1_HOUR,
+    LAST_3_HOURS,
+    LAST_6_HOURS,
+    LAST_12_HOURS,
+    LAST_1_DAY,
+    LAST_2_DAYS,
+    LAST_7_DAYS,
+    LAST_14_DAYS,
+    LAST_30_DAYS
```
darunrs, line 20: Rather than making these individual values, can you encapsulate them under some Map or Enum? I don't see why we would need individual values when they could easily all be under say DateOptions as DateOptions.LAST_7_DAYS. You can get values from an enum as well. Plus, you can map this name to a string such as "7d" which the user can later be allowed to type into some search box. You can just generate an enum off that value and get the corresponding enum type. Or null if invalid.

frontend/src/components/Logs/LogsViewContainer/IndexerLogsContainer.tsx
darunrs: Let's do this in this PR?

frontend/src/constants/DurationMap.ts
```diff
@@ -0,0 +1,39 @@
+import {
```
darunrs, line 1: Like mentioned using an enum would help with this. Let's also define all related things in the same file too. I don't see why TImeIntervals and DURATION_MAP should be separate when they're tightly coupled.

frontend/src/components/Logs/GraphQL/GetSearchLogsQueryDefinition.ts
morgsmccauley: My two main concerns with this are:
1. **Indirect relationship between the code using it and the query itself** - It's hard to distinguish which query will be used under which conditions. Yes, the query and function names describe the variables used, but we have to assume they are correct, which may not always be the case. For example, some the the queries which mention block height (e.g. `getBlockHeightAndMessageSearchQueryWithSeverityAndLogTypeAndTimeRange`) actually have a hard-coded value of `123`. I assume this is a bug, and confirms that we can't trust the names for these alone.
2. **Complicated ongoing maintenance** - All queries are essentially the same, they are just used in different ways. If we wanted to query a new field, we'd have to update _all_ queries. It's even worse for adding a new variable/parameter. Let's say we want to paginate, essentially _every_ query needs to add a new parameter, so we'd need to double the queries so that they work with and without pagination?

It would be easier to both reason about and maintain if we just express it this for what it really is - a single query which is used in multiple different ways. Disregarding `table_name`, we can reduce this to the query below. I would only opt for more queries if they are truely distinct, i.e. fetching completely different data.

Again note that `{}` reduces to `true` in Hasura, so variables which aren't passed essentially get stripped from the underlying Postgres query. This also makes the logic simpler, we no longer need `_or`, we can just `_and` everything, and only the variables which are passed end up in the resulting filter. All we need to pass are the variables which exist (have been selected in the UI) and the query takes care of the rest.

Going back to my initial concerns; the relationship is now direct, we know which query is being used. If we need to debug we can inspect the network requests being made in the console. Ongoing maintenance is also simpler; adding a new field or variable is a 1 or 2 line change.

```graphql
  query LogsQuery(
    $limit: Int = 10,
    $offset: Int = 0,
    $order_by_timestamp: order_by = desc,
    $level: String_comparison_exp = {},
    $type: String_comparison_exp = {},
    $message: String_comparison_exp = {},
    $block_height: numeric_comparison_exp = {},
    $timestamp: timestamp_comparison_exp = {},
  ) {
    darunrs_near_bitmap_v5_sys_logs(
      limit: $limit,
      offset: $offset,
      order_by: {timestamp: $order_by_timestamp},
      where: {
        _and: [
          {level: $level},
          {type: $type},
          {message: $message},
          {block_height: $block_height},
          {timestamp: $timestamp}
        ]
      }
    ) {
      block_height
      level
      message
      timestamp
      type
      
    }
    darunrs_near_bitmap_v5_sys_logs_aggregate(
      where: {
        _and: [
          {level: $level},
          {type: $type},
          {message: $message},
          {block_height: $block_height},
          {timestamp: $timestamp}
        ]
      }
    ) {
      aggregate {
        count
      }
    }
  }
```

To provide additional context, here are some example variable sets:
- Filter by level:
```json
{
  "level": { "_eq": "INFO" }
}
```
- Filter by type and search by message
```json
{
  "type": { "_eq": "system" },
  "message": { "_ilike": "%Running%" }
}
```

frontend/src/components/Logs/LogsViewContainer/IndexerLogsContainer.tsx
```diff
@@ -0,0 +1,199 @@
...
+        return {
+            url: GRAPHQL_ENDPOINT,
+            method: 'POST',
+            headers: {
+                ['x-hasura-role']: sanitizedAccountId,
+                ['Content-Type']: 'application/json',
+            },
+            body: JSON.stringify({
+                query: Query(tableName),
+                variables: QueryValidation({ limit: LOGS_PER_PAGE, offset: 0, order_by_timestamp: 'desc', level: severity, type: logType, timestamp: startTime })
```
morgsmccauley, line 57: Does `offset` need to change based on page number?

morgsmccauley (approved):
Nice work, I like how you separated out the smaller view components 🙌🏼

frontend/src/components/Logs/LogsViewContainer/IndexerLogsContainer.tsx
Kevin101Zhang: We do not have any form of pagination currently. Each filter/added filter is its individual call. But yes it would have to change when we implement infinite scroll.

darunrs (approved):
LGTM! Nice work on it! The broken up files are much easier to ingest mentally.

PR #729, opened 2024-05-13
feat: Expose RPC from Coordinator to enable/disable Indexers

### LINKED ISSUE #667: Ability to manually suspend indexers

morgsmccauley:
This PR exposes a new gRPC endpoint from Coordinator to "manage" indexers. Currently, this only allows for  enabling/disabling, but will probably be expanded over time. There isn't any intention to use this from another service, it's more of a manual & internal tool that we can use.

The endpoint is essentially just a wrapper over the persistent Redis state. The exposed methods end up mutating this state, which in turn, is then used to govern how Indexers should be synchronised.

Within the `coordinator/` directory, the endpoint can be used with `grpcurl` like so:
- enable:  `grpcurl -plaintext -proto proto/indexer_manager.proto -d '{"account_id": "morgs.near", "function_name": "test"}'  0.0.0.0:8002 indexer.IndexerManager.Enable`
- disable:  `grpcurl -plaintext -proto proto/indexer_manager.proto -d '{"account_id": "morgs.near", "function_name": "test"}'  0.0.0.0:8002 indexer.IndexerManager.Disable`
- list:  `grpcurl -plaintext -proto proto/indexer_manager.proto 0.0.0.0:8002 indexer.IndexerManager.List`

### DIFF
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -2,7 +2,7 @@
 
 use std::cmp::Ordering;
 
-use crate::indexer_config::{IndexerConfig, IndexerIdentity};
+use crate::indexer_config::IndexerConfig;
 use crate::redis::RedisClient;
 use crate::registry::IndexerRegistry;
 
@@ -48,8 +48,8 @@ impl IndexerStateManagerImpl {
         Self { redis_client }
     }
 
-    pub async fn get_state(&self, identity: &IndexerIdentity) -> anyhow::Result<IndexerState> {
-        let raw_state = self.redis_client.get_indexer_state(identity).await?;
+    pub async fn get_state(&self, indexer_config: &IndexerConfig) -> anyhow::Result<IndexerState> {
+        let raw_state = self.redis_client.get_indexer_state(indexer_config).await?;
 
         if let Some(raw_state) = raw_state {
             return Ok(serde_json::from_str(&raw_state)?);
@@ -60,25 +60,25 @@ impl IndexerStateManagerImpl {
 
     async fn set_state(
         &self,
-        identity: &IndexerIdentity,
+        indexer_config: &IndexerConfig,
         state: IndexerState,
     ) -> anyhow::Result<()> {
         let raw_state = serde_json::to_string(&state)?;
 
         self.redis_client
-            .set_indexer_state(identity, raw_state)
+            .set_indexer_state(indexer_config, raw_state)
             .await
     }
 
     pub async fn set_enabled(
         &self,
-        identity: &IndexerIdentity,
+        indexer_config: &IndexerConfig,
         enabled: bool,
     ) -> anyhow::Result<()> {
-        let mut indexer_state = self.get_state(identity).await?;
+        let mut indexer_state = self.get_state(indexer_config).await?;
         indexer_state.enabled = enabled;
 
-        self.set_state(identity, indexer_state).await?;
+        self.set_state(indexer_config, indexer_state).await?;
 
         Ok(())
     }
@@ -91,7 +91,7 @@ impl IndexerStateManagerImpl {
 
         for (account_id, indexers) in indexer_registry.iter() {
             for (function_name, indexer_config) in indexers.iter() {
-                let indexer_state = self.get_state(&indexer_config.into()).await?;
+                let indexer_state = self.get_state(indexer_config).await?;
 
                 if indexer_state.enabled {
                     filtered_registry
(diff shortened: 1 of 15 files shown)


coordinator/src/indexer_state.rs
```diff
@@ -55,33 +70,104 @@ impl IndexerStateManagerImpl {
...
+        Ok(filtered_registry)
+    }
+
     pub async fn migrate_state_if_needed(
         &self,
         indexer_registry: &IndexerRegistry,
     ) -> anyhow::Result<()> {
-        if self.redis_client.is_migration_complete().await?.is_some() {
-            return Ok(());
+        if self.redis_client.is_migration_complete().await?.is_none() {
```
darunrs, line 112: Why do we still check for and try to perform the original migration? Wasn't it completed in both dev and prod already anyway?

coordinator/src/indexer_state.rs
```diff
@@ -55,33 +70,104 @@ impl IndexerStateManagerImpl {
...
-                        },
-                    )
-                    .await?;
+        if self
+            .redis_client
+            .get::<_, bool>("state_migration:enabled_flag")
+            .await?
+            .is_none()
+        {
+            tracing::info!("Migrating enabled flag");
```
darunrs, line 143: nit: I feel like this is the log which should say "Migrating indexer state" whereas the section above this should be maybe "Completing old migration" or something. 

Or at least "Migration flag enabled. Migrating...".

coordinator/src/indexer_state.rs
```diff
@@ -13,11 +13,26 @@ pub enum SyncStatus {
...
-#[derive(Default, Debug, Clone, serde::Serialize, serde::Deserialize)]
-struct IndexerState {
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct OldIndexerState {
     block_stream_synced_at: Option<u64>,
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct IndexerState {
+    pub block_stream_synced_at: Option<u64>,
```
darunrs, line 23: Do we still need this? What about simply removing it? Even if migrating state goes wrong, losing this information doesn't impact us, I think.

coordinator/src/indexer_state.rs
```diff
@@ -136,7 +222,79 @@ mod tests {
...
+                        .to_string(),
+                ))
+            })
+            .once();
+        mock_redis_client
+            .expect_get_indexer_state()
+            .with(predicate::eq(darunrs_config.clone()))
+            .returning(|_| {
+                Ok(Some(
+                    serde_json::json!({ "block_stream_synced_at": 1, "enabled": false })
```
darunrs, line 280: How dare you disable my indexer! :rage:

darunrs (approved):
Looks good overall! Could you also add the new env variable to the Docker Compose? Also, as far as running grpcurl, how would we do that in the dev/prod machines? Should we install grpcurl in the Dockerfile so that it is available if you enter the docker container?

coordinator/src/indexer_state.rs
morgsmccauley: Ah sorry, I meant to call this out - Only dev has been migrated, so we still need this so prod can be migrated. They also need to be separate steps because dev is partially migrated. Essentially, this code will work for both dev/prod, even with them being in different states.

coordinator/src/indexer_state.rs
morgsmccauley: The first migration will not run for dev, only prod, and for prod it isn't really "old".

These are only temporary logs, as the migration will be removed once completed. I just wanted something to signify start/end, the logs themselves don't really matter.

coordinator/src/indexer_state.rs
morgsmccauley: Yes. This is the "stream version", previously stored under e.g. `morgs.near/test:block_stream:version`, but rather than have it as a separate key it's stored with under Indexer state. This is what prevents us from starting a historical backfill again.

coordinator/src/indexer_state.rs
morgsmccauley: 😅

morgsmccauley:
> how would we do that in the dev/prod machines?

@darunrs we'll need to ssh tunnel to the machine like we do with the 'examples', and can then run `grpcurl` from our local machine.

PR #730, opened 2024-05-14
refactor: Simplify nested registry iteration

morgsmccauley:
In Coordinator, to iterate over all Indexers you need two loops: one for the accounts, and another for its functions. As iteration is quite common, I've added a custom `Iterator` implementation which achieves the same with a single loop.

### DIFF
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -65,17 +65,15 @@ impl IndexerStateManagerImpl {
 
         tracing::info!("Migrating indexer state");
 
-        for (_, indexers) in indexer_registry.iter() {
-            for (_, indexer_config) in indexers.iter() {
-                if let Some(version) = self.redis_client.get_stream_version(indexer_config).await? {
-                    self.set_state(
-                        indexer_config,
-                        IndexerState {
-                            block_stream_synced_at: Some(version),
-                        },
-                    )
-                    .await?;
-                }
+        for indexer_config in indexer_registry.iter() {
+            if let Some(version) = self.redis_client.get_stream_version(indexer_config).await? {
+                self.set_state(
+                    indexer_config,
+                    IndexerState {
+                        block_stream_synced_at: Some(version),
+                    },
+                )
+                .await?;
             }
         }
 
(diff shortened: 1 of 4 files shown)


coordinator/src/indexer_state.rs
```diff
@@ -87,18 +87,17 @@ impl IndexerStateManagerImpl {
...
 
-                if indexer_state.enabled {
-                    filtered_registry
-                        .entry(account_id.clone())
-                        .or_default()
-                        .insert(function_name.clone(), indexer_config.clone());
-                }
+            if indexer_state.enabled {
+                filtered_registry
+                    .0
```
darunrs, line 97: What is this `.0`? Is this another way of getting the first item in an array? Curious why it's an array anyway if so.

coordinator/src/registry.rs
```diff
@@ -13,7 +14,58 @@ use registry_types::AllIndexers;
...
+
+    pub fn iter(&self) -> IndexerRegistryIter {
+        IndexerRegistryIter {
+            account_iter: self.0.iter(),
+            function_iter: None,
+        }
+    }
+}
+
+pub struct IndexerRegistryIter<'a> {
```
darunrs, line 37: Is the position of the iterator stateful? Like, if I looped through the registry, and then broke the loop half way. And then tried to iterate again.  Would it reset or pick up where I left off?

darunrs (approved):
Cool refactoring!

coordinator/src/indexer_state.rs
morgsmccauley: `IndexerRegistry` is a "Tuple Struct", `.0` is how you access elements within a tuple

coordinator/src/registry.rs
morgsmccauley: The iterator itself is stateful, yes. That would be possible, but you'd need to store the iterator outside the look, i.e.
```rust
const iter = registry.iter();

for i in iter {
  // break
}

for j in iter {
}
```

If you called `.iter()` twice you'd get two separate stateful instances.

PR #738, opened 2024-05-16
fix: Ensure Terminating Thread error is correctly templatized

darunrs:
The error given to Stream Handler is occasionally a JSON object in disguise, and not an Error. As a result, calling error.toString() returns `[Object object]` rather than the error contents. I've added a check so that if the result of toString() is the earlier value, return a JSON.stringify result instead. Doing JSON.stringify on a proper Error type results in `{}`. The two cases must be handled separately. 

To test this, I created test indexers which called one of the two pieces of code which throw unawaited async errors. I verified both wrote the error message and stack trace into the log table. 

```
const timeoutPromise = new Promise((_, reject) => {
  setTimeout(() => {
      reject(new Error('Error thrown after 100ms'));
  }, 100);
});
```
```context.db.IndexerStorage.upsert({}, [], []); ```

### DIFF
--- a/runner/src/stream-handler/stream-handler.ts
+++ b/runner/src/stream-handler/stream-handler.ts
@@ -67,7 +67,7 @@ export default class StreamHandler {
     indexer.setStoppedStatus().catch((e) => {
       this.logger.error('Failed to set stopped status for indexer', e);
     });
-    const errorAsString = error.toString() === '[object Object]' ? JSON.stringify(error) : error.toString();
+    const errorAsString = error instanceof Error ? error.toString() : JSON.stringify(error);
     const streamErrorLogEntry = LogEntry.systemError(`Encountered error processing stream: ${this.indexerConfig.redisStreamKey}, terminating thread\n${errorAsString}`, this.executorContext.block_height);
 
     indexer.writeCrashedWorkerLog(streamErrorLogEntry)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: runner/src/stream-handler/stream-handler.ts [hunk 1]
```diff
@@ -67,8 +67,8 @@ export default class StreamHandler {
     indexer.setStoppedStatus().catch((e) => {
       this.logger.error('Failed to set stopped status for indexer', e);
     });
-
-    const streamErrorLogEntry = LogEntry.systemError(`Encountered error processing stream: ${this.indexerConfig.redisStreamKey}, terminating thread\n${error.toString()}`, this.executorContext.block_height);
+    const errorAsString = error.toString() === '[object Object]' ? JSON.stringify(error) : error.toString();
```
morgsmccauley, line 70:
```suggestion
    const errorMessage = error instanceof Error ? error.toString() : JSON.stringify(error);
```

I believe `[Object object]` would only appear if we are throwing an Object somewhere. I think we can just change this to the above to ensure we are only calling `toString()` on `Error`s.

morgsmccauley (approved):

runner/src/stream-handler/stream-handler.ts
darunrs: Tested this and it seems to work. I think this is cleaner so I'll go with this. Thanks!

PR #745, opened 2024-05-20
feat: Query Hasura Endpoint through Block Streamer

### LINKED ISSUE #741: Integrate graphQL Querying into Block Streamer
Adds ability to make queries to a Hasura endpoint to Block Streamer.

darunrs:
Block Streamer will be querying block match data from an Indexer's postgres tables through Hasura. Thus, Block Streamer needs to be able to query and parse data returned by Hasura's graphQL APIs. 

This PR introduces a crate which can generate code necessary to parse returned data from a graphQL query. I've also created a struct which encapsulates the code for making graphQL calls for ease of use when integrating or mocking this feature in the future.


block-streamer/examples/get_bitmaps.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use std::error::Error;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let client = graphql::GraphqlClient::new(
+        "https://queryapi-hasura-graphql-mainnet-vcqilefdcq-ew.a.run.app/v1/graphql".to_string(),
+    );
+
+    let exact_query = client
```
darunrs, line 12: Both queries are necessary as exact match is actually significantly faster than the wildcard query if we are matching against exact receiver IDs.

block-streamer/Cargo.toml
```diff
@@ -29,10 +33,6 @@ tokio-stream = "0.1.14"
 tonic = "0.10.2"
 wildmatch = "2.1.1"
 
-registry-types = { path = "../registry/types" }
```
morgsmccauley, line 32: Nit: can we keep local crates separate 🙏🏼

block-streamer/src/graphql/client.rs
```diff
@@ -0,0 +1,126 @@
...
+            .header("x-hasura-role", HASURA_ACCOUNT)
+            .json(&request_body)
+            .send()
+            .await
+            .expect("Failed to query bitmaps for list of exact receivers");
+        let response_body: Response<get_bitmaps_exact::ResponseData> = res.json().await?;
+        match response_body.data {
+            Some(data) => Ok(data.darunrs_near_bitmap_v5_actions_index),
+            None => Ok([].into()),
+        }
```
morgsmccauley, line 53: This can be reduced to the following by using [post_graphql](https://docs.rs/graphql_client/latest/graphql_client/reqwest/fn.post_graphql.html), but you'll need to enable the `reqwest` feature for `graphql_client`.

```rust
post_graphql::<GetBitmapsExact, _>(
    &self.client,
    &self.graphql_endpoint,
    get_bitmaps_exact::Variables {
        receiver_ids: Some(receiver_ids),
        block_date: Some(block_date),
        limit: Some(limit),
        offset: Some(offset),
    },
)
.await?
.data
.ok_or(anyhow!("No data"))
.map(|data| data.darunrs_near_bitmap_v5_actions_index)

```

block-streamer/src/graphql/client.rs
```diff
@@ -0,0 +1,126 @@
...
+    pub async fn get_bitmaps_exact(
+        &self,
+        receiver_ids: Vec<String>,
+        block_date: String,
+        limit: i64,
+        offset: i64,
+    ) -> Result<
+        Vec<get_bitmaps_exact::GetBitmapsExactDarunrsNearBitmapV5ActionsIndex>,
+        Box<dyn Error>,
+    > {
```
morgsmccauley, line 33:
```suggestion
    ) -> anyhow::Result<Vec<get_bitmaps_exact::GetBitmapsExactDarunrsNearBitmapV5ActionsIndex>> {
```

`anyhow` is essentially the same, but is easier to work with

block-streamer/src/graphql/queries/get_bitmaps_exact.rs
```diff
@@ -0,0 +1,49 @@
+// TODO: Replace this file with query agaisnt production Indexer
```
morgsmccauley, line 1: This is pretty verbose, can we use [#[derive(GraphQLQuery)]](https://docs.rs/graphql_client/latest/graphql_client/trait.GraphQLQuery.html) instead?

block-streamer/src/graphql/client.rs
```diff
@@ -0,0 +1,126 @@
+use crate::graphql::queries::get_bitmaps_exact::{get_bitmaps_exact, GetBitmapsExact};
+use crate::graphql::queries::get_bitmaps_wildcard::{get_bitmaps_wildcard, GetBitmapsWildcard};
+use ::reqwest;
+use graphql_client::{GraphQLQuery, Response};
+use std::error::Error;
+
+// TODO: Use Dataplatform account
+const HASURA_ACCOUNT: &str = "darunrs_near";
+
+pub struct GraphqlClient {
```
morgsmccauley, line 10:
```suggestion
pub struct GraphQLClient {
```
Nit: the QL are usually uppercase

block-streamer/README.md
```diff
@@ -0,0 +1,14 @@
+// TODO: Improve README further
+
+## GraphQL Code Generation
```
morgsmccauley, line 3: Why don't we just store the schema in git? Then everyone else doesn't need to go through this process. I think we need to store it for `#[derive(GraphQLQuery)]` anyway.

morgsmccauley: Might be better to export as `.graphql` instead of `.json` though, as the latter is quite lengthly. This [doc](https://hasura.io/docs/latest/schema/common-patterns/export-graphql-schema/) shows how.

block-streamer/Cargo.toml
```diff
@@ -13,11 +13,15 @@ borsh = "0.10.2"
...
 futures = "0.3.5"
+graphql_client = "0.14.0"
 lazy_static = "1.4.0"
 mockall = "0.11.4"
+near-lake-framework = "0.7.8"
 prometheus = "0.13.3"
 prost = "0.12.3"
 redis = { version = "0.21.5", features = ["tokio-comp", "connection-manager"] }
+registry-types = { path = "../registry/types" }
+reqwest = { version = "^0.12.4", features = ["json", "blocking"] }
```
morgsmccauley, line 24:
```suggestion
reqwest = { version = "0.11", features = ["json"] }
```
- `blocking` shouldn't be necessary since we're using `async`?
- `0.11` is the version used internally to `graphql_client` so this makes them compatible

block-streamer/src/graphql/queries/get_bitmaps_exact.rs
darunrs: The reason I switched away from using derive is because it would require having the schema file available as well. I thought this would be more inconvenient to keep around. I can look into generating the schema.graphql file like you mentioned earlier. If it's less verbose, that might be the move.

block-streamer/src/graphql/client.rs
darunrs: I actually can't use that directly has we need to attach headers to the client call and that function doesn't allow for it. I can use your return object optimizations you provided though.

darunrs: Actually I'll just re-implement their function.

block-streamer/src/graphql/client.rs
```diff
@@ -0,0 +1,147 @@
...
+    normalization = "rust"
+)]
+struct GetBitmapsWildcard;
+
+pub struct GraphQLClient {
+    client: reqwest::Client,
+    graphql_endpoint: String,
+}
+
+async fn post_graphql<Q: GraphQLQuery, U: reqwest::IntoUrl>(
```
morgsmccauley, line 34: Let's just add this to `GraphQLClient`, then we don't need to pass the `client` and `url`, we can just use `&self`

block-streamer/src/graphql/client.rs
```diff
@@ -0,0 +1,147 @@
...
+            &self.graphql_endpoint,
+            get_bitmaps_wildcard::Variables {
+                receiver_ids: Some(receiver_ids),
+                block_date: Some(block_date),
+                limit: Some(limit),
+                offset: Some(offset),
+            },
+        )
+        .await
+        .expect("Failed to query bitmaps for list of wildcard receivers")
```
morgsmccauley, line 103: This panics, which kills the entire process. We should propagate the `Err` instead. We can just replace this with `?`.

block-streamer/README.md
```diff
@@ -0,0 +1,14 @@
+// TODO: Improve README further
```
morgsmccauley, line 1: Do you need to update this to match the new workflow?

block-streamer/src/graphql/client.rs
darunrs: I actually tried to do this but I was getting into some trouble with the parameter types (Q and U). automock was asking them to be static lifetimes and I couldn't get it to work. Do you know what I should do to make that work?

block-streamer/README.md
darunrs: Ah yes. Let me go update it.

block-streamer/src/graphql/client.rs
darunrs: Ah that's it. The issue I was having was when I retained passing self.client and such to the function, it was saying self would escape the function and need to outlive static. But by making them internally called, the lifetime of self is bound by the call to post_graphql rather than to that function AND get_bitmaps_exact.

darunrs:
Should I also move `src/graphql/client.rs` into the src folder as `src/graphql_client.rs` instead? The name clashes with the graphql_client crate we are using but Rust doesn't seem to mind when I tried it. It build and ran. We technicalyl no longer need the graphql folder since only the client lives inside it now. Previously it was the codegen files.

morgsmccauley (approved):

PR #747, opened 2024-05-24
feat: Implement Block Streamer Bitmap Operations

### LINKED ISSUE #742: Implement Bitmap Operations into Block Streamer
Create a struct which can perform all necessary bitmap operations provided it is given correct data. Specifically, given a day's worth of bitmaps, properly merge them all together and return a starting blockheight and a single bitmap indicating matching blocks from that block onward.

darunrs:
The bitmap indexer will return a list of bitmaps in the form of base 64 strings, and associated start block heights. We need a way to convert all that data into a single block height and an associated bitmap. 

This PR introduces a new BitmapOperator class which holds all the operations necessary to perform the function of returning a combined binary bitmap with the lowest start block height as index 0.

### DIFF
--- a/block-streamer/src/bitmap.rs
+++ b/block-streamer/src/bitmap.rs
@@ -1,8 +1,6 @@
 use anyhow::anyhow;
 use base64::{engine::general_purpose, Engine as _};
 
-// const BLOCK_HEIGHTS_IN_DAY: usize = 86000;
-
 pub struct Base64Bitmap {
     pub start_block_height: usize,
     pub base64: String,
--- a/block-streamer/src/bitmap.rs
+++ b/block-streamer/src/bitmap.rs
@@ -11,6 +11,7 @@ pub struct Bitmap {
     pub bitmap: Vec<u8>,
 }
 
+#[derive(Default)]
 struct EliasGammaDecoded {
     pub value: usize,
     pub last_bit_index: usize,
@@ -24,11 +25,11 @@ impl BitmapOperator {
         Self {}
     }
 
-    pub fn get_bit(&self, byte_array: &[u8], bit_index: usize) -> bool {
+    pub fn get_bit(&self, bytes: &[u8], bit_index: usize) -> bool {
         let byte_index: usize = bit_index / 8;
         let bit_index_in_byte: usize = bit_index % 8;
 
-        (byte_array[byte_index] & (1u8 << (7 - bit_index_in_byte))) > 0
+        (bytes[byte_index] & (1u8 << (7 - bit_index_in_byte))) > 0
     }
 
     fn set_bit(&self, byte_array: &mut [u8], bit_index: usize, bit_value: bool, write_zero: bool) {
@@ -39,7 +40,7 @@ impl BitmapOperator {
         }
     }
 
-    fn get_number_between_bits(
+    fn read_integer_from_binary(
         &self,
         byte_array: &[u8],
         start_bit_index: usize,
@@ -56,24 +57,20 @@ impl BitmapOperator {
         number
     }
 
-    fn index_of_first_bit(
-        &self,
-        byte_array: &[u8],
-        start_bit_index: usize,
-    ) -> anyhow::Result<usize> {
+    fn index_of_first_set_bit(&self, byte_array: &[u8], start_bit_index: usize) -> Option<usize> {
         let mut first_bit_index: usize = start_bit_index % 8;
-        for byte_index in (start_bit_index / 8)..byte_array.len() {
-            if byte_array[byte_index] > 0 {
+        for (byte_index, byte) in byte_array.iter().enumerate().skip(start_bit_index / 8) {
+            if *byte > 0 {
                 for bit_index in first_bit_index..=7 {
-                    if byte_array[byte_index] & (1u8 << (7 - bit_index)) > 0 {
-                        return Ok(byte_index * 8 + bit_index);
+                    if *byte & (1u8 << (7 - bit_index)) > 0 {
+                        return Some(byte_index * 8 + bit_index);
                     }
                 }
             }
             first_bit_index = 0;
         }
 
-        Err(anyhow!("Failed to find a bit with value 1 in byte array"))
+        None
     }
 
     fn decode_elias_gamma_entry(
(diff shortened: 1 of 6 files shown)


block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+            }
+
+            decompressed_bit_index += decoded_elias_gamma.value;
+            current_bit_value = !current_bit_value;
+        }
+
+        decompressed_byte_array
+    }
+
+    fn merge_compressed_bitmap_into_base_bitmap(
```
morgsmccauley, line 157: What is the difference between `base_bitmap` and `compressed_bitmap`? Maybe this would be more obvious if we just had a `merge` function, and called `decompress` from the outside?

morgsmccauley: `merge` could even be defined on the `Bitmap` struct instead for further clarity

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+                start_bit_index + bit_index_offset,
+                decompressed_bit_value,
+                false,
+            );
+        }
+
+        Ok(())
+    }
+
+    pub fn get_merged_bitmap(
```
morgsmccauley, line 192:
```suggestion
    pub fn merge_bitmaps(
```
Nit: this seems more clear?

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+        Ok(merged_bitmap)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_getting_bit_from_array() {
```
morgsmccauley, line 221:
```suggestion
    fn getting_bit_from_array() {
```
Nit: `test_` seems superfluous here

morgsmccauley (approved):
Honestly, I have no idea what's going on here, mainly due to unfamiliarity with the algorithm used. It will take some time for me to understand, so rather than block you, I've provided the best review I can with my limited knowledge. I'm trusting that you know what's going on 😄.

If it's possible, it would be good to break up the code up in to more well-defined/cohesive methods, but that doesn't need to happen here.

There are a lot of `clippy` errors, could you run `cargo clippy` and fix please?

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    pub fn get_bit(&self, byte_array: &[u8], bit_index: usize) -> bool {
+        let byte_index: usize = bit_index / 8;
+        let bit_index_in_byte: usize = bit_index % 8;
+
+        (byte_array[byte_index] & (1u8 << (7 - bit_index_in_byte))) > 0
+    }
+
+    fn set_bit(&self, byte_array: &mut [u8], bit_index: usize, bit_value: bool, write_zero: bool) {
```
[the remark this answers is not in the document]
darunrs: This is necessary because we sometimes want to write the 0 over a 1. Specifically when we are replacing one Elias Gamma encoding over another, as the length might be shorter (leaving extra 1's that should be zero). Technically we don't need it in the current code, but I ported it over as its exactly how we have it in the indexer logic.

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+        if !bit_value && write_zero {
+            byte_array[bit_index / 8] &= !(1u8 << (7 - (bit_index % 8)));
+        } else if bit_value {
+            byte_array[bit_index / 8] |= 1u8 << (7 - (bit_index % 8));
+        }
+    }
+
+    fn get_number_between_bits(
```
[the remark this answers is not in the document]
darunrs: Hmm basically we encoded a number as binary and are simply reading the binary value from a particular stretch of bits. Perhaps I can rename this to `read_unsigned_integer_from_binary`? Even though all our functions deal with binary, maybe this can explicitly state this binary is utilized to build an integer.

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+        for curr_bit_index in (start_bit_index..=end_bit_index).rev() {
+            if self.get_bit(byte_array, curr_bit_index) {
+                number |= 1u32 << (end_bit_index - curr_bit_index);
+            }
+        }
+
+        number
+    }
+
+    fn index_of_first_bit(
```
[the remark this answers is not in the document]
darunrs: Maybe I can go with `index_of_first_set_bit` instead? I can't think of a better way to refer to a bit with value 1 other than saying set bit or outright saying "bit with value 1". I think `set bit` is more clear than what I originally had.

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+    fn decode_elias_gamma_entry(
+        &self,
+        byte_array: &[u8],
+        start_bit_index: usize,
+    ) -> EliasGammaDecoded {
+        if byte_array.len() == 0 {
+            return EliasGammaDecoded {
+                value: 0,
+                last_bit_index: 0,
+            };
```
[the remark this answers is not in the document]
darunrs: That's a good idea actually. It would definitely make the match look nicer too.

block-streamer/src/bitmap.rs
```diff
@@ -0,0 +1,367 @@
...
+        EliasGammaDecoded {
+            value: 2_usize.pow(zero_count.try_into().unwrap()) + remainder,
+            last_bit_index: first_bit_index + zero_count,
+        }
+    }
+
+    fn decompress_bitmap(&self, compressed_bitmap: &[u8]) -> Vec<u8> {
+        let compressed_bit_length: usize = compressed_bitmap.len() * 8;
+        let mut current_bit_value: bool = (compressed_bitmap[0] & 0b10000000) > 0;
+        let mut decompressed_byte_array: Vec<u8> = Vec::new();
```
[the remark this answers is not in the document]
darunrs: We technically don't. We do know the upper bound, which is 86000 bits, since 1 bit per block and 86000 seconds in a day. But, I felt it was unnecessary to create 12KB byte arrays every time as we usually don't need that many.

block-streamer/src/bitmap.rs
darunrs: Yeah that probably makes more sense than a standalone function. I'll define it on Bitmap. Thanks for the idea! I think I was treating the data structs as more like enums when they're really classes still.

block-streamer/src/bitmap.rs
darunrs: I'll go with `merge_compressed_bitmaps`. This way, it's clear this is taking in compressed bitmaps, and the argument type Base64Bitmap will also indicate this requires a undecoded compressed bitmap too.

block-streamer/src/bitmap.rs
darunrs: That's true. I'll reword the test names.

PR #758, opened 2024-05-29
feat: Expose gRPC service to handle provisioning

### LINKED ISSUE #759: Expose endpoint from Runner to control provisioning

morgsmccauley:
This PR adds the `DataLayerService`, which allows for provisioning the "Data Layer" of a QueryAPI Indexer. Data Layer in this case refers to the Postgres Database and Hasura GraphQL endpoint. I wanted to move away from the more general "provisioning" term as it is a bit overloaded, and may get conflated with other provisioning steps related to Executors (which may become more involved with Firecracker). This PR doesn't replace the existing provisioning flow - yet, I'll do that in a follow up PR.

As provisioning can take some time, rather than keeping the request/connection open while we wait, the `Provision` command triggers the process, and returns immediately. The `CheckProvsioningStatus` command will then be used to poll its completion.

On top the new service, the following changes have been made:
- the gRPC `server/` directory has been slightly restructured to accommodate additional services, i.e. `DataLayerService`, as a result you'll see lots of `runner` files being moved around
- `Provisioner` now takes `ProvisioningConfig`, which is a subset of `IndexerConfig`. On the gRPC side we don't have a full `IndexerConfig`, nor do we need it, so I created the subset to make it possible to provision with only the data we need. `IndexerConfig extends ProvisioningConfig` so it's not a breaking change to the existing use-cases.
- I've removed all the unused code in `Provisioner`


runner/src/logger.ts
```diff
@@ -23,15 +23,17 @@ const logger = winston.createLogger({
...
-if (process.env.GCP_LOGGING_ENABLED) {
-  logger.add(new LoggingWinston({ redirectToStdout: true }));
-} else {
-  logger.add(new transports.Console({
-    format: format.combine(
-      format.colorize(),
-      format.simple(),
-    ),
-  }));
+if (process.env.NODE_ENV !== 'test') {
```
morgsmccauley, line 26: Silence logs in test

gabehamilton:
Where did all the provisioner.ts code move to? Provisioning the logs & metadata.

morgsmccauley:
> Where did all the provisioner.ts code move to? Provisioning the logs & metadata.

The removed code was a provision hook for _existing_ Indexers, and is no longer needed. We still provision logs/metadata for _new_ indexers in the main `provisionUserApi()` method.

runner/protos/data-layer.proto
```diff
@@ -0,0 +1,33 @@
...
+    string function_name = 2;
+    string schema = 3;
+}
+
+message CheckProvisioningStatusRequest {
+    string account_id = 1;
+    string function_name = 2;
+}
+
+enum ProvisioningStatus {
```
darunrs, line 24: Will this status be used for any other data layer tasks we will need to do in the future? Such as schema edits or migrating Indexers between versions. We can rename this to TaskStatus or something. Or, if this status is specific to provisioning, should we distinguish between retryable failure modes and manual intervention failure modes? Such as failing to create a PG DB vs Hasura failed to track foreign keys?

runner/src/server/services/data-layer/data-layer-service.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      const callback = (error: any): void => {
+        expect(error.code).toBe(status.ALREADY_EXISTS);
+        expect(error.details).toBe('Provisioning task already exists');
+        done();
+      };
+
+      createDataLayerService(undefined, tasks).Provision(call, callback);
+    });
+
+    it('should return ALREADY_EXISTS if the task has already completed', (done) => {
```
darunrs, line 86: What kind of scenario is this testing? Is this Coordinator tries to start the same task on its next loop or Coordinator trying to start this task when it first connects to the DataLayer service? We are going to store the state of an indexer's data in Redis still as a source of truth right?

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -0,0 +1,110 @@
...
+import { type CheckProvisioningStatusRequest__Output } from '../../../generated/data_layer/CheckProvisioningStatusRequest';
+import { type DataLayerHandlers } from '../../../generated/data_layer/DataLayer';
+import { type ProvisionRequest__Output } from '../../../generated/data_layer/ProvisionRequest';
+import { type ProvisionResponse } from '../../../generated/data_layer/ProvisionResponse';
+import { ProvisioningStatus } from '../../../generated/data_layer/ProvisioningStatus';
+
+export class ProvisioningTask {
+  public failed: boolean;
+  public pending: boolean;
+  public completed: boolean;
```
darunrs, line 15: Why are there three different booleans exposed instead of one Status enum or something?

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    this.failed = false;
+    this.pending = true;
+    this.completed = false;
+  }
+}
+
+type ProvisioningTasks = Record<string, ProvisioningTask>;
+
+const generateTaskId = (accountId: string, functionName: string): string => `${accountId}:${functionName}`;
```
darunrs, line 35: This will ensure that DataLayer cannot spin up more than one task for one indexer. I believe this should make sense even going forward when we add more tasks.

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export function createDataLayerService (
+  provisioner: Provisioner = new Provisioner(),
+  tasks: ProvisioningTasks = {}
+): DataLayerHandlers {
+  return {
+    CheckProvisioningStatus (call: ServerUnaryCall<CheckProvisioningStatusRequest__Output, ProvisionResponse>, callback: sendUnaryData<ProvisionResponse>): void {
```
darunrs, line 42: Should this return if a task exists or if the indexer was ever provisioned? If it is the former, maybe we can rename this to CheckProvisioningTaskStatus? 

The way I currently envision how this service is used is that Coordinator reads the permanent Indexer provisioning status from Redis. It then starts a provisioning task if needed. We need to handle partial provisioning failures somehow since not all provisioning errors are retryable and the service could also just restart in the middle of provisioning something successfully. Previously, Runner was doing this unilaterally by checking Postgres and Hasura. I'm wondering if that logic should still exist in DataLayer. It's obviously easier to write, but it would also be nice if we had that information stored in Redis for Coordinator to instead be the one to make that decision. 

Then we need to extend this to have other tasks like I mentioned earlier. So, we should probably figure out a JSON or something else format to store in Redis which makes sense.

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -0,0 +1,110 @@
...
+        const exists = new StatusBuilder()
+          .withCode(status.ALREADY_EXISTS)
+          .withDetails('Provisioning task already exists')
+          .build();
+        callback(exists);
+
+        return;
+      };
+
+      provisioner.fetchUserApiProvisioningStatus(provisioningConfig).then((isProvisioned) => {
```
darunrs, line 87: So seeing this, it seems like the design is to have Coordinator call provision on all indexers and have DataLayer check if it did exist, and cache the result of that as a successful run of the task. Do you think storing the status of an indexer's data in Redis makes sense over using our current function? My issue with the current function is that reading from Hasura is slow and prone to transient failures. Especially when many requests to metadata are made. I think Hasura must be doing something behind the scenes to ensure metadata state across all instances are mathcing, which slows down any calls to it. It makes sense to have Hasura be our source of truth though. What do you think?

darunrs (approved):
Most of my comments are conversational. Happy to proceed with the current PR as its mainly establishing the foundation, but I'd love to get a clearer picture of where we're headed with the service with schema editing and future QueryApi migrations in mind. 

Also, having multiple gRPC dispatch destinations from the same server running on one port was a surprise. Very cool!

darunrs:
Also, sorry I took so long with the review. Really wanted to look at it during the trip but when I saw the PR content, I wanted to be thorough with it. :sweat_smile:

runner/src/logger.ts
morgsmccauley: `test` is set during `npm test`. Nice, I didn't know about `silent`, I'll use that instead :)

runner/protos/data-layer.proto
```diff
@@ -0,0 +1,33 @@
...
+    UNSPECIFIED = 0;
+    PENDING = 1;
+    COMPLETE = 2;
+    FAILED = 3;
+}
+
+message ProvisionResponse {
```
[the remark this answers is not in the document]
morgsmccauley: Yes, my intention is for this service to manage all "Data Layer" related tasks; provisioning, deprovisioning, deleting data etc.

I'm not sure I follow what this DataAction field would do? Regardless, I'm usually against adding things we don't need right now. I'd prefer to do that when we add those features, as we'll probably get the shape wrong if we do it pre-maturely.

runner/protos/data-layer.proto
morgsmccauley: `TaskStatus` is a good idea, but I'm not quite sure if we'd re-use this for the other methods. I can rename if if/when it does get re-used :).

In regards to failure modes, yes, we should eventually have that. But I think that also can be added later. I wanted to go for a 1:1 refactor, just so I can have Coordinator control provisioning, we can build on the robustness later.

runner/src/server/services/data-layer/data-layer-service.test.ts
morgsmccauley: Not really testing any specific scenario. It's unlikely this will happen, but provides a fail safe _in case_ we do something wrong in Coordinator :)

runner/src/server/services/data-layer/data-layer-service.ts
morgsmccauley: Oh yeah true, I could just use an `enum`.

I started the refactor, but think the current state actually reads better, i.e. we can just do `task.completed` or `task.pending`. But with a `enum` we need to pass the type around and do `task.status === TaskStatus.COMPLETED`. 

Which do you think is better? I'll merge as is but can refactor later if you prefer the other :)

runner/src/server/services/data-layer/data-layer-service.ts
morgsmccauley: This only returns relevant information if the task exists, so good point, I will rename to `CheckProvisioningTaskStatus`.

Perhaps, we should be returning be returning a random `taskId`, which Coordinator then uses to check the status of. That would be more clear, but there wouldn't be any reason to have multiple tasks for a given indexer so maybe not 🤔 I'll take a deeper look at this when I integrate with Coordinator.

> Runner was doing this unilaterally by checking Postgres and Hasura. I'm wondering if that logic should still exist in DataLayer

Yes this still exists, if you call `Provision` on an Indexer which has already been provisioned, it will check Pg/Hasura and return `ALREADY_EXISTS`.

runner/src/server/services/data-layer/data-layer-service.ts
morgsmccauley: Yes we should definitely store the provisioning state in Redis, but we should still have this here as a safe guard incase there is a bug in Coordinator which makes us call this again.

Even though we are the only consumers of this API, we may still make mistakes, so we should design it so it can't be abused 😅.

runner/src/server/services/data-layer/data-layer-service.ts
darunrs: Gotcha. I guess it depends on what we end up doing with this list of statuses. If we want to add another one, it would be more difficult. Similarly, if we are passing them around, the parameter type will be a simple boolean, which is fine, but I think TaskStatus more readily describes what this boolean actually is, without needing to encode that in the variable name itself. If we intend to make it a part of another enum, that will also help. Basically, if the usage is simple, I think I also am cool with multiple values. But if we want to do more with them or use them frequently, I feel an enum makes more sense.

PR #762, opened 2024-05-31
refactor: Merge sync logic to provide fine-grained life-cycle control

### LINKED ISSUE #760: Handle provisioning within Coordinator

morgsmccauley:
To synchronise Block Streams/Executors we compare the latest registry with the rest of the system. Components are started/stopped as needed, but we lack control over life-cycle events such as registration/deletion. As we build towards Coordinator managing provisioning, we need a way to hook in to these life-cycle events, so we can provision/de-provision accordingly.

This PR refactors the current synchronisation logic to provide greater control over these life-cycle events. Existing Block Stream/Executor sync logic have been merged to to a single `Synchronisation` struct, which now manages new/existing/deleted indexers as a whole, providing the required life-cycle hooks described above. Merging the sync logic allows us to control _when_ Block Streams/Executors are started/stopped, allowing us to do things before and after, specifically, de-/provisioning. I wanted to do this refactor first to make reviewing easier, I'll follow this up with the provisioning changes next.

To achieve the above, each Indexer now has it's own `IndexerState` object in Redis. After synchronising a newly registered Indexer we write its state object, and after removing it, we delete the state object. This means we have the following states, allowing us to handle each accordingly:
- Non-existent State/Existing Config - New Indexer
- Existing State/Exisiting Config - Existing Indexer
- Exisiting State/Non-existent Config - Deleted Indexer

The synchronisation logic is essentially the same, expect that now it is driven off the persistent Redis state, rather than the current Block Stream/Executors state.


coordinator/src/block_streams_handler.rs
```diff
@@ -62,9 +62,7 @@ impl BlockStreamsHandlerImpl {
             .clone()
             .stop_stream(Request::new(request.clone()))
             .await
-            .map_err(|e| {
-                tracing::error!(stream_id, "Failed to stop stream\n{e:?}");
-            });
+            .context(format!("Failed to stop stream: {stream_id}"))?;
```
morgsmccauley, line 65: Propagate the error up, don't swallow, so we can handle it in `Synchroniser`.

coordinator/src/indexer_state.rs
```diff
@@ -48,14 +54,65 @@ impl IndexerStateManagerImpl {
         Self { redis_client }
     }
 
+    pub async fn migrate(&self, registry: &IndexerRegistry) -> anyhow::Result<()> {
```
morgsmccauley, line 57: This serves two purposes:
1. Migrating from `OldIndexerState` to the new `IndexerState`, and
2. Ensuring all state objects are added to the Redis Set (handled via `set_state()`)

We need 2. to be able to list current `IndexerStates`, we'd need to `scan` the DB otherwise which is not performant.

This migration could leave us in a bad state as we use the "existence" of the state object to signify it's `SynchronisationState`, i.e. new/existing. By using the current registry to update the exisiting states we may end up writing state for an Indexer which was registered but not yet synchronised (i.e. registered while Coordinator was down). We would then incorrectly treat this Indexer as "existing" rather than "new". At this point, this isn't necessarily a problem as the new/existing flows aren't too different, but this is handled in the `sync_existing_block_stream()` method which I'll call out.

coordinator/src/indexer_state.rs
```diff
@@ -1,32 +1,41 @@
...
+    pub function_name: String,
     pub block_stream_synced_at: Option<u64>,
     pub enabled: bool,
 }
 
+impl IndexerState {
+    // FIX `IndexerConfig` does not exist after an Indexer is deleted, and we need a way to
+    // construct the state key without it. But, this isn't ideal as we now have two places which
+    // define this key - we need to consolidate these somehow.
+    pub fn get_state_key(&self) -> String {
```
morgsmccauley, line 29: As described in the comment, this isn't ideal, I'll need to re-think this but didn't want to bloat the current PR.

coordinator/src/synchroniser.rs
```diff
@@ -0,0 +1,1186 @@
+use registry_types::StartBlock;
```
morgsmccauley, line 1: This file is huge, but it's mostly tests.

coordinator/src/synchroniser.rs
```diff
@@ -0,0 +1,1186 @@
...
+        )
+    )]
+    async fn sync_new_indexer(&self, config: &IndexerConfig) -> anyhow::Result<()> {
+        if let Err(err) = self.executors_handler.start(config).await {
+            tracing::error!(?err, "Failed to start Executor");
+            return Ok(());
+        }
+
+        // FIX if this fails, then subsequent control loops will perpetually fail since the
+        // above will error with ALREADY_EXISTS
```
morgsmccauley, line 81: Will handle this in a follow up

coordinator/src/synchroniser.rs
```diff
@@ -0,0 +1,1186 @@
...
+
+            self.reconfigure_block_stream(config).await?;
+
+            return Ok(());
+        }
+
+        if state.block_stream_synced_at.is_none() {
+            // NOTE: A value of `None` would suggest that `state` was created before initialisation,
+            // which is currently not possible, but may be in future
+            tracing::warn!("Existing block stream has no previous sync state, treating as new");
```
morgsmccauley, line 192: As mentioned above with the migration, this is the new/existing case we need to handle. The migration _may_ end up with new indexers being treated as existing, but we'd know if that's the case if `block_stream_synced_at` is `None`.

coordinator/src/synchroniser.rs
```diff
@@ -0,0 +1,1186 @@
...
+                    && block_stream.function_name == state.function_name
+            });
+
+            if let Some(config) = config {
+                sync_states.push(SynchronisationState::Existing(
+                    config,
+                    state,
+                    executor.cloned(),
+                    block_stream.cloned(),
+                ))
```
darunrs, line 307: If a return is used in a for loop, does the for loop just continue on instead of breaking and returning? My impression here is that we should be continuing through the loop as we have more work to do. The return here is confusing me a bit regarding that.

coordinator/src/indexer_state.rs
darunrs: Should Config and State share some interface to guarantee certain fields, which can constitute the key? Such as accountID and functionName. By state key, we really mean the key value in the set right?

coordinator/src/main.rs
```diff
@@ -63,27 +72,11 @@ async fn main() -> anyhow::Result<()> {
 
     loop {
         let indexer_registry = registry.fetch().await?;
+        indexer_state_manager.migrate(&indexer_registry).await?;
```
darunrs, line 75: This only needs to be successful once right? Once it is done, we just remove it? I'm assuming you have it in the loop so you can manually resolve any problems while Coordinator is up.

coordinator/src/main.rs
```diff
@@ -63,27 +72,11 @@ async fn main() -> anyhow::Result<()> {
...
-                &indexer_state_manager,
-                &redis_client,
-                &block_streams_handler
-            ),
-            async {
-                sleep(CONTROL_LOOP_THROTTLE_SECONDS).await;
-                Ok(())
-            }
-        )?;
+        tokio::try_join!(synchroniser.sync(), async {
```
darunrs, line 77: It seems like all sync tasks are designed to succeed and instead just log errors. How do we enable visibility of sync errors beyond manually checking logs?

coordinator/src/synchroniser.rs
```diff
@@ -0,0 +1,1186 @@
...
+        }
+
+        if let Some(block_stream) = block_stream {
+            self.block_streams_handler
+                .stop(block_stream.stream_id.clone())
+                .await?;
+        }
+
+        self.state_manager.delete_state(state).await?;
+
```
darunrs, line 278: Is this where we would call resource clean up? Since Block Streamer is technically responsible for creating the redis block height stream, who is going to delete the redis stream? This would be more important as well when we migrate to storing blocks on redis instead of heights.

darunrs (approved):
Super cool work! I think the existing synchronization logic is more readily understood now and feels like block streams and executors are more logically tied together now.

coordinator/src/synchroniser.rs
morgsmccauley: Oh I didn't even notice this, I'm assuming you are talking about the lack of `;`?

In the context of a for loop, the implicit return value is ignored, so in this case the loop will continue. If I were to add `return` it would break the loop.

coordinator/src/main.rs
morgsmccauley: Good point, yeah, this doesn't even need to be in the loop there isn't actually any benefit it being there 😅

coordinator/src/main.rs
morgsmccauley: Do you mean like Grafana? We'd hook in to those Errors directly. Eventually I think we'd want to short-circuit the sync for that Indexer, store some state, and then act accordingly. But we're not quite there yet.

coordinator/src/synchroniser.rs
morgsmccauley: Coordinator should delete the stream, just like it does when we republish the Indexer. The Redis stream is created implicitly when it is `xadd`ed to, Block Streamer doesn't explicitly create it.

coordinator/src/indexer_state.rs
morgsmccauley: Yup, but we can't do exactly that in Rust.

My current thoughts are to have a common trait, probably `RedisKeyProvider` with some default methods provided which generate the keys using `account_id` and `function_name`. But we can't require fields, we can only require methods, so best we can do is have methods for each of those. Which does still leave room for error but I think is both obvious enough to not be abused, and is better than this :)

> By state key, we really mean the key value in the set right?

Yes, that's right

PR #781, opened 2024-06-07
fixed context db type generation, added overlay(tooltip) component, minor UI changes, Updated Packages and Imports

### LINKED ISSUE #737: Incorrect context.db types
**Describe the bug**
image

**Schema**
```
CREATE TABLE
  "receivers" (
    "id" BIGSERIAL NOT NULL PRIMARY KEY,
    "receiver" TEXT NOT NULL
  );
CREATE UNIQUE INDEX idx_receivers_by_receiver ON receivers (receiver);

CREATE TABLE
  "actions_index" (
    "receiver_id" bigint NOT NULL,
    "block_date" date NOT NULL,
    "first_block_height" int NOT NULL,
    "last_elias_gamma_start_bit" int NOT NULL,
    "max_index" int NOT NULL,
    "bitmap" TEXT NOT NULL,
    PRIMARY KEY ("block_date", "receiver_id"),
    CONSTRAINT "actions_index_receiver_id_fkey" FOREIGN KEY ("receiver_id") REFERENCES "receivers"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
  );

```

### LINKED ISSUE #782: Improvements to Logs/Indexer UI
1̶.̶ ̶A̶d̶d̶ ̶l̶i̶n̶k̶ ̶t̶o̶ ̶q̶u̶e̶r̶y̶ ̶d̶o̶c̶s̶ ̶
2̶.̶ ̶s̶w̶a̶p̶ ̶h̶e̶i̶g̶h̶t̶/̶s̶t̶a̶t̶u̶s̶ ̶i̶n̶d̶i̶c̶a̶t̶o̶r̶ ̶i̶n̶ ̶l̶o̶g̶s̶ ̶
3̶.̶ ̶r̶e̶n̶a̶m̶e̶ ̶A̶p̶p̶ ̶t̶o̶ ̶I̶n̶d̶e̶x̶e̶r̶ ̶
4̶.̶ ̶a̶d̶d̶e̶d̶ ̶o̶v̶e̶r̶l̶a̶y̶/̶t̶o̶o̶l̶t̶i̶p̶ ̶t̶o̶ ̶b̶u̶t̶t̶o̶n̶s̶ ̶(̶t̶r̶a̶s̶h̶/̶r̶e̶g̶e̶n̶e̶r̶a̶t̶e̶ ̶c̶o̶n̶t̶e̶x̶t̶)̶
5. Fix negative blocks issue

Kevin101Zhang:
Just some final touch ups. Spoke with Pavel and he requested a few changes here and there. In the push there are the following changes.

1.  (HTML/JSX CHANGE): Changed SignUp to (Read The Docs)
2. (HTML/JSX CHANGE): Swapped Component position of indexer height and indexer status in logs
3. (HTML/JSX CHANGE): Renamed App.js to Indexer.js 
4. (NEW COMPONENT): Added overlay to buttons
5. (LOGIC REFACTOR/FIX): Fixed context db type generation. Refactored Logic.
6. (CHORE): Updated packages 
7. (CHORE): Removed Dead Code and Refactored relative Imports using base '@' 
8. (NEW COMPONENT): Added NEAR latest block height


frontend/src/components/Common/LatestBlock.tsx
```diff
@@ -0,0 +1,81 @@
...
+                } else {
+                    setErrors("Failed to fetch final block height");
+                }
+            } catch (error) {
+                setErrors((error as Error).message || "Error fetching final block height");
+            }
+        };
+
+        updateFinalBlock();
+        const intervalId = setInterval(updateFinalBlock, 1000);
```
darunrs, line 66: Is this making this effect which would normally run each loop (Or whatever you call the state machine that checks for effect triggers) run at least 1 second later, if not more?

frontend/src/components/Editor/EditorView/EditorMenuView.jsx
```diff
@@ -34,10 +35,12 @@ const EditorMenuView = ({
         </div>
         <ButtonGroup className="mt-3 mt-md-0">
           {isUserIndexer && !isCreateNewIndexer && (
-            <Button variant="outline-primary" size="sm" className="d-flex align-items-center" onClick={() => handleDeleteIndexer()}>
-              <TrashFill size={20} />
-              Trash
-            </Button>
+            <CustomTooltip message="Delete Indexer Code" direction={TooltipDirection.Bottom} >
```
darunrs, line 38: This sounds misleading. This button is the one which deletes the entire indexer right? I would read this and think it just clears my code in monaco. Let's have the tooltip say something like "Deletes Indexer from QueryAPI" or something.

frontend/src/utils/pgSchemaTypeGen.test.ts
```diff
@@ -0,0 +1,7 @@
+export {};
```
darunrs, line 1: Let's add this file when it actually does something?

frontend/src/utils/validator.test.ts
```diff
@@ -1,24 +1,25 @@
+export {}
```
darunrs, line 1: What does this do? Weird to see an export in a test file.

frontend/widgets/src/QueryApi.Dashboard.jsx
```diff
@@ -424,9 +425,9 @@ return (
           QueryApi
         </NavBarLogo>
 
-          <SignUpLink target="_blank" href={`http://bit.ly/near-queryapi-beta`}>
-            (Sign Up)
-          </SignUpLink>
+        <SignUpLink target="_blank" href={`https://docs.near.org/build/data-infrastructure/query-api/intro`}>
+          (Read The Docs)
```
darunrs, line 429: Can we name this `Documentation` instead? I feel that's better than just saying "Read the Docs"

darunrs (approved):
Some small comments. Otherwise, looks good. Deleting the package.lock could be a problem. We'll know when we try to send it up to dev. Also curious to see if it all works in dev too. Such as the context.db fixes and any visual changes.

frontend/src/components/Common/LatestBlock.tsx
Kevin101Zhang: the updateFinalBlock(); is just for the initial call as its en empty dependency array. Afterwards updateFinalBlock is repeatedly called by the setInterval function every second.

frontend/src/components/Editor/EditorView/EditorMenuView.jsx
Kevin101Zhang: Yeah that makes sense will change.

frontend/src/utils/validator.test.ts
Kevin101Zhang: This is due to the isolated flag
```
test.ts' cannot be compiled under '--isolatedModules' because it is considered a global script file. Add an import, export, or an empty 'export {}' statement to make it a module.
```

Ill set the flag in the config file to false and remove them.

PR #803, opened 2024-06-14
feat: Implement Historical Backfill using Bitmap Indexer

### LINKED ISSUE #744: Replace Index File Backfill with Bitmap Indexer Backfill
Fully replace Index file backfill with Bitmap query backfill. This involves switching out the delta lake get_matching_block_heights function call with the BlockHeightStream function of the same name. The yield of the BlockHeightStream call is a single block height instead of a full list, so some behavior of backfill could change, potentially for the better. In addition, the first iteration of the new backfill may have some performance problems. Thus, the performance needs to be evaluated and any less complex improvements should be made (Such as preloading the next day's bitmaps while yielding the current day).

darunrs:
Replaces Delta Lake Stream with Bitmap Indexer Stream as the method for backfilling matching block heights before switching to real time processing. Bitmap Indexer resolves many issues the Delta Lake approach had: 
1. Can be extended more easily.
2. Can be queried much faster thanks to storage in a database as well as in a compressed format.
3. Queries can be customized, enabling better matching. 
4. Supports wildcard contract filters much better than the previous approach. 
5. Is much harder to overload with too much data, which was a problem with the previous approach.


block-streamer/src/block_stream.rs
```diff
@@ -172,81 +173,66 @@ pub(crate) async fn start_block_stream(
...
-    let blocks_from_index = match &indexer.rule {
+    let contract_pattern: String = match &indexer.rule {
         Rule::ActionAny {
             affected_account_id,
             ..
         } => {
             if affected_account_id
                 .split(',')
-                .any(|account_id| DELTA_LAKE_SKIP_ACCOUNTS.contains(&account_id.trim()))
+                .any(|account_id| account_id.trim().eq("*"))
```
darunrs, line 192: Any *.[root account] is specially supported by bitmap indexer so we only need to skip now if it is an all account filter.

block-streamer/src/block_stream.rs
```diff
@@ -172,81 +173,66 @@ pub(crate) async fn start_block_stream(
...
 
-            delta_lake_client
-                .list_matching_block_heights(start_block_height, affected_account_id)
-                .await
+            anyhow::Ok(affected_account_id.to_owned())
         }
         Rule::ActionFunctionCall { .. } => {
             tracing::error!("ActionFunctionCall matching rule not yet supported for delta lake processing, function: {:?} {:?}", indexer.account_id, indexer.function_name);
-            Ok(vec![])
+            Ok("".to_string())
```
darunrs, line 209: I chose to make this match return a contract filter string as returning an "empty" stream sounded like more work than was worth.

block-streamer/src/block_stream.rs
```diff
@@ -336,24 +345,40 @@ mod tests {
 
         mock_lake_s3_client
```
darunrs, line 346: For some reason I kept getting an error from the mock saying the function has no matching expectation, even though it is never used. I had no choice but to keep it in, which felt odd as a default mock should have a default expect for each function right? Maybe I'm doing something wrong?

block-streamer/src/main.rs
```diff
@@ -32,6 +32,7 @@ async fn main() -> anyhow::Result<()> {
     }
 
     let redis_url = std::env::var("REDIS_URL").expect("REDIS_URL is not set");
+    let graphql_url = "https://queryapi-hasura-graphql-mainnet-vcqilefdcq-ew.a.run.app/v1/graphql";
```
darunrs, line 35: So I'm not sure what to do regarding where the URL for graphql should come from. Theoretically this is a env variable and dev and prod use different endpoints, if in the future we make calls. However, the use case for Coordinator or Block Streamer making any graphQL calls would be for some production use case. In which case these indexers should be hosted on the prod QueryApi. 

I'm thinking maybe they be made ENV variables anyway, but we jsut configure both dev and prod to point to prod Hasura? Let me know what you think.

block-streamer/src/test_utils.rs
```diff
@@ -173,3 +174,63 @@ pub fn get_streamer_message(block_height: u64) -> near_indexer_primitives::Strea
 
     near_indexer_primitives::StreamerMessage { block, shards }
 }
+
+pub fn utc_date_time_from_date_string(date: &str) -> chrono::DateTime<chrono::Utc> {
```
darunrs, line 178: This function is necessary for using the new bitmap processor, as the class is not mockable. As a result, at best you can mock the S3 client and GraphQL, which means the struct will always try to fetch the block date since it is passed a block height. So, I moved it to utils.

block-streamer/src/block_stream.rs
morgsmccauley: Yup, makes sense

block-streamer/src/main.rs
morgsmccauley: I think env var makes sense, to make updating easier. We can just have the same value for both like you've suggested.

morgsmccauley: Added `HASURA_GRAPHQL_ENDPOINT` in https://github.com/near/near-ops/pull/1793

morgsmccauley:
Updates:
- Did some minor clean up, mostly clippy lints
- Use graphql endpoint from env rather than hard code
- Rename `BitmapProcessor` to `ReceiverBlocksProcessor` to align with indexer name, adjusted file system to accommodate
- Merged `main` and resolved conflicts

PR #805, opened 2024-06-17
refactor: Move Data Layer provisioning from Runner to Coordinator

### LINKED ISSUE #760: Handle provisioning within Coordinator

morgsmccauley:
This PR updates Coordinator to handle Data Layer provisioning, removing the implicit step from Runner. Provisioning itself is still completed within Runner, but Coordinator will trigger and monitor it. Functionally, provisioning is the same, but there are some subtle differences around how it is handled:
- Provisioning will now happen as soon as the Indexer is registered, rather than when the first matched block is executed.
- Block Streams/Executors will not be started until provisioning is successful, neither will start when either pending or failed.

A `provisioned_state` enum has been added to the Indexer state within Redis. This is used to persist what stage of provisioning the Data Layer is at, as well as ensuring we only provision once.

## Concerns with current implementation
Overall, I'm not happy with the implementation here, but feel it is the best we can do given the current structure of Coordinator. As to not block this feature I decided to go forward with this approach, and will create a ticket to refactor/update later.

Provisioning is triggered within the "new" handler, and then polled within the "existing" handler, which seems a little awkward. The separated handling is necessary since no operation within the control loop (i.e. `Synchroniser`) should block, as that would block synchronisation for all other Indexers. So we need to trigger the provisioning initially, and then poll the completion each subsequent control loop.

I feel we have outgrown the current control loop, and am planning to refactor later. Rather than have a single control loop for _all_ Indexers, I'm thinking we can have dedicated loops for each of them. We could spawn a new task for each Indexer, which then manages its own lifecycle. Then each Indexer is free to wait for as long as it wants, without impacting other Indexers. This would allow us to handle the blocking provisioning step much more elegantly.


runner/tests/integration.test.ts
```diff
@@ -126,6 +126,8 @@ describe('Indexer integration', () => {
       }
     );
 
+    await provisioner.provisionUserApi(indexerConfig);
```
morgsmccauley, line 129: Need to provision manually since it's no longer called within `execute`

coordinator/src/indexer_state.rs
```diff
@@ -57,41 +55,36 @@ impl IndexerStateManagerImpl {
         Self { redis_client }
     }
 
-    pub async fn migrate(&self, registry: &IndexerRegistry) -> anyhow::Result<()> {
-        if self.redis_client.indexer_states_set_exists().await? {
-            return Ok(());
-        }
+    pub async fn migrate(&self) -> anyhow::Result<()> {
```
morgsmccauley, line 58: Migration to add `provisioned_state` enum, defaulting to `Provisioned`. There is some risk here if an Indexer hasn't actually been provisioned, but I believe it's unlikely and can handle manually if so.

PR #808, opened 2024-06-18
feat: Deprovision Data Layer on delete

### LINKED ISSUE #761: De-provision resources on Indexer deletion

morgsmccauley:
This PR removes Data Layer resources on Indexer Delete. To achieve this, the following has been added:
- `Provisioner.deprovision()` method which removes: schema, cron jobs, and if necessary, Hasura source, database, and role
- `DataLayerService.StartDeprovisioningTask` gRPC method
- Calling the above from Coordinator within the delete lifecycle hook

In addition to the above, I've slightly refactored DataLayerService to make the addition of de-provisioning more accomodating:
- `StartDeprovisioningTask` and `StartProvisioningTask` now return opaque IDs rather than using `accountId`/`functionName` - to avoid conflicts with eachother
- There is a single `GetTaskStatus` method which is used for both, before it was specific to provisioning

As mentioned in #805, the Coordinator implementation is a little awkward due to the shared/non-blocking Control Loop. I'll look to refactor this later and hopefully improve on this.

### DIFF
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -32,12 +32,16 @@ pub struct IndexerState {
     pub provisioned_state: ProvisionedState,
 }
 
+// FIX `IndexerConfig` does not exist after an Indexer is deleted, and we need a way to
+// construct the state key without it. But, this isn't ideal as we now have two places which
+// define this key - we need to consolidate these somehow.
 impl IndexerState {
-    // FIX `IndexerConfig` does not exist after an Indexer is deleted, and we need a way to
-    // construct the state key without it. But, this isn't ideal as we now have two places which
-    // define this key - we need to consolidate these somehow.
     pub fn get_state_key(&self) -> String {
         format!("{}/{}:state", self.account_id, self.function_name)
+    }
+
+    pub fn get_redis_stream_key(&self) -> String {
+        format!("{}/{}:block_stream", self.account_id, self.function_name)
     }
 }
 
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -11,6 +11,7 @@ pub enum ProvisionedState {
     Unprovisioned,
     Provisioning { task_id: String },
     Provisioned,
+    Deprovisioning { task_id: String },
     Failed,
 }
 
@@ -130,6 +131,22 @@ impl IndexerStateManagerImpl {
         indexer_state.block_stream_synced_at = Some(indexer_config.get_registry_version());
 
         self.set_state(indexer_config, indexer_state).await?;
+
+        Ok(())
+    }
+
+    pub async fn set_deprovisioning(
+        &self,
+        indexer_state: &IndexerState,
+        task_id: String,
+    ) -> anyhow::Result<()> {
+        let mut state = indexer_state.clone();
+
+        state.provisioned_state = ProvisionedState::Deprovisioning { task_id };
+
+        self.redis_client
+            .set(state.get_state_key(), serde_json::to_string(&state)?)
+            .await?;
 
         Ok(())
     }
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -9,7 +9,7 @@ use crate::redis::RedisClient;
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
 pub enum ProvisionedState {
     Unprovisioned,
-    Provisioning,
+    Provisioning { task_id: String },
     Provisioned,
     Failed,
 }
@@ -134,10 +134,14 @@ impl IndexerStateManagerImpl {
         Ok(())
     }
 
-    pub async fn set_provisioning(&self, indexer_config: &IndexerConfig) -> anyhow::Result<()> {
+    pub async fn set_provisioning(
+        &self,
+        indexer_config: &IndexerConfig,
+        task_id: String,
+    ) -> anyhow::Result<()> {
         let mut indexer_state = self.get_state(indexer_config).await?;
 
-        indexer_state.provisioned_state = ProvisionedState::Provisioning;
+        indexer_state.provisioned_state = ProvisionedState::Provisioning { task_id };
 
         self.set_state(indexer_config, indexer_state).await?;
 
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -110,6 +110,12 @@ impl IndexerStateManagerImpl {
             return Ok(serde_json::from_str(&raw_state)?);
         }
 
+        tracing::info!(
+            account_id = indexer_config.account_id.to_string(),
+            function_name = indexer_config.function_name.as_str(),
+            "Creating new state using default"
+        );
+
         Ok(self.get_default_state(indexer_config))
     }
 
--- a/coordinator/src/synchroniser.rs
+++ b/coordinator/src/synchroniser.rs
@@ -355,6 +355,8 @@ impl<'a> Synchroniser<'a> {
             return Ok(());
         }
 
+        self.redis_client.del(state.get_redis_stream_key()).await?;
+
         self.state_manager.delete_state(state).await?;
 
         Ok(())
@@ -1306,7 +1308,7 @@ mod test {
         }
 
         #[tokio::test]
-        async fn deprovisions_data_layer() {
+        async fn cleans_indexer_resources() {
             let config = IndexerConfig::default();
             let provisioned_state = IndexerState {
                 account_id: config.account_id.clone(),
@@ -1339,15 +1341,24 @@ mod test {
             let mut data_layer_handler = DataLayerHandler::default();
             data_layer_handler
                 .expect_start_deprovisioning_task()
-                .with(eq(config.account_id), eq(config.function_name))
+                .with(
+                    eq(config.clone().account_id),
+                    eq(config.clone().function_name),
+                )
                 .returning(|_, _| Ok("task_id".to_string()));
             data_layer_handler
                 .expect_get_task_status()
                 .with(eq("task_id".to_string()))
                 .returning(|_| Ok(TaskStatus::Complete));
 
+            let mut redis_client = RedisClient::default();
+            redis_client
+                .expect_del::<String>()
+                .with(eq(config.get_redis_stream_key()))
+                .returning(|_| Ok(()))
+                .once();
+
             let registry = Registry::default();
-            let redis_client = RedisClient::default();
             let block_streams_handler = BlockStreamsHandler::default();
             let executors_handler = ExecutorsHandler::default();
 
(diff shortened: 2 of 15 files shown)


coordinator/src/indexer_state.rs
```diff
@@ -57,41 +60,36 @@ impl IndexerStateManagerImpl {
...
+                    "{}/{} already migrated, skipping",
+                    state.account_id,
+                    state.function_name
+                );
+                continue;
+            }
+
+            tracing::info!("Migrating {}", raw_state);
+
+            let old_state: IndexerState = serde_json::from_str(&raw_state)?;
```
darunrs, line 78: What's the difference between the call on 67 and this one? If the first call fails, is it possible for the second call to succeed? Should it be `let old_state: OldIndexerState = serde_json::from_str(&raw_state)?;`?

coordinator/src/indexer_state.rs
```diff
@@ -101,6 +99,7 @@ impl IndexerStateManagerImpl {
             function_name: indexer_config.function_name.clone(),
             block_stream_synced_at: None,
             enabled: true,
+            provisioned_state: ProvisionedState::Unprovisioned,
```
darunrs, line 102: I'm wondering if we should log if default state is used. It could help us in scnearios where some obscure bug in state migration or issue sin Redis cause Coordinator to do the wrong thing.

coordinator/src/indexer_state.rs
```diff
@@ -196,72 +245,6 @@ mod tests {
         assert!(indexer_manager.list().await.is_err());
     }
 
-    #[tokio::test]
```
darunrs, line 199: Removed because we no longer expect to migrate?

coordinator/src/synchroniser.rs
```diff
@@ -226,6 +250,18 @@ impl<'a> Synchroniser<'a> {
         executor: Option<&ExecutorInfo>,
         block_stream: Option<&StreamInfo>,
     ) -> anyhow::Result<()> {
+        match &state.provisioned_state {
+            ProvisionedState::Provisioning { task_id } => {
+                self.ensure_provisioned(config, task_id.clone()).await?;
+                return Ok(());
+            }
+            ProvisionedState::Failed => return Ok(()),
```
darunrs, line 258: Is this return Ok() to just not do anything? I figure we'd spot something when we get an alarm when Executed or Processed block count falls to 0, or whatever our error metric evolves to. 
Where should retry logic exist? I'm thinking maybe each service can have a "restart" API which coordinator can call to perform a restart with some given configuration. This way, Coordinator is only responsible for starting the process, whereas the service itself owns the business logic.

coordinator/src/synchroniser.rs
```diff
@@ -226,6 +250,18 @@ impl<'a> Synchroniser<'a> {
...
+            ProvisionedState::Provisioned => {}
+            ProvisionedState::Unprovisioned | ProvisionedState::Deprovisioning { .. } => {
+                anyhow::bail!("Provisioning task should have been started")
+            }
+        }
+
         if !state.enabled {
```
darunrs, line 265: DO you think it makes sense to break these out into small functions? The syncing logic seems complex so it might be nice to have the main functions be understandable at a glace and functions encapsulate the underlying logic. I think it would look nice to see a sync_exisitng_indexer function which looks like:
```
verify_provisioned()
check_enabled()
sync_block_stream()
sync_executor()
sync_data_layer()
```

or something like that. I think a lot of those are shared between each sync types (new, existing, deleted, etc.).

coordinator/src/synchroniser.rs
```diff
@@ -211,6 +206,35 @@ impl<'a> Synchroniser<'a> {
         Ok(())
     }
 
+    async fn ensure_provisioned(
```
darunrs, line 209: How is this function related to checking the state of the indexer itself. It seems this function is focused on ensuring a related provisioning task exists. If this function also checked state, we could ignore context right? When I look at the function as it is, I wonder if this is before or after a state check. Then again, separating logic is good too. I think it would be good to establish an order of precedence for a data layer task and the corresponding indexer state value, and keep the checks collected under one function, even if its like below. What do you think? 

```
check_data_layer_state() (returns if fine)
check_data_layer_task_status() (calls if above requires checking of task)
```

coordinator/src/synchroniser.rs
```diff
@@ -289,6 +325,38 @@ impl<'a> Synchroniser<'a> {
...
+            match self
+                .data_layer_handler
+                .get_task_status(task_id.clone())
+                .await?
+            {
+                TaskStatus::Complete => {
+                    tracing::info!("Data layer deprovisioning complete");
+                }
+                TaskStatus::Failed => {
+                    tracing::info!("Data layer deprovisioning failed");
```
darunrs, line 338: This is an error log right? Would this be repeated each loop? Perhaps we can simply let the error log be the single error log instance, and return Ok() here?

runner-client/examples/check_provisioning_task_status.rs
```diff
@@ -0,0 +1,20 @@
...
+use runner::data_layer::CheckProvisioningTaskStatusRequest;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut client = DataLayerClient::connect("http://localhost:7001").await?;
+
+    let response = client
+        .check_provisioning_task_status(Request::new(CheckProvisioningTaskStatusRequest {
+            account_id: "morgs.near".to_string(),
+            function_name: "test2".to_string(),
```
darunrs, line 13: Given they're examples, maybe we can simply call them something like "data_layer_example" and have that be the function name for any data layer examples? And a similar thing for other types of examples: "executor_example", "block_stream_example".

runner/src/provisioner/provisioner.test.ts
```diff
@@ -71,6 +72,108 @@ describe('Provisioner', () => {
     indexerConfig = new IndexerConfig('', accountId, functionName, 0, '', databaseSchema, LogLevel.INFO);
   });
 
+  describe('deprovision', () => {
+    it('removes schema level resources', async () => {
+      userPgClientQuery = jest.fn()
+        .mockResolvedValueOnce(null) // drop schema
+        .mockResolvedValueOnce(null) // unschedule create partition job
+        .mockResolvedValueOnce(null) // unschedule delete partition job
+        .mockResolvedValueOnce({ rows: [{ schema_name: 'another_one' }] }); // list schemas
```
darunrs, line 81: Considering its possible for a cron job to silently fail sometimes, I think we should also verify the cron jobs are indeed deleted, before fully accepting a success. A failing cron job doesn't harm the DB but it does add noise to the cron job status and history table, which would make debugging harder, if we ever need to do so.

darunrs (approved):
Looks mostly good to me! Obviously this is a large impact change. Have you done any local tests of running all four services locally? It also seems like you've got a somewhat large merge conflict to fix haha...

coordinator/src/indexer_state.rs
```diff
@@ -2,14 +2,23 @@
...
-use crate::registry::IndexerRegistry;
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
+pub enum ProvisionedState {
+    Unprovisioned,
+    Provisioning { task_id: String },
+    Provisioned,
+    Deprovisioning { task_id: String },
+    Failed,
```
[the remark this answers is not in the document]
morgsmccauley: No - but at this point it isn't my goal, I will address redundancy in future work.

coordinator/src/indexer_state.rs
morgsmccauley: Yes you're correct, this is a bug, fixed :)

coordinator/src/indexer_state.rs
morgsmccauley: Yeah good point, I'll do that!

coordinator/src/indexer_state.rs
morgsmccauley: Removed because it was for the old migration, i've added a new one for the latest migration

coordinator/src/synchroniser.rs
morgsmccauley: For now, it's probably fine. This means we don't start the block stream or executor, which makes sense because it won't work anyway.

Restart/retry will be tricky, we need to expose more information from the `DataLayerService`, to know if we actually _can_ retry. If it's a schema error there is no point.

From the user side, they'll just see the Indexer hanging on "PROVISIONING" status, but I'm going to add some more logging so they can see the error.

coordinator/src/synchroniser.rs
morgsmccauley: Yeah, we should. The whole thing is getting a bit messy tbh, I'm planning a [refactor](https://github.com/near/queryapi/issues/811) soon, I'll clean this up then.

coordinator/src/synchroniser.rs
morgsmccauley: It's not quite an error log, provisioning may fail because there schema is messed up, which isn't necessarily a system error. Maybe a warn? 🤔

runner-client/examples/check_provisioning_task_status.rs
morgsmccauley: Can do 👍🏼

runner/src/provisioner/provisioner.test.ts
morgsmccauley: Good point. I feel it makes more sense to assert this behaviour in tests, rather than creating additional code for it? I'll update the integration tests to ensure things are being removed.

coordinator/src/synchroniser.rs
morgsmccauley: I'll so it now since I'm not sure when the refactor will happen :)

morgsmccauley: Ah, I remember why I didn't do this. I need some of these inline to have more control over the control flow of the function, i.e. I want to short-circuit synchronisation if it isn't provisioned yet. I can't do that from another function unless I add similar code to check and return.

I'll have a deeper think about this going forward, and see how I can tidy things up.

coordinator/src/synchroniser.rs
morgsmccauley: Sorry, I'm not quite sure I follow. Can you please explain again?

The purpose of this function is to poll the pending Data Layer Provisioning task, and update the Indexer State when complete (or failed). At which point we will move to the next "phase" of the Indexer, starting block streams/executors.

I think part of the confusion here comes from the shared control loop, which I outline here: https://github.com/near/queryapi/issues/811

runner/src/provisioner/provisioner.test.ts
morgsmccauley: Wait, what do you mean cron jobs silently fail? Removal silently fails?

morgsmccauley:
@darunrs merging now, but let's keep discussing and I can make updates in future PRs

PR #809, opened 2024-06-18
feat: In-Memory DmlHandler Test Fixture

darunrs:
In order to enable quick testing of indexer code for the purposes of local indexer development, there needs to be a way to functionally mock dependencies that are used by the Indexer code. These dependencies are roughly encapsulated under the context object. In particular context.db is of high importance as it is the method through with Indexers interact with their persistent data. This PR focuses on creating an MVP of an in memory DML Handler Test Fixture which can be used interchangeably with an actual DmlHandler. This allows context.db calls to pretend to be real calls to an actual Postgres DB without actually having an instance running (Which would be hard to integrate into unit testing). 

This PR is mainly focused on getting a basic DmlHandler test fixture out the door. It has the same functions which roughly behave similarly to if they were called on an actual DB. There is the added benefit of being extremely quick to tear down as all data is represented in memory, making it suitable for fast iteration through unit testing. However, since I am essentially mocking Postgres DB responses, the correctness standard is much lower than using an actual PG DB, but since DmlHandler simplifies user interactions with the DB anyway, we can mock a great deal of Indexer context.db use cases sufficiently to serve as useful for end users.

### DIFF
--- a/runner/src/dml-handler/dml-handler-fixture.test.ts
+++ b/runner/src/dml-handler/dml-handler-fixture.test.ts
@@ -70,6 +70,21 @@ describe('DML Handler Fixture Tests', () => {
     expect(result).toEqual(correctResult);
   });
 
+  test('get error from inserting conflicting row', async () => {
+    const inputObj = [{
+      account_id: 'morgs_near',
+      block_height: 1,
+      receipt_id: 'abc',
+      content: "some content",
+      block_timestamp: 123,
+      accounts_liked: [],
+      last_comment_timestamp: 456
+    }];
+
+    await dmlHandler.insert(TABLE_DEFINITION_NAMES, inputObj);
+    expect(dmlHandler.insert(TABLE_DEFINITION_NAMES, inputObj)).rejects.toThrow('Cannot insert row twice into the same table');
+  });
+
   test('select rows', async () => {
     const inputObj = [{
       account_id: 'morgs_near',
--- a/runner/src/dml-handler/dml-handler-fixture.test.ts
+++ b/runner/src/dml-handler/dml-handler-fixture.test.ts
@@ -100,5 +100,38 @@ describe('DML Handler Fixture Tests', () => {
     expect(selectB[1].account_id).toEqual('morgs_near');
     expect(selectB[0].block_height).toEqual(1);
     expect(selectB[1].block_height).toEqual(2);
+
+    expect(await dmlHandler.select(TABLE_DEFINITION_NAMES, { account_id: 'unknown_near' })).toEqual([]);
   });
+
+  test('update rows', async () => {
+    const inputObj = [{
+      account_id: 'morgs_near',
+      block_height: 1,
+      receipt_id: 'abc',
+      content: "some content",
+      block_timestamp: 123,
+      accounts_liked: [],
+      last_comment_timestamp: 456
+    },
+    {
+      account_id: 'morgs_near',
+      block_height: 2,
+      receipt_id: 'abc',
+      content: "some content",
+      block_timestamp: 123,
+      accounts_liked: [],
+      last_comment_timestamp: 456
+    }];
+
+    await dmlHandler.insert(TABLE_DEFINITION_NAMES, inputObj);
+
+    const updateOne = await dmlHandler.update(TABLE_DEFINITION_NAMES, { account_id: 'morgs_near', block_height: 2 }, { content: 'updated content' });
+    const selectOneUpdate = await dmlHandler.select(TABLE_DEFINITION_NAMES, { account_id: 'morgs_near', block_height: 2 });
+    expect(updateOne).toEqual(selectOneUpdate);
+
+    const updateAll = await dmlHandler.update(TABLE_DEFINITION_NAMES, { account_id: 'morgs_near' }, { content: 'final content' });
+    const selectAllUpdated = await dmlHandler.select(TABLE_DEFINITION_NAMES, { account_id: 'morgs_near' });
+    expect(updateAll).toEqual(selectAllUpdated);
+  })
 });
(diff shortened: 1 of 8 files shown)


runner/src/dml-handler/dml-handler-fixture.ts
```diff
@@ -0,0 +1,193 @@
...
+        results.push(existingRow.data);
+        if (limit && results.length >= limit) {
+          return results;
+        }
+      }
+    }
+    return results;
+  }
+
+  insertRow(tableName: string, row: any): DataRow {
```
pkudinov, line 138: Instead of any, it is better to introduce something like `type RowType = Record<string, string|number>`.

darunrs: Yep that's a good idea. I will do so. The only thing is that it might possible for the type to be things other than string or number. Such as Buffer or Date. But I think the return from the actual Postgres query would serialize the result to a string anyway? It needs more testing, but string and number should sufficiently cover use cases for the time being.

darunrs: Actually a quick test on the receivers indexer shows we insert a string type into the `date` type and that works. So, I actually need to update the type gen a bit here to allow string or Date type. I just need to later test if inserting a Date type even works at all. But at least I know string works.

runner/src/dml-handler/dml-handler-fixture.test.ts
```diff
@@ -0,0 +1,204 @@
+import { TableDefinitionNames } from "../indexer";
+import InMemoryDmlHandler from "./dml-handler-fixture";
+
+const DEFAULT_ITEM_1_WITHOUT_ID = {
```
darunrs, line 4: I may later on try to make some kind of builder function for creating these objects, but for now a couple static objects suffice for most unit test cases.

runner/src/dml-handler/dml-handler-fixture.ts
```diff
@@ -0,0 +1,357 @@
...
+  serialColumns: string[]
+}
+
+class PostgresRowEntity {
+  data: PostgresRow;
+  private primaryKeys: string[];
+
+  constructor(data: any, primaryKeys: string[]) {
+    this.data = data;
+    this.primaryKeys = primaryKeys.sort();
```
darunrs, line 20: I sort the primary keys to ensure that the below primaryKey() function can reliably stringify the primary key attribute values and compare against other object's calls of it without worrying about order. 

I did think about maybe comparing attribute values directly too though.

runner/src/dml-handler/dml-handler-fixture.ts
```diff
@@ -0,0 +1,357 @@
...
+        }
+      }
+    });
+    return matchedRows;
+  }
+
+  private getSerialValue(columnName: string): number {
+    const serialCounterKey = `${this.specification.tableName}-${columnName}`;
+    let counterValue: number = this.serialCounter.get(serialCounterKey) ?? 0;
+    this.serialCounter.set(serialCounterKey, counterValue + 1);
```
darunrs, line 92: Auto-increments each time this function is called.

runner/src/dml-handler/dml-handler-fixture.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  public insert(tableName: string, rowsToInsert: PostgresRow[]): PostgresRow[] {
+    // TODO: Check types of columns
+    // TODO: Verify columns are correctly named, and have any required values
+    // TODO: Verify inserts are unique before actual insertion
+    const tableData = this.getTableData(tableName);
+    const insertedRows: PostgresRowEntity[] = [];
+
+    for (const row of rowsToInsert) {
+      const rowCopy = this.copyRow(row);
```
darunrs, line 261: Since JS objects are passed by reference, I need to copy the data to ensure that input or output data don't impact stored table data. 

I chose to do this copying as part of the interface DML functions rather than more closely to functions like insertRow to hopefully make it more clear that this copying is being performed. Open to suggesitons on any other ideas for where this copying should be done. 

I went with parsing a stringify to do a deep copy, which should work for nearly all cases.

runner/src/dml-handler/dml-handler-fixture.ts
```diff
@@ -0,0 +1,360 @@
...
+  columnNames: string[]
+  primaryKeyColumns: string[]
+  serialColumns: string[]
+}
+
+class PostgresRowEntity {
+  data: PostgresRow;
+  private primaryKeys: string[];
+
+  constructor(data: any, primaryKeys: string[]) {
```
morgsmccauley, line 18:
```suggestion
  constructor(data: Record<string, any>, primaryKeys: string[]) {
```
Would this be more correct?

runner/src/dml-handler/dml-handler.ts
```diff
@@ -6,15 +6,24 @@ import type IndexerConfig from '../indexer-config/indexer-config';
...
-type WhereClauseMulti = Record<string, (string | number | Array<string | number>)>;
-type WhereClauseSingle = Record<string, (string | number)>;
-
-export default class DmlHandler {
+export type PostgresRowValue = string | number | any;
+export type PostgresRow = Record<string, PostgresRowValue>;
+export type WhereClauseMulti = Record<string, (PostgresRowValue | Array<PostgresRowValue>)>;
+export type WhereClauseSingle = Record<string, PostgresRowValue>;
+
+export interface DmlHandlerI {
```
morgsmccauley, line 14:
```suggestion
export interface IDmlHandler {
```
I think this is more conventional

morgsmccauley (approved):
Really minor comments, overall looks good :)

runner/src/dml-handler/dml-handler-fixture.ts
darunrs: Oops! You're correct. I even defined an exlicit type to represent it but I must have forgotten to put it here during the refactoring!

PR #817, opened 2024-06-19
fix: Ensure state is updated after starting data layer provisioning

morgsmccauley:
- Need to write the state to move it out of "new" and in to "existing"
- Added some logs to `DataLayerService`

### DIFF
--- a/runner/src/server/services/data-layer/data-layer-service.ts
+++ b/runner/src/server/services/data-layer/data-layer-service.ts
@@ -2,6 +2,7 @@ import { type ServerUnaryCall, type sendUnaryData, status, StatusBuilder } from 
 
 import Provisioner from '../../../provisioner';
 import { ProvisioningConfig } from '../../../indexer-config/indexer-config';
+import parentLogger from '../../../logger';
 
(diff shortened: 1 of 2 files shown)


darunrs (approved):

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -91,7 +98,19 @@ export function createDataLayerService (
           return;
         }
 
-        tasks[generateTaskId(accountId, functionName)] = new ProvisioningTask(provisioner.provisionUserApi(provisioningConfig));
+        logger.info('Provisioning Data Layer');
```
darunrs, line 101: The word "provision" is hanging me up a tiny bit. This makes me indicate there's some persistent state that is created one time, and never again. Is that a correct assumption?

morgsmccauley: - "Data Layer" is what I'm referring to as Hasura + Postgres, rather than the generic term we have been using "Provisioning"
- "Provisioning" I'm not using as the term for creating the above
- "DataLayerService" is the code which manages the Data Layer. I see now that this could be confusing.

Does that make sense? "Indexer Resources" is quite generic, and may make sense at the Control Plane level, but within the implementation I think it makes sense to be more concrete.

PR #823, opened 2024-06-20
fix: added polling to latest indexer block height

### LINKED ISSUE #820: Block lag shows incorrect values during live update
The live updates to the Block Lag (`Indexer is X blocks or Y behind the blockchain tip`) incorrectly grows in value, even when the actual block lag is reducing. On fresh load, and refresh, the value appears to be correct, but it becomes incorrect after the live updates.

image

Kevin101Zhang:
props.indexerBlockHeight is static and fetches from GQL. added functionality to update the current Indexer blockheight.

### DIFF
--- a/frontend/src/components/Logs/LogsMenu.tsx
+++ b/frontend/src/components/Logs/LogsMenu.tsx
@@ -40,6 +40,7 @@ const LogsMenu: React.FC<LogsMenuProps> = ({
   `;
 
   const { loading, error, data, refetch } = useQuery(GET_METADATA, {
+    pollInterval: 1000,
     context: {
       headers: {
         'x-hasura-role': hasuraRole,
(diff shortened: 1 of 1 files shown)


frontend/src/components/Logs/LogsMenu.tsx
```diff
@@ -67,6 +67,37 @@ const LogsMenu: React.FC<LogsMenuProps> = ({
     }
   }, [attributeMap]);
 
+  useEffect(() => {
```
morgsmccauley, line 70: This is basically the same as the two `useEffect`s above. Why don't we just set an interval to call `refetch()`.  `useQuery()` would then return the latest `data`, and the above `useEffect`s would then trigger and update the status/block height.

But this functionality is already built in to `useQuery`, so let's just remove set a [polling interval](https://www.apollographql.com/docs/react/data/queries#polling), and then just rely on the existing `useEffect`s

morgsmccauley:
We should be using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for our PR titles, can you please update yours.

morgsmccauley (approved):

PR #833, opened 2024-06-25
fix: Remove role restriction for removing functions

### LINKED ISSUE #822: Unable to delete Indexer after publishing
We've removed the whitelist restrictions for publishing Indexers, but they still exist for deletion. The restrictions should also be removed from delete.

image

morgsmccauley:
Also runs `cargo format` (1st commit)

closes: #822

### DIFF
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -329,10 +329,7 @@ impl Contract {
                     env::panic_str(&format!("Account ID {} is invalid", account_id));
                 })
             }
-            None => {
-                self.assert_roles(vec![Role::Owner, Role::User]);
-                env::signer_account_id()
-            }
+            None => env::signer_account_id(),
         };
 
         log!(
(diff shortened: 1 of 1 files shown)


morgsmccauley:
Deployed to both dev/prod contracts

registry/contract/src/lib.rs
```diff
@@ -329,10 +329,7 @@ impl Contract {
                     env::panic_str(&format!("Account ID {} is invalid", account_id));
                 })
             }
-            None => {
-                self.assert_roles(vec![Role::Owner, Role::User]);
-                env::signer_account_id()
-            }
+            None => env::signer_account_id(),
```
gabehamilton, line 332: Would also be useful if owner's could remove any indexer; large suspended indexers for instance.

morgsmccauley: They can - if `account_id` is specified, we'll only permit the tx if the signer is `Owner`

PR #830, opened 2024-06-24
refactor: Refactor Editor into TypeScript with Smaller Components and Separate Concerns

Kevin101Zhang:
Refactored Editor Component to TypeScript. This refactoring involved breaking down the Editor file into smaller chunks and separating concerns into distinct components. Also did some minor work on validator to ts as it is a major consumer in the editor. It is setup to later iterate on some additional test for validators.

PR #835, opened 2024-06-25
feat: line position now persist for indexer.js when switching tabs

### LINKED ISSUE #757: Persist line/cursor position of indexer.js when moving between different file tabs

Kevin101Zhang:
Added logic to monaco lifecycle method and react lifecycle methods to persist line/cursor position between the swapped files.


morgsmccauley (commented):
Reviewed the last 2 commits as it looks like this still contains commits from #830. Looks good, but can you remove the duplicate commits please?

PR #855, opened 2024-07-08
feat: Enhance Schema Error Visualization with New Icons

### LINKED ISSUE #847: Monaco Editor Alert Jumps to top

Kevin101Zhang:
Optimized Editor Lifecycle methods and removed the 'janky' Alert that would insert itself between the developer tools and Editor component in favor of a small icon that is in the fileswitcher.


frontend/widgets/src/QueryApi.Dashboard.jsx
```diff
@@ -17,30 +17,6 @@ State.init({
   selected_account: undefined,
 });
 
-Near.asyncView(`${REPL_REGISTRY_CONTRACT_ID}`, "list_all").then((data) => {
```
Kevin101Zhang, line 20: Cause of multiple rerendering

Kevin101Zhang:
Schema now no longer throws an Alert component in between the Editor and the DevelopMenu cause a 'janky experience'. Instead schema not showcases errors based on the file tab itself through small icons. Similarly the debugMode no longer has a Alert component but is now directs the user to open their console through a onHover tooltip. Indexer.js has a similar funtionality but is locked through a feature flag as it is dependent on Monaco having types of node, near-lake first before we can properly prompt the user there is an error in the Indexer.

morgsmccauley:
There's a lot going on in this PR, could you please split it in to smaller ones, one for each of the `fix`/`feat` added in the title?

Smaller PRs are preferable because they:
1. Make the review process easier, smaller diff to review.
2. Reduce the likelihood of introducing bugs, as changes are more focused and easier to understand.
5. Improve the ability to identify the source of issues, as smaller, isolated changes make debugging more straightforward.

frontend/src/utils/validators.ts
Kevin101Zhang, line 1: Small change in validators to make the error handling more consistent with others in the same file

frontend/src/pages/query-api-editor/index.js
Kevin101Zhang, line 1: small change in indexer, no functionality change just made the component more readable.

Kevin101Zhang:
@morgsmccauley I removed the "fix: multiple rerenders" to a different PR. 
The Editor optimizations heavily revolve around the the introduction of the new feature of icons in fileSwitcher to showcase error messages. All the other files here are tightly coupled are children component of Editor (fileSwitcher, developertools) are the larger number of file changes are due to view and view Container. 

The exception are the 2 small more readability refactors in the files commented above that are trivial.

frontend/src/components/Editor/EditorComponents/FileSwitcher.jsx
```diff
@@ -1,25 +1,48 @@
...
       <button
         className={`flex-1 px-4 py-2 text-base text-xs font-medium ${
           fileName === 'indexer.js' ? 'bg-gray-700 text-gray-100' : 'bg-gray-300 text-gray-700 hover:bg-gray-400'
         } border-r border-gray-400 last:border-r-0`}
         onClick={() => setFileName('indexer.js')}
       >
-        Indexer.js
+        <span>Indexer.js</span>
+        <div className="inline-flex items-center">
+          {/* {indexerError ? (
```
morgsmccauley, line 21: should we just remove this?

frontend/src/utils/validators.ts
```diff
@@ -47,13 +47,11 @@ export function validateJSCode(code: string): { data: string | null; error: Erro
   try {
     const formattedCode = formatIndexingCode(code);
     return { data: formattedCode, error: null };
-  } catch (error: unknown) {
-    if (error instanceof Error) {
-      console.error(error.message);
-      return { data: code, error };
-    } else {
-      throw error;
```
morgsmccauley, line 55: I don't understand, there is a functional change here. Previously, if the error was not an `Error` it would be re-thrown. Are you sure we don't need that functionality?

frontend/src/pages/query-api-editor/index.js
```diff
@@ -11,34 +11,20 @@ const QueryApiEditorPage = ({ router }) => {
...
-      return;
-    }
+    if (!accountId || !indexerName) return;
     setAccountId(accountId);
     setIndexerName(indexerName);
-  }, [accountId, indexerName, setAccountId, setIndexerName]);
+  }, [accountId, indexerName]);
 
-  if (accountId == undefined || indexerName == undefined) {
+  if (!accountId || !indexerName) {
```
morgsmccauley, line 19: I agree this is more readable but it isn't exactly the same. This condition will be `true` for _any_ falsy value, so `null`, `false`, `''`, etc. Where as the previous would _only_ be `false` for `undefined`

morgsmccauley (approved):
Looks good, just a few minor comments

frontend/src/pages/query-api-editor/index.js
Kevin101Zhang: that makes sense

frontend/src/components/Editor/EditorComponents/FileSwitcher.jsx
Kevin101Zhang: I want to add this the second I get monaco Types inserted. I attempted it with a little time but I was having trouble. Will retackle it in the next ticket so I would like to keep it

frontend/src/utils/validators.ts
Kevin101Zhang: Yes! Im fairly certain the error will be handled properly.


```
const debouncedValidateCode = useDebouncedCallback((_code: string) => {
  const { error } = validateJSCode(_code);
  console.log(error);
  indexerErrorHandler(error);
}, 500);

export function validateJSCode(code: string): { data: string | null; error: Error | null } {
  if (!code) return { data: null, error: null };

  try {
    const formattedCode = formatIndexingCode(code);
    return { data: formattedCode, error: null };
  } catch (error: any) {
    return {
      data: code,
      error: new Error(`Validation error: ${error.message}`), // Improved error message
    };
  }
}

export const formatIndexingCode = (code) => {
  try {
    return prettier.format(code, {
      parser: 'babel',
      plugins: [parserBabel],
    });
  } catch (error) {
    throw new Error(`Failed to format code: ${error.message}`);
  }
};
```

PR #856, opened 2024-07-09
test: Cancel block_stream task to workaround infinite lake loop

morgsmccauley:
`near-lake-framework` will infinitely retry, making our tests infinitely hang. This PR monitors the `last_processed_ block`, and cancels the task from the outside when the last desired block is reached.

### DIFF
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -346,11 +346,9 @@ mod tests {
 
     use mockall::predicate;
     use near_lake_framework::s3_client::GetObjectBytesError;
+    use std::sync::atomic::{AtomicU64, Ordering};
     use std::sync::Arc;
 
-    // FIX: near lake framework now infinitely retires - we need a way to stop it to allow the test
-    // to finish
-    #[ignore]
     #[tokio::test]
     async fn adds_matching_blocks_from_bitmap_and_lake() {
         let contract_filter = "queryapi.dataplatform.near";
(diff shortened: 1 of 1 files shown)


block-streamer/src/block_stream.rs
```diff
@@ -479,7 +493,7 @@ mod tests {
 
         mock_lake_s3_client
             .expect_list_common_prefixes()
-            .returning(|_, _| Ok(vec![]));
+            .returning(|_, _| Ok(vec![107503704.to_string(), 107503705.to_string()]));
```
darunrs, line 496: What caused this to change?

block-streamer/src/block_stream.rs
```diff
@@ -448,23 +452,33 @@ mod tests {
...
+            .start(
+                91940840,
+                std::sync::Arc::new(mock_redis),
+                std::sync::Arc::new(mock_reciever_blocks_processor),
+                mock_lake_s3_client,
+            )
+            .unwrap();
+
+        loop {
+            if last_processed_block_height.load(Ordering::Relaxed) == 107503705 {
```
darunrs, line 472: Should use >= instead just to be absolutely sure we stop?

darunrs (approved):
Smart way of solving this issue!

block-streamer/src/block_stream.rs
morgsmccauley: The test setup wasn't correct, I think they probably got edited while they were `ignored`

block-streamer/src/block_stream.rs
morgsmccauley: Good point!

PR #863, opened 2024-07-10
fix: Filter receiver blocks before starting block

morgsmccauley:
As we fetch blocks for a given day, there is possibility for that day to include blocks lesser than the provided `start_block_height`. This PR ensures that those blocks are filtered out.

### DIFF
--- a/block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
+++ b/block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
@@ -187,7 +187,7 @@ impl ReceiverBlocksProcessor {
                 }
 
                 for block_height in bitmap_for_day.iter() {
-                    if block_height > start_block_height {
+                    if block_height >= start_block_height {
                         yield block_height;
                     }
                 }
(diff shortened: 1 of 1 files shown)


block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
```diff
@@ -187,7 +187,9 @@ impl ReceiverBlocksProcessor {
                 }
 
                 for block_height in bitmap_for_day.iter() {
-                    yield block_height;
+                    if block_height > start_block_height {
```
darunrs, line 190: Our [previous](https://github.com/near/queryapi/blob/adaadfc3b2c94ad7bfdf54cd7d9d887b0b55899a/block-streamer/src/delta_lake_client.rs#L250) implementation had `>=`. What's the correct behavior here? It seems to me like >= is what's right.

block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
```diff
@@ -388,11 +416,10 @@ mod tests {
...
             .stream_matching_block_heights(0, "*.someone.near".to_string());
-        tokio::pin!(stream);
-        let mut result_heights = vec![];
-        while let Some(Ok(height)) = stream.next().await {
-            result_heights.push(height);
-        }
-        assert_eq!(result_heights, vec![1, 5, 10, 15, 100, 105]);
+
+        assert_eq!(
+            stream.try_collect::<Vec<u64>>().await.unwrap(),
```
darunrs, line 421: This is very clean!

darunrs (approved):
Great catch! Thanks for spotting and fixing it!

block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
morgsmccauley: Ah yeah, `>=` is correct 😅

PR #864, opened 2024-07-11
fix: Support array values for queries and string type for inserts

darunrs:
Context DB in the frontend was showing an error for inserts and select. For inserts, some types such as `date` would expect `Date` type when string type also works. I updated the type gen to allow string type input for these options. This ensures string or number input is acceptable for essentially all types. 
image

In addition, selection queries allow for array input to match muliple values for a certain column. However, the frontend was not updated to accept this input. I've added a new interface called Query which allows for array values in input objects.

### DIFF
--- a/frontend/src/utils/pgSchemaTypeGen.js
+++ b/frontend/src/utils/pgSchemaTypeGen.js
@@ -273,8 +273,8 @@ export class PgSchemaTypeGen {
       // UUID type
       uuid: 'string',
       // JSON types
-      json: 'any',
-      jsonb: 'any',
+      json: 'string | any',
+      jsonb: 'string | any',
       // Arrays
       'integer[]': 'number[]',
       'text[]': 'string[]',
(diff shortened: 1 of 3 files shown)


frontend/src/utils/pgSchemaTypeGen.js
```diff
@@ -204,30 +204,33 @@ export class PgSchemaTypeGen {
 
       tableList.add(sanitizedTableName);
 
+      let queryDefinition = `declare interface ${sanitizedTableName}Query {\n`;
```
darunrs, line 207: I will actually get rid of this file relatively soon in favor of a new class in Runner which handles Postgres DLL Parsing logic. Part of its functionality will be producing the TS file string. This will prevent issues like this in the future where backend features get supported while the frontend acts as though it isn't supported.

morgsmccauley (approved):

PR #870, opened 2024-07-16
feat!: Return Option rather than panic

morgsmccauley:
Option is more convenient as we don't need to manually inspect the error to determine whether or not the Indexer exists. This is a breaking change, but there aren't any consumers of this so it is safe.


morgsmccauley:
Deployed to `dev-queryapi.dataplatform.near`

darunrs (approved):

morgsmccauley:
Deployed to `queryapi.dataplatform.near`

PR #873, opened 2024-07-16
fix: fetch BOS indexerdata from near RPC call and conditionally render metadata if available in databricks

### LINKED ISSUE #872: Augment explore indexers page with the number of requests in the past 7 days

Kevin101Zhang:
BOS component data is fetched from NEAR RPC call 'list_all'. We use a map to set a k/v pair for indexer metadata recieved on databricks. When we render each card we check the map to see if the indexer to be rendered from NEAR RPC exist in the map. If it does exist we conditionally render metadata.

### DIFF
--- a/frontend/widgets/src/QueryApi.IndexerCard.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerCard.jsx
@@ -1,9 +1,7 @@
 const { accountId, indexerName, lastDeploymentDate, numDeployements, numQueries, originalDeploymentDate } = props;
-const editUrl = `https://dev.near.org/${REPL_ACCOUNT_ID}/widget/QueryApi.App?selectedIndexerPath=${accountId}/${indexerName}`;
-const statusUrl = `https://dev.near.org/${REPL_ACCOUNT_ID}/widget/QueryApi.App?selectedIndexerPath=${accountId}/${indexerName}&view=indexer&activeIndexerView=status`;
+const editUrl = `https://dev.near.org/${REPL_ACCOUNT_ID}/widget/QueryApi.App?selectedIndexerPath=${accountId.replaceAll("_", ".")}/${indexerName}`;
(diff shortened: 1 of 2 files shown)


frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -1,12 +1,31 @@
...
+const myAccountId = context.accountId;
 
 const [selectedTab, setSelectedTab] = useState(props.tab && props.tab !== "all" ? props.tab : "all");
 const [myIndexers, setMyIndexers] = useState([]);
 const [allIndexers, setAllIndexers] = useState([]);
 const [error, setError] = useState(null);
+const [indexerMetadata, setIndexerMetaData] = useState(new Map());
 
 const fetchIndexerData = () => {
+  Near.asyncView(`${REPL_REGISTRY_CONTRACT_ID}`, "list_all").then((data) => {
```
Kevin101Zhang, line 10: To be replaced with indexer

frontend/widgets/src/QueryApi.IndexerCard.jsx
```diff
@@ -175,11 +189,6 @@ return (
       <ButtonLink
         primary
         href={editUrl}
-        onClick={() =>
```
darunrs, line 178: Is this not necessary anymore? I think we use `active tab` in various places?

darunrs (approved):
LGTM! To confirm my understanding, the new code now does:
1. Fetch active indexers using RPC
2. Fetch usage data for all indexers from some endpoint
3. Pass active indexer list to some component as cards
4. For each card, look at the map of usage data and populate usage if present

Did I get that right?

frontend/widgets/src/QueryApi.IndexerCard.jsx
Kevin101Zhang: the state did not exist on the page. I assumed maybe it was a hidden property passed from a parent but i didnt see anything of the sort. Removing it I did not see any differences.

PR #889, opened 2024-07-17
feat: Expose health info from Block Stream/Executor RPC

### LINKED ISSUE #875: Restart unhealthy block streams and executors

morgsmccauley:
This PR exposes a `health` field on the Block Stream/Executor info, which can be accessed via RPC. The intent of this field is for Coordinator to monitor it, and then act accordingly. I wanted to raise this work first, so that the overall result is not too large.

Essentially, `health` contains only a single `enum` describing the "state" of the process, but this can be expanded over time as needed.

### DIFF
--- a/block-streamer/proto/block_streamer.proto
+++ b/block-streamer/proto/block_streamer.proto
@@ -108,4 +108,18 @@ message StreamInfo {
     string function_name = 4;
     // Block height corresponding to the created/updated height of the indexer
     uint64 version = 5;
+    // Health
+    Health health = 6;
+}
+
+message Health {
+    ProcessingState processing_state = 1;
+}
+
+enum ProcessingState {
+    UNSPECIFIED = 0;
+    IDLE = 1;
+    ACTIVE = 2;
+    PAUSED = 3;
+    HALTED = 4;
 }
--- a/block-streamer/proto/block_streamer.proto
+++ b/block-streamer/proto/block_streamer.proto
@@ -114,6 +114,7 @@ message StreamInfo {
 
 message Health {
     ProcessingState processing_state = 1;
+    uint64 updated_at_timestamp_secs = 2;
 }
 
 enum ProcessingState {
--- a/block-streamer/proto/block_streamer.proto
+++ b/block-streamer/proto/block_streamer.proto
@@ -120,7 +120,7 @@ message Health {
 enum ProcessingState {
     UNSPECIFIED = 0;
     IDLE = 1;
-    ACTIVE = 2;
+    RUNNING = 2;
     PAUSED = 3;
-    HALTED = 4;
+    STALLED = 4;
 }
--- a/block-streamer/proto/block_streamer.proto
+++ b/block-streamer/proto/block_streamer.proto
@@ -108,19 +108,26 @@ message StreamInfo {
     string function_name = 4;
     // Block height corresponding to the created/updated height of the indexer
     uint64 version = 5;
-    // Health
+    // Contains health information for the Block Stream
     Health health = 6;
 }
 
+// Contains health information for the Block Stream
 message Health {
+    // The processing state of the block stream
     ProcessingState processing_state = 1;
+    // When the health info was last updated
     uint64 updated_at_timestamp_secs = 2;
 }
 
 enum ProcessingState {
     UNSPECIFIED = 0;
+    // Not started, or has been stopped
     IDLE = 1;
+    // Running as expected
     RUNNING = 2;
+    // Intentionally paused due to some internal condition
     PAUSED = 3;
+    // Stopped due to some unknown error
     STALLED = 4;
 }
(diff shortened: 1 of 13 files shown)


block-streamer/src/block_stream.rs
```diff
@@ -52,16 +54,43 @@ impl<F: Future> Future for PollCounter<F> {
...
+
+    /// Block Stream has stalled due to an error or other condition. Must be manually
+    /// restarted.
+    Stalled,
+}
+
+#[derive(Clone)]
+pub struct BlockStreamHealth {
+    pub processing_state: ProcessingState,
+    pub last_updated: SystemTime,
```
morgsmccauley, line 84: Unlike Runner which pushes the information up, Block Streamer polls it. For this reason I've added a timestamp, so that we can determine whether the data is stale or not. Stale will probably result in restart.

runner/src/stream-handler/worker.ts
```diff
@@ -162,7 +162,7 @@ async function blockQueueConsumer (workerContext: WorkerContext): Promise<void>
         });
 
         const postRunSpan = tracer.startSpan('Delete redis message and shift queue', {}, context.active());
-        parentPort?.postMessage({ type: WorkerMessageType.STATUS, data: { status: IndexerStatus.RUNNING } });
+        parentPort?.postMessage({ type: WorkerMessageType.EXECUTION_STATE, data: { state: ExecutionState.RUNNING } });
```
morgsmccauley, line 165: `IndexerStatus` is outward facing, and we need a bit more granularity for internal use, so opted to create a separate type.

block-streamer/proto/block_streamer.proto
```diff
@@ -108,4 +108,26 @@ message StreamInfo {
...
+
+// Contains health information for the Block Stream
+message Health {
+    // The processing state of the block stream
+    ProcessingState processing_state = 1;
+    // When the health info was last updated
+    uint64 updated_at_timestamp_secs = 2;
+}
+
+enum ProcessingState {
```
darunrs, line 123: It seems you just want to represent any possible states of the service from the get go? I imagine scenarios for UNSPECIFIED and WAITING are vague right now. Specifically for WAITING, I'm confused what that refers to. Does Runner back pressure count as WAITING? If it does, then it could swing between WAITING and RUNNING repeatedly. Otherwise, I'm not sure. I'm also trying to think what action items Coordinator would intend to have when receiving each of these states.

block-streamer/src/block_stream.rs
```diff
@@ -52,16 +54,43 @@ impl<F: Future> Future for PollCounter<F> {
...
-    handle: JoinHandle<anyhow::Result<()>>,
+    stream_handle: JoinHandle<anyhow::Result<()>>,
+    monitor_handle: JoinHandle<()>,
     cancellation_token: tokio_util::sync::CancellationToken,
 }
 
+/// Represents the processing state of a block stream
+#[derive(Clone)]
+pub enum ProcessingState {
+    /// Block Stream is not currently active but can be started. Either has not been started or was
```
darunrs, line 65: A block stream which wasn't started wouldn't even return a state right? It wouldn't be present in the list of block streams.

darunrs (approved):
Looks fine to me! I'm definitely still confused about the states as they look and feel very distinct from any action items though. A state like `INDEXER_BACKFILL` or `LAKE_BACKFILL` make more direct sense to what the block stream is actually doing for example. But I imagine you're keeping the states vague on purpose to avoid having to manage them more frequently? I'd say from a clarity perspective I'd prefer more detailed state information provided there's distinct Coordinator behavior for them.

block-streamer/src/block_stream.rs
morgsmccauley: Yes, and no. The RPC logic is to only add "started" `BlockStream`s. But it's still possible to create a `BlockStream` and not start it, and it would make sense to default to `Running`

morgsmccauley:
> I'd say from a clarity perspective I'd prefer more detailed state information provided there's distinct Coordinator behavior for them.

At this stage we don't have requirement for these more granular states. The goal here is to remove the need for manually restarting the Block Streamer/Runner services. To achieve that, we only need to know if the process has stalled.

We could perhaps expand this to have granular states, and then restart if the bitmap backfill abnormally finished, but I wouldn't say that's required in the short term.

block-streamer/proto/block_streamer.proto
morgsmccauley: > It seems you just want to represent any possible states of the service from the get go?

Yes essentially. While we only really need `Stalled`, it was pretty trivial to add these other states.

And yes, `Waiting` was created specifically for the back pressure case. Doesn't really serve any purpose now. But it could be beneficial to expose these as metrics from Coordinator so we can view what state each indexer is in 🤔

PR #890, opened 2024-07-17
refactor: Fetch Database Connection Parameters in Stream Handler

### LINKED ISSUE #896: Fetch Database Connection Parameters in Stream Handler
Due to our migration of Provisioning logic to a separate API in Runner, we can expect that an Indexers resources will be initialized before we reach the point of executing blocks. So, it is no longer necessary to lazy init DmlHandler or IndexerMeta due to not having the connection parameters at the time of class initialization. We will create a provisioner, get the parameters, create relevant classes, and directly pass them to the Indexer class.

darunrs:
Due to our migration of Provisioning logic to a separate API in Runner, we can expect that an Indexers resources will be initialized before we reach the point of executing blocks. So, it is no longer necessary to lazy init DmlHandler or IndexerMeta due to not having the connection parameters at the time of class initialization. Worker can poll for these params before even beginning to execute any blocks. This also continues to simplify both the code and behavior of the execute function.

### DIFF
--- a/runner/src/stream-handler/stream-handler.ts
+++ b/runner/src/stream-handler/stream-handler.ts
@@ -2,18 +2,17 @@ import path from 'path';
 import { Worker, isMainThread } from 'worker_threads';
 
 import { registerWorkerMetrics, deregisterWorkerMetrics } from '../metrics';
-import Indexer from '../indexer';
 import LogEntry from '../indexer-meta/log-entry';
 import logger from '../logger';
 
 import type IndexerConfig from '../indexer-config';
-import { PostgresConnectionParams } from '../pg-client';
+import type IndexerMeta from '../indexer-meta';
+import { IndexerStatus } from '../indexer-meta';
 
 export enum WorkerMessageType {
   METRICS,
   BLOCK_HEIGHT,
-  STATUS,
-  DATABASE_CONNECTION_PARAMS,
+  EXECUTION_STATE,
 }
 
 export interface WorkerMessage {
@@ -36,25 +35,28 @@ interface ExecutorContext {
 
 export default class StreamHandler {
   private readonly logger: typeof logger;
-  private readonly worker: Worker;
+  private worker: Worker | undefined;
   public readonly executorContext: ExecutorContext;
-  private database_connection_parameters: PostgresConnectionParams | undefined;
+  private readonly indexerMeta: IndexerMeta | undefined;
 
-  constructor(
+  constructor (
     public readonly indexerConfig: IndexerConfig,
   ) {
-    if (isMainThread) {
-      this.logger = logger.child({ accountId: indexerConfig.accountId, functionName: indexerConfig.functionName, service: this.constructor.name });
+    this.logger = logger.child({ accountId: indexerConfig.accountId, functionName: indexerConfig.functionName, service: this.constructor.name });
 
+    this.executorContext = {
+      executionState: ExecutionState.RUNNING,
+      block_height: indexerConfig.version,
+    };
+  }
+
+  async start (): Promise<void> {
+    if (isMainThread) {
       this.worker = new Worker(path.join(__dirname, 'worker.js'), {
         workerData: {
-          indexerConfigData: indexerConfig.toObject(),
+          indexerConfigData: this.indexerConfig.toObject(),
         },
       });
-      this.executorContext = {
-        executionState: ExecutionState.RUNNING,
-        block_height: indexerConfig.version,
-      };
 
       this.worker.on('message', this.handleMessage.bind(this));
       this.worker.on('error', this.handleError.bind(this));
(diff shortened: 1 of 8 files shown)


runner/src/stream-handler/stream-handler.ts
```diff
@@ -53,41 +55,47 @@ export default class StreamHandler {
...
-    indexer.setStoppedStatus().catch((e) => {
-      this.logger.error('Failed to set stopped status for indexer', e);
-    });
-    const errorContent = error instanceof Error ? error.toString() : JSON.stringify(error);
-    const streamErrorLogEntry = LogEntry.systemError(`Encountered error processing stream: ${this.indexerConfig.redisStreamKey}, terminating thread\n${errorContent}`, this.executorContext.block_height);
-
-    indexer.writeCrashedWorkerLog(streamErrorLogEntry)
-      .catch((e) => {
-        this.logger.error('Failed to write failure log for stream', e);
+    if (this.database_connection_parameters) {
```
darunrs, line 68: The fetching of the parameters is still a little bit tricky as the params are needed in both the worker and in StreamHandler (Setting STOPPED status). We also cannot fetch them within the gRPC call as the call is not async. So, we can only do the call in worker and then pass it to StreamHandler.

darunrs:
Performed the following tests with local QueryApi:

- Started a basic indexer and perform a console.log and console.insert and verify DB operations worked
- Set up worker thread to throw error and verified STOPPED status was set in metadata table
- Set up stream handler start function to fail and verified Stalled status was set

runner/src/stream-handler/stream-handler.ts
```diff
@@ -34,59 +37,79 @@ interface ExecutorContext {
...
-        workerData: {
-          indexerConfigData: indexerConfig.toObject(),
-        },
-      });
-      this.executorContext = {
-        executionState: ExecutionState.RUNNING,
-        block_height: indexerConfig.version,
-      };
+    this.executorContext = {
+      executionState: ExecutionState.WAITING,
```
morgsmccauley, line 50: `WAITING` means waiting for some internal condition, i.e. for blocks to be added to the stream. This is really just `IDLE`, can we add that please?

runner/src/stream-handler/stream-handler.ts
```diff
@@ -34,59 +37,79 @@ interface ExecutorContext {
...
+
+        this.executorContext.executionState = ExecutionState.RUNNING;
+      } catch (error: any) {
+        const errorContent = error instanceof Error ? error.toString() : JSON.stringify(error);
+        this.logger.error('Terminating thread', error);
+        this.executorContext.executionState = ExecutionState.STALLED;
+        throw new Error(`Failed to start Indexer: ${errorContent}`);
+      }
     } else {
       throw new Error('StreamHandler should not be instantiated in a worker thread');
```
morgsmccauley, line 80:
```suggestion
      throw new Error('StreamHandler should not be started in a worker thread');
```

runner/src/stream-handler/stream-handler.ts
```diff
@@ -34,59 +37,79 @@ interface ExecutorContext {
...
     }
   }
 
   async stop (): Promise<void> {
-    deregisterWorkerMetrics(this.worker.threadId);
-
+    if (this.worker) {
```
morgsmccauley, line 85: A cool pattern is to separate classes based on their functionality, i.e. `StreamHandler` and `StartedSreamHandler`. The latter would contain the `stop()` method and the former wouldn't. This allows us to have explicit APIs for each and avoid these methods which do nothing under certain conditions, and also the properties which are potentially `undefined`. This is similar to https://github.com/near/queryapi/blob/main/runner/tests/testcontainers/hasura.ts.

Just an FYI, not something you necessarily need to implement here :)

runner/src/stream-handler/stream-handler.ts
```diff
@@ -34,59 +37,79 @@ interface ExecutorContext {
...
-    indexer.setStoppedStatus().catch((e) => {
-      this.logger.error('Failed to set stopped status for indexer', e);
-    });
-    const errorContent = error instanceof Error ? error.toString() : JSON.stringify(error);
-    const streamErrorLogEntry = LogEntry.systemError(`Encountered error processing stream: ${this.indexerConfig.redisStreamKey}, terminating thread\n${errorContent}`, this.executorContext.block_height);
-
-    indexer.writeCrashedWorkerLog(streamErrorLogEntry)
-      .catch((e) => {
-        this.logger.error('Failed to write failure log for stream', e);
+    if (this.indexerMeta) {
```
morgsmccauley, line 96: I feel like this is almost impossible. We won't worker unless this exists, and therefore can't throw these worker errors without it.

Maybe separated classes like I discussed would be a good idea 😅. The `StartedStreamHandler` could require `indexerMeta` to be defined, avoiding this check here.

_but_ probably not a biggy, your call.

morgsmccauley (approved):
Couple small comments, but overall looks good, good job!

runner/src/stream-handler/stream-handler.ts
darunrs: Yep you're right. This felt odd to me as well. I guess if we want to have a separate class, how does that look like from gRPC's perspective? Should I have an array of promises which can be awaited to receive the class? The Started class is what would end up having any useful state information and actionable functions (e.g. stop) after creation. So, after calling start, we would probably discard StreamHandler?

PR #892, opened 2024-07-17
feat: added component, api route, for generate Dummycode

### LINKED ISSUE #876: Create a dummy API to generate code based on selected methods and events
1. An API route in the frontend next.js to generate indexer code (js and sql) based on the provided contract filter, list of selected methods and list of selected events
2. The method should return current default indexer and sql code
3. A unit test that tests this behaviour

Kevin101Zhang:


morgsmccauley (approved):
So this is just a placeholder? You'll integrate with AI later?

Kevin101Zhang:
yes. Created a real API route but it current generates dummy code. The API route is implemented with test. A dummy component accessible through a specific endpoint (hidden from the user) is also added for easy testing/consumption of the endpoint

PR #898, opened 2024-07-19
fix: Ensure provisioning failures are handled

morgsmccauley:
De/provisioning tasks are polled to completion, but we weren't checking if they had failed. This PR updates it so that these failures are picked up.

Additionally, I've added an upper limit to the task polling. All tasks longer than 5mins will be considered timed out.

### DIFF
--- a/coordinator/src/handlers/data_layer.rs
+++ b/coordinator/src/handlers/data_layer.rs
@@ -14,6 +14,8 @@ use crate::indexer_config::IndexerConfig;
 
 type TaskId = String;
 
+const TASK_TIMEOUT_SECONDS: u64 = 300; // 5 minutes
+
 #[derive(Clone)]
 pub struct DataLayerHandler {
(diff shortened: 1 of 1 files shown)


coordinator/src/handlers/data_layer.rs
```diff
@@ -140,30 +152,40 @@ impl DataLayerHandler {
         account_id: AccountId,
         function_name: String,
     ) -> anyhow::Result<()> {
-        tracing::info!(?account_id, ?function_name, "Deprovisioning data layer");
```
morgsmccauley, line 143: `account_id`/`function_name` are already logged by the span, so removing these duplicates.

morgsmccauley:
GitHub actions are currently down: https://www.githubstatus.com/incidents/m70hk23gx3nx

I've run all checks locally, so will proceed with merge.

PR #907, opened 2024-07-22
feat: use QueryAPI Indexer for Explorers and added pagination

### LINKED ISSUE #906: Replace NEAR RPC call with new queryAPI indexer
DEV: https://dev.near.org/dev-queryapi.dataplatform.near/widget/QueryApi.App?selectedIndexerPath=eduohe.near/queryapi_indexer
PROD: https://dev.near.org/dev-queryapi.dataplatform.near/widget/QueryApi.App?selectedIndexerPath=eduohe.near/queryapi_indexer

Kevin101Zhang:
QueryAPI Indexer's explore page now fetches from Indexers and will default to NEAR RPC request if something goes wrong. Added Pagination/Loading more for All Indexers


Kevin101Zhang:
PR is dependent on the creation of QueryAPI Indexer on DEV under dataplatform.near account. Currently dataplatform.near owns QueryAPI Indexer for prod but Eduohe.near owns dev.

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -1,41 +1,132 @@
...
-      Object.keys(data[accId]).forEach((functionName) => {
-        const indexer = {
-          accountId: accId,
-          indexerName: functionName,
-        };
-        if (accId === myAccountId) myIndexers.push(indexer);
-        allIndexers.push(indexer);
-      });
+
+  fetchGraphQL(GET_MY_ACTIVE_INDEXERS, 'getMyActiveIndexers', {
```
darunrs, line 62: It is entirely possible for some problem to occur with fetching the indexers from Hasura. Either now or in the future. I think it would be a good idea to retain the RPC fetch as a fallback. Sort of the same way the social feed does a fallback if the social_feed indexer has some problem.

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -1,41 +1,132 @@
...
+
 const storeIndexerMetaData = () => {
   const url = `${REPL_QUERY_API_USAGE_URL}`;
 
   asyncFetch(url)
     .then(response => {
       if (!response.ok) {
         setError('There was an error fetching the data');
-        throw new Error(`HTTP error! status: ${response.status}`);
+        return;
```
darunrs, line 129: What behavior did throwing the error do, compared to setting the error and returning?

darunrs:
Don't forget to put a PR description!

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -1,41 +1,143 @@
...
+      if (result.status === 200) {
+        const data = result?.body?.data?.[TABLE_NAME];
+        if (Array.isArray(data)) {
+          const newIndexers = data.map(({ author_account_id, indexer_name }) => ({
+            accountId: author_account_id,
+            indexerName: indexer_name,
+          }));
+          setMyIndexers(newIndexers);
+        } else {
+          throw new Error('Data is not an array:', data);
```
darunrs, line 83: Is this possible? I don't see why we can't just convert the data into an array if we had a successful response. What scenario is this covering, where we successfully get data but its malformed in some way? Maybe we should have a more specific data health check than isArray.

darunrs (approved):
Nice work! Hopefully it runs fine in dev. Have you already tested that it works locally? Could you also test the failure case? Specifically, by forcing the graphql call to throw an error so we can verify the fallback works. Eduardo has already posted the indexer under dataplatform_near I believe.

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
Kevin101Zhang: the type from the schema is an array. I can't see the possibility of it being anything else since eduaro's indexer takes care of the inserting.

query MyQuery {
  dataplatform_near_queryapi_indexer_indexers(limit: 1) {
    author_account_id
    indexer_name
  }
}

Response:
{
  "data": {
    "dataplatform_near_queryapi_indexer_indexers": [
      {
        "author_account_id": "roshaan.near",
        "indexer_name": "demo-indexer"
      }
    ]
  }
}

Kevin101Zhang:
Tested in dev and prod. Also tested fallback and adapted the logic to Loadmore as NEAR RPC lists all so load more btn needs to be stateful.

frontend/src/pages/query-api-editor/index.js
```diff
@@ -25,7 +25,7 @@ const QueryApiEditorPage = ({ router }) => {
     );
   }
 
-  if (accountId == 'test' || indexerName == 'test') {
+  if (accountId == 'test' && indexerName == 'test') {
```
Kevin101Zhang, line 28: Logic issue here as dev indexer morg.near/test was defaulting into GenerateCode.

PR #908, opened 2024-07-23
feat: Programmatically Execute Indexer Locally

### LINKED ISSUE #884: Execute indexer with in-memory DmlHandler programmatically
Create Indexer class which uses in memory DmlHandler and fully runs execute.

darunrs:
This PR sets up a local indexer class which leverages the in memory dml handler class and a no op indexer meta class to fully execute indexer logic locally on a block. The class is responsible for downloading block data and providing helper functions to abstract testing logic away. This PR also sets up a core-indexers folder which is used to verify that the created local indexer is actually capable of executing indexer code locally. The core-indexers folder may be temporary as we do intend to eventually publish a testing framework npm package. For the time being, it uses relative imports to work.


runner/src/indexer-meta/no-op-indexer-meta.ts
```diff
@@ -0,0 +1,32 @@
+import { type LocalIndexerConfig } from '../indexer-config/indexer-config';
```
darunrs, line 1: A very simple indexer meta class. I was tempted to roll these if statements into the original class, but aside from feeling like we're doing two different things in the same class, the class also takes in PgClient which is very inconvenient for a local indexer oriented class. So, I broke it out into its own class.

runner/src/indexer-meta/indexer-meta.ts
```diff
@@ -19,7 +19,13 @@ export enum MetadataFields {
   LAST_PROCESSED_BLOCK_HEIGHT = 'LAST_PROCESSED_BLOCK_HEIGHT'
 }
 
-export default class IndexerMeta {
+export interface IndexerMetaInterface {
```
darunrs, line 22: IIndexerMeta did not seem like a good name, so I explicitly put Interface as a suffix instead. I'll rename IDmlHandler to DmlHandlerInterface to match the new convention.

runner/src/indexer/indexer.ts
```diff
@@ -40,11 +41,12 @@ export interface TableDefinitionNames {
...
   hasuraAdminSecret: string
   hasuraEndpoint: string
+
 }
 
 const defaultConfig: Config = {
-  hasuraAdminSecret: process.env.HASURA_ADMIN_SECRET,
-  hasuraEndpoint: process.env.HASURA_ENDPOINT,
+  hasuraAdminSecret: process.env.HASURA_ADMIN_SECRET = '',
+  hasuraEndpoint: process.env.HASURA_ENDPOINT = '',
```
darunrs, line 49: I have a problem here where the environment of Indexer tells NodeJS that these env variables are set, whereas wherever I use this class for local development may say it is not set. I've decided to set them to empty string and assert on them. However, this will be extracted anyway when I pull context.db creation to a separate class. This is a simple short term solution.

runner/src/lake-client/lake-client.ts
```diff
@@ -68,15 +65,6 @@ export default class LakeClient {
   }
 
   async fetchBlock (blockHeight: number): Promise<Block> {
-    const cachedMessage = await this.redisClient.getStreamerMessage(blockHeight);
```
darunrs, line 71: I wanted to reuse this class for local development, but it contained this cache stuff which was leading to hanging RedisClient connections. I decided to extract the cache logic to worker for now. I think what I'll do is create a LakeClientWithCache interface and implement a LakeClientWithRedisCache class which has this logic added back in. I can later create a LakeClientWithLocalCache class which caches blocks locally as files, which can be used with LocalIndexer.

PR #913, opened 2024-07-23
fix: used monaco compiler options latest

### LINKED ISSUE #865: enable js to a higher version in monaco editor
Image

Kevin101Zhang:
fix: Set monaco compiler options to latest

### DIFF
--- a/frontend/src/components/Editor/EditorComponents/Editor.tsx
+++ b/frontend/src/components/Editor/EditorComponents/Editor.tsx
@@ -284,7 +284,7 @@ const Editor: React.FC = (): ReactElement => {
     );
     
     monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
-      target: monaco.languages.typescript.ScriptTarget.Latest,
+      target: monaco.languages.typescript.ScriptTarget.ES2018,
       allowNonTsExtensions: true,
       moduleResolution: monaco.languages.typescript.ModuleResolutionKind.NodeJs,
     });
--- a/frontend/src/components/Editor/EditorComponents/Editor.tsx
+++ b/frontend/src/components/Editor/EditorComponents/Editor.tsx
@@ -282,8 +282,9 @@ const Editor: React.FC = (): ReactElement => {
       `${primitives}}`,
       'file:///node_modules/@near-lake/primitives/index.d.ts',
     );
+    
     monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
-      target: monaco.languages.typescript.ScriptTarget.ES2016,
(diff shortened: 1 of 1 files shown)


frontend/src/components/Editor/EditorComponents/Editor.tsx
```diff
@@ -282,8 +282,9 @@ const Editor: React.FC = (): ReactElement => {
       `${primitives}}`,
       'file:///node_modules/@near-lake/primitives/index.d.ts',
     );
+    
     monaco.languages.typescript.typescriptDefaults.setCompilerOptions({
-      target: monaco.languages.typescript.ScriptTarget.ES2016,
+      target: monaco.languages.typescript.ScriptTarget.Latest,
```
Kevin101Zhang, line 287: print: monaco.languages.typescript.ScriptTarget

{
    "0": "ES3",
    "1": "ES5",
    "2": "ES2015",
    "3": "ES2016",
    "4": "ES2017",
    "5": "ES2018",
    "6": "ES2019",
    "7": "ES2020",
    "99": "Latest",
    "100": "JSON",
    "ES3": 0,
    "ES5": 1,
    "ES2015": 2,
    "ES2016": 3,
    "ES2017": 4,
    "ES2018": 5,
    "ES2019": 6,
    "ES2020": 7,
    "ESNext": 99,
    "JSON": 100,
    "Latest": 99
}

Kevin101Zhang:
Are there any issues where setting to the latest compiler options can be harmful?

Kevin101Zhang:
Screenshot 2024-07-23 at 4 38 12 PM
Screenshot 2024-07-23 at 4 50 08 PM

darunrs:
So the backend [runs](https://github.com/near/queryapi/blob/main/runner/tsconfig.json) with a target of es2018 and a runtime of 2021. I'm not totally sure of the difference, even with the comments there. But in any case, that's what we use. I would say maybe we should match them. I'm not clear on what the differences between the versions are, but I would surmise there's at least some, if not significant, differences. Since we use VM2 under the hood, we're kinda forced to be stuck with specific Node/TS/etc. versions. So, perhaps we can stick to either 2018 or 2021.

Kevin101Zhang:
Sorry this got lost in the sprint cycle from last week. I modified the monaco editor to run in 2018 to match the backend.
In this case - all the frontend code is compatible with the same features as the backend target. It should make it easier to share code if necessary.

ES2021 may introduce discrepancies with the backend.

regardless .entries got introduced in 2017 so either work.

darunrs (approved):

PR #918, opened 2024-07-23
fix: Add gensonjs to changed selectMethod and selectEvent to proper TS shape

### LINKED ISSUE #914: Change method and eventFilters to proper shape

Kevin101Zhang:


frontend/src/pages/api/generateCode.ts
```diff
@@ -1,90 +1,142 @@
...
+  Null = 'null',
+  Boolean = 'boolean',
+  Integer = 'integer',
+  Number = 'number',
+  String = 'string',
+  Object = 'object',
+  Array = 'array',
+}
+
+type Schema = {
```
darunrs, line 16: Why are you defining these instead of importing them from the gensen library? You should be npm installing them and using them directly?

I'm also confused why there is so much type validation code here. If your endpoint returns with a 200, just use the data like you want to. Put var: Schema all over. It's TS so you don't even have runtime type checks anyway. The way I see it, there's only two options if, for some reason, the data is not a Schema type (Again, how could that be. We use gensen-js in the endpoint):
1. Your validation code sees this, and throws an error. We stop. 
2. Your code tries to use the data like normal, but it fails, throwing an error. 

At best, you throw an error anyway, maybe with a nicer error message, which you could do anyway with broader try catches. At the end of the day, if the data is bad, there's nothing this code can do about it other than throw its error. So, I don't see a point in adding a ton of extra code to do validation which then needs to be maintained, maybe even migrated, over time. 

I would say you should just parse the JSON, and set it into a variable of type Schema. And then move along, using types as you need. If the data is malformed, you'll get a "Tried to access [some attribute] but was undefined" or something.

darunrs (commented):
I'm not too keen on the reinventing of the wheel going on in this PR. We should import from genson-js, not define Schema and other types. In addition, genson-js already has functions for parsing an object and returning a Schema type anyway in functions like createSchemaForObject. We do not need so much validation code that we then need to maintain and write unit tests for. 

Can you remove any extra code you added and instead just do something like this:
```
const json = JSON.parse(response);
const schema: Schema = createSchemaForObject(json);
```

This does validation of the entire object and returns the right type.

Kevin101Zhang:
I initially came in with genson and as I was exploring the package I thought of our use case and said "it can't be worth importing an entire library with 23 dependents where we just need their types right?" The js library differs from its python parent and its main functionality is to infer a schema from a JSON object with additional functionality supported. The response we receive from Eduardo had just came from genson so it further detered me.  So I moved away from using genson and started drafting it without the use of the package. after seeing the file today and its growth due to validation and with a fresh set of eyes and not 7:30pm the use case is definitely stronger

darunrs:
Haha no worries. I think we should do the import of the library if it makes our lives easier. We're not in a position where we need to maximize compile speed or size. If that were a concern, we would certainly do things differently. Plus, as mentioned, by using the same library, it provides some level of confidence, which means we have less things that need thorough testing. I know we have various deadline targets we try to hit, but please don't need to feel so stressed over the deliverable soft deadline to feel the need to put out work when you're really tired. I think it's worth waiting the day if it brings the best performance you can to the task! The rest of us can surely wait the night (Unless its a critical failure blocking customers).

Kevin101Zhang:
I didnt have direct access to createSchemaForObject. createSchema is one the accessible methods.

index.d.ts
```
export * from './types';
export { extendSchema, mergeSchemas, createSchema, createCompoundSchema } from './schema-builder';
export * from './schema-comparator';
```

darunrs (approved):
Nice work!

frontend/src/components/Editor/EditorComponents/GenerateCode.tsx
```diff
@@ -0,0 +1,164 @@
+//Dummy component
+import { useState } from 'react';
+
+type Schema = {
+  type: string;
+  properties?: Record<string, { type: string }>;
```
pkudinov, line 6: I think this would be recursive, @Kevin101Zhang , because I think @eduohe is not really converting nested schemas to strings, does he?

pkudinov (changes requested):

frontend/src/pages/api/generateCode.ts
```diff
@@ -1,90 +1,138 @@
...
+    return false;
+  }
+};
 
+export const validateRequestBody = (body: any): body is RequestBody => {
   return (
     isStringOrArray(body.contractFilter) &&
     Array.isArray(body.selectedMethods) &&
-    body.selectedMethods.every((method: any) => typeof method === 'string') &&
+    body.selectedMethods.every(isValidMethod) &&
```
Kevin101Zhang, line 40: Iterate through every select method and confirm with genson

frontend/src/components/Editor/EditorComponents/GenerateCode.tsx
pkudinov: So basically you need to use `Schema` type from genson as is, unless I misunderstand something

Kevin101Zhang: Oh I didnt update the frontend component since its a dummy

Kevin101Zhang: I should be grabbing the types from genson. Please check the api

pkudinov (approved):

PR #926, opened 2024-07-25
feat: explorers page indexer to find most popular indexers (desc)

### LINKED ISSUE #925: Allow explorers page to find most popular indexers

Kevin101Zhang:
display indexers by numQueries Desc. 

**High level Approach:**
onLoad set metadata from databricks to local state. 
immediately after QueryAPI indexer grabs all results and any matching metadata there is from the above state and sets all data new state that has all indexer data.

pagination loads 50 and loads locally from all indexer data leading to an instant interaction and loading

### DIFF
--- a/frontend/widgets/src/QueryApi.IndexerCard.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerCard.jsx
@@ -1,15 +1,12 @@
-const { accountId, indexerName, indexerMetadata } = props;
-const sanitizedAccountID = accountId.replace(/\./g, '_');
-const key = `${sanitizedAccountID}/${indexerName}`;
+const { accountId, indexerName, numQueries } = props;
 
 const indexer = {
   accountId,
   indexerName,
-  ...(indexerMetadata.has(key) && indexerMetadata.get(key))
+  numQueries,
 };
 
 const editUrl = `https://dev.near.org/${REPL_ACCOUNT_ID}/widget/QueryApi.App?selectedIndexerPath=${accountId}/${indexerName}`;
-
 const playgroundLink = `https://cloud.hasura.io/public/graphiql?endpoint=${REPL_GRAPHQL_ENDPOINT}/v1/graphql&header=x-hasura-role%3A${accountId.replace(/\./g, '_')}`;
 const formatNumberWithCommas = (number) => number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
 
--- a/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
@@ -34,18 +34,17 @@ query getMyActiveIndexers($authorAccountId: String!) {
 `;
 
 const [selectedTab, setSelectedTab] = useState(props.tab && props.tab !== "all" ? props.tab : "all");
+const [indexerMetaData, setIndexerMetaData] = useState(new Map());
+const [hasMetadataRendered, setHasMetadataRendered] = useState(false);
 
 const [indexers, setIndexers] = useState([]);
 const [total, setTotal] = useState(0);
 const [currentPageIndexer, setCurrentPageIndexer] = useState([]);
+const [page, setPage] = useState(0);
 
 const [myIndexers, setMyIndexers] = useState([]);
-const [indexerMetaData, setIndexerMetaData] = useState(new Map());
-const [page, setPage] = useState(0);
-const [hasMetadataRendered, setHasMetadataRendered] = useState(false);
 
 const [loading, setLoading] = useState(false);
-const [isLoadingMore, setIsLoadingMore] = useState(false);
 const [error, setError] = useState(null);
 
 const fetchGraphQL = (operationsDoc, operationName, variables) => {
@@ -84,6 +83,7 @@ const fetchIndexerData = () => {
         const sortedIndexers = newIndexers.sort((a, b) => (b.numQueries ?? 0) - (a.numQueries ?? 0));
         setTotal(totalCount);
         setIndexers(sortedIndexers);
+        setCurrentPageIndexer(sortedIndexers.slice(0, PAGE_SIZE));
       } else {
         throw new Error('Data is not an array:', data);
       }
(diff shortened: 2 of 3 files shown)


frontend/widgets/src/QueryApi.IndexerCard.jsx
```diff
@@ -1,15 +1,12 @@
-const { accountId, indexerName, indexerMetadata } = props;
-const sanitizedAccountID = accountId.replace(/\./g, '_');
-const key = `${sanitizedAccountID}/${indexerName}`;
+const { accountId, indexerName, numQueries } = props;
 
 const indexer = {
   accountId,
   indexerName,
-  ...(indexerMetadata.has(key) && indexerMetadata.get(key))
+  numQueries,
```
darunrs, line 6: I preferred having all available metadata here. It's easier to then build upon for other sorting methods we can make available to users. Can we return to that?

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -3,11 +3,9 @@ const myAccountId = context.accountId;
 const PAGE_SIZE = 50;
 const TABLE_NAME = "dataplatform_near_queryapi_indexer_indexers";
 const GET_ALL_ACTIVE_INDEXERS = `
-query getAllActiveIndexers($limit: Int!, $offset: Int!) {
+query getAllActiveIndexers {
```
darunrs, line 6: Are we simply accepting the latency hit of fetching all indexers with the intention that having them all available is just more convenient for end users? Or did we solve the query speed somehow?

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -64,87 +62,101 @@ const fetchGraphQL = (operationsDoc, operationName, variables) => {
...
-      if (result.status === 200) {
-        const data = result?.body?.data?.[TABLE_NAME];
-        if (Array.isArray(data)) {
-          const newIndexers = data.map(({ author_account_id, indexer_name }) => ({
+  fetchGraphQL(GET_ALL_ACTIVE_INDEXERS, 'getAllActiveIndexers').then((result) => {
+    if (result.status === 200) {
+      const data = result?.body?.data?.[TABLE_NAME];
+      const totalCount = result?.body?.data?.[`${TABLE_NAME}_aggregate`]?.aggregate?.count;
+      if (Array.isArray(data)) {
+        //inject metadata if exist
```
darunrs, line 72: What happens if the metadata loads after the indexer fetch?

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -64,87 +62,101 @@ const fetchGraphQL = (operationsDoc, operationName, variables) => {
...
-          throw new Error('Data is not an array:', data);
-        }
+            ...(indexerMetaData.has(key) && indexerMetaData.get(key))
+          })
+        });
+        // sort by numQueries
+        const sortedIndexers = newIndexers.sort((a, b) => (b.numQueries ?? 0) - (a.numQueries ?? 0));
+        setTotal(totalCount);
+        setIndexers(sortedIndexers);
+        setCurrentPageIndexer(sortedIndexers.slice(0, PAGE_SIZE));
```
darunrs, line 86: Does any set operation rerender the frontend? Trying to wrap my head around the timing of the following:
1. Completion of the active indexer fetch
2. Completion of the metadata fetch
3. Enriching of the active indexer array
4. Reloading of frontend to sort

If metadata comes after the active fetch, does the setting of metadata trigger something which leads to the frontend to reorder indexers after new info has been introduced?

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -64,87 +62,101 @@ const fetchGraphQL = (operationsDoc, operationName, variables) => {
...
       } else {
-        throw new Error('Failed to fetch data:', result);
+        throw new Error('Data is not an array:', data);
       }
-      setLoading(false);
-    })
+    } else {
+      throw new Error('Failed to fetch data:', result);
+    }
+    setLoading(false);
```
darunrs, line 93: What does this do?

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -155,30 +167,34 @@ const storeIndexerMetaData = () => {
...
+  fetchIndexerMetadata();
+}, []);
 
 useEffect(() => {
-  if (selectedTab === "my-indexers") {
-    if (myIndexers.length <= 0) fetchMyIndexerData();
-  }
-  if (selectedTab === "all") {
-    if (indexers.length <= 0) fetchIndexerData(page, page > 0);
+  if (hasMetadataRendered) {
```
darunrs, line 182: Oh I see. So we only fetch AFTER we have metadata. Hopefully the latency is still decent for loading. If not, we may need to find something we can do to do both queries at the same time. Although, I do remember you mentioning that is very vhard to do.

darunrs (approved):
Looks good! Some comments around understanding it.

frontend/widgets/src/QueryApi.IndexerCard.jsx
Kevin101Zhang: Previously we were doing a call per indexer to the Map() object to check if the map has the data if so give the value as props. Now we do it in the call when we get the data instead and pass it simply as props.

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
Kevin101Zhang: The Query is simply just the graphql request of 205 indexers which is trivial. I dont expect and major latency issues in the future unless we have a massive increase in indexers. The issue came from the frontend rendering the 205 all at once. What I do now is store the data in a state and have a separate state to load out 50 indexers at a time. In this case when they click load more we are simply updating a state with data we already have.

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
Kevin101Zhang: If the request fails the user has already waited for a period of time. So this sets the loading spinner to false. Showcases them an error message 
```'An error occurred while retrieving indexer data. Attempting to fetch from NEAR RPC'```
while fetching from the backupNEARRPC to replace the error message. It just keeps the user in the loop

PR #929, opened 2024-07-25
refactor: Move Context object to new class

### LINKED ISSUE #909: Refactor Context.Db Building to Separate Class
In order to expose context.db methods for local testing, the context object needs to be exposed as a class variable. Currently, the object is built inside execute. The building of the context object can be extracted to a separate class and set as a class variable.

darunrs:
All logic associated with the creation of the context object resides inside the Indexer class, which bloats the file and complicates unit testing. In addition, it prevents access to the context object itself so that these calls can be made through the Indexer class itself. 

This PR refactors this logic into its own class, allowing access to context methods through Indexer. The object is still recreated each time it is accessed as the object bakes in some state information (block height and log entry array) which need to be updated on subsequent accesses. 

I've also reorganized the folders so that we have fewer folders under src as it is getting cluttered.


runner/src/context-builder/context-builder.ts
```diff
@@ -0,0 +1,292 @@
...
+          this.logger.error('Error writing log of graphql error', e);
+        }
+      }
+      throw new Error(`Failed to write graphql, http status: ${response.status}, errors: ${JSON.stringify(errors, null, 2)}`);
+    }
+
+    return data;
+  }
+
+  buildContext (blockHeight: number, logEntries: LogEntry[]): ContextObject {
```
darunrs, line 123: I thought about integrating the functions into this class itself, or having a persistent object to update, but I wasn't sure how that would work when frozen into the VM or with any references to class variables. I decided to continue building an object when requested.

runner/src/context-builder/context-builder.ts
```diff
@@ -0,0 +1,292 @@
...
+      return result;
+    } catch (err) {
+      const error = err as Error;
+      logEntries.push(LogEntry.systemWarn(`Caught error when generating context.db methods: ${error.message}`));
+    }
+    return {}; // Default to empty object if error
+  }
+}
+
+// TODO: Migrate all below code to separate class
```
darunrs, line 240: All of the below code will be moved to its own class and have its own unit tests. I've simply placed them here for now to reduce PR size.

runner/src/context-builder/context-builder.ts
```diff
@@ -0,0 +1,292 @@
+import fetch from 'node-fetch';
```
morgsmccauley, line 1: It would be good to start nesting relating components in the filesystem to provide some additional context in to the relationships in our codebase, e.g. butting `context-builder` under `indexer` as it's directly related to that? `src/` is starting to grow so would be good to start encapsulating/grouping related logic.

morgsmccauley (approved):

PR #956, opened 2024-07-31
fix: Require exact config response for lifecycle transitions

### LINKED ISSUE #940: Update contract to return specific data when fetching deleted indexer
Deploy change to contract which has it return data instead of null for a matched indexer which was deleted.

### LINKED ISSUE #941: Add Status Code and content check for indexer config fetch
Coordinator's lifecycle polls the contract for indexer config through RPC. The response for a no match case is null. We handled this as "The indexer was deleted", which is not incorrect. However, it was possible for the call to successfully return with an incorrect null response. As a result, lifecycle would erroneously enter its indexer deletion workflow, which is not reversible. 

This PR takes advantage of contract updates to now expect a response from the contract when an indexer is deleted, rather than expecting a null response for the same.

darunrs:
We've noticed that it is occasionally possible for the indexer config returned by the contract to be parseable/null with a success status code, even though the indexer itself was not in fact deleted. This triggers the deprovisioning workflow anyway, which currently throws the Indexer into a repairing state, where it is no longer operating. 

This PR updates the contract to retain configuration if the indexer was deleted. Coordinator will now expect a specific configuration to take deletion actions, instead of the previous method which left things open to interpretation.

In the near future, the contract will be migrated (#964) to use an enum instead of contract format, to avoid wastefully storing things like code/schema for deleted indexers forever.

### DIFF
--- a/coordinator/src/indexer_config.rs
+++ b/coordinator/src/indexer_config.rs
@@ -13,6 +13,7 @@ pub struct IndexerConfig {
     pub rule: Rule,
     pub updated_at_block_height: Option<u64>,
     pub created_at_block_height: u64,
+    pub deleted_at_block_height: Option<u64>,
 }
 
 impl KeyProvider for IndexerConfig {
@@ -39,6 +40,7 @@ impl Default for IndexerConfig {
             },
             created_at_block_height: 1,
             updated_at_block_height: Some(2),
+            deleted_at_block_height: Some(3),
             start_block: StartBlock::Height(100),
         }
     }
@@ -52,5 +54,9 @@ impl IndexerConfig {
     pub fn get_registry_version(&self) -> u64 {
         self.updated_at_block_height
             .unwrap_or(self.created_at_block_height)
+    }
+
+    pub fn is_deleted(&self) -> bool {
+        self.deleted_at_block_height.is_some()
     }
 }
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -46,6 +46,8 @@ enum StorageKeys {
     AccountV3(CryptoHash),
     RegistryV4,
     AccountV4(CryptoHash),
+    RegistryV5,
+    AccountV5(CryptoHash),
 }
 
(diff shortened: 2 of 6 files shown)


coordinator/src/indexer_config.rs
```diff
@@ -13,6 +13,7 @@ pub struct IndexerConfig {
     pub rule: Rule,
     pub updated_at_block_height: Option<u64>,
     pub created_at_block_height: u64,
+    pub deleted_at_block_height: Option<u64>,
 }
```
darunrs, line 17: I get the idea of having an indexer config redefined here so that we have helper functions. But, it does lead to duplication of attribute definitions. I wonder if there's a way to make this a new type with registry's IndexerConfig inside it? Like a `IndexerLifecycleConfig` or something. Decided not to mess with it in this PR for now though. Would like to get your thoughts.

registry/contract/src/lib.rs
```diff
@@ -275,7 +273,7 @@ impl Contract {
         let account_indexers =
             self.registry
                 .entry(account_id.clone())
-                .or_insert(IndexerConfigByFunctionName::new(StorageKeys::AccountV3(
```
darunrs, line 278: So apparently register was inserting using V3's hash this whole time... I don't know how this will impact migration. Hopefully it goes fine?

registry/contract/src/lib.rs
```diff
@@ -359,11 +357,12 @@ impl Contract {
     pub fn list_by_account(&self, account_id: AccountId) -> AccountIndexers {
         self.registry
             .get(&account_id)
-            .unwrap_or(&IndexerConfigByFunctionName::new(StorageKeys::AccountV3(
+            .unwrap_or(&IndexerConfigByFunctionName::new(StorageKeys::AccountV4(
```
darunrs, line 360: Should this actually be V5?

morgsmccauley (approved):

darunrs:
Verified that register and remove work as expected with actual contract. The code should be good to go.

PR #957, opened 2024-07-31
fix: Increase timeout for block stream health check

### LINKED ISSUE #948: Increase block stream stalled health check

darunrs:
Increasing timeout for the block stream health check to ensure unwanted stalled flagging of properly executing block streams. When QueryApi is restarted, a thundering herd situation can lead to slower queries.


morgsmccauley (commented):
We also need to increase the timeout in coordinator

PR #962, opened 2024-07-31
fix: removed overwritting of indexername in obj

### LINKED ISSUE #961: indexer name is served incorrectly to the URL

Kevin101Zhang:
Databricks stores account and function names in their sanitized form (Only alphanumerics and underscores) as this is how they are stored in Hasura. However, the actual indexer and accounts are stored in both RPC and the queryapi_indexer under their original names. 

This PR prevents the metadata from databricks overwriting the original names with the sanitized ones.

### DIFF
--- a/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
@@ -158,7 +158,6 @@ const fetchIndexerMetadata = () => {
           original_deployment_date
         }) => {
           const indexer = {
-            indexerName: indexer_name,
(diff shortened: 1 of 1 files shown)


frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -158,7 +158,6 @@ const fetchIndexerMetadata = () => {
           original_deployment_date
         }) => {
           const indexer = {
-            accountId: indexer_account_id,
             indexerName: indexer_name,
```
darunrs, line 161: Can we remove this too?

darunrs (approved):

PR #969, opened 2024-08-01
feat: events workflow in wizard

### LINKED ISSUE #966: Launchpad Events

### LINKED ISSUE #971: add select and unselect *

Kevin101Zhang:
added events into launchpad.


frontend/src/test/api/generateCode.test.ts
```diff
@@ -65,6 +65,7 @@ describe('generateCode API', () => {
           },
         },
       ],
+      selectedEvents: [],
```
darunrs, line 68: Is there really no other testing that can be performed? It feels odd to see that the only modification to tests needed is adding one line, when enabling some feature.

frontend/widgets/src/QueryApi.Launchpad.jsx
```diff
@@ -547,31 +442,69 @@ const handleFetchCheckboxData = async () => {
...
+      setEventsData(data.events);
+      setEventCount(data.events.length);
       setLoading(false);
     }).catch(error => {
       setLoading(false);
       setError('There was an error fetching the data');
     });
+};
+
+const handleSelectUnselectAll = (action) => {
```
darunrs, line 456: handleSelectUnselectAll is a really confusing name. Do you mean selecting unselect all? Or it works for both? Could you rename this to something else, closer to what it is actually doing?

frontend/widgets/src/QueryApi.Launchpad.jsx
```diff
@@ -580,20 +513,65 @@ const handleParentChange = (methodName) => {
 };
 
 const handleChildChange = (key) => {
```
darunrs, line 515: I've seen these references to things like parents and children. They do not inherently explain what that is. I feel a lot of this code is really living in a context which is really hard to understand if you weren't the one who wrote it. Instead of using terms like child or parent, I would prefer something more declarative. If parent meant say main method and child is arguments, they should just be referenced that way. Let's try to use more direct self explanatory terms.

frontend/widgets/src/QueryApi.Launchpad.jsx
```diff
@@ -632,67 +610,121 @@ return (
...
-                          </>
-                        )
+                      {methodsData.length > 0 && tab === 'methods_tab' && (
+                        <>
+                          <CheckboxContainer>
+                            <CheckboxLabel>
+                              <Checkbox
+                                type="checkbox"
+                                id="select_all"
+                                onChange={() => handleSelectUnselectAll('select_all')}
```
darunrs, line 622: Ok it seems like you've got this handle managing this checkbox. Maybe something like handle[Specific thing here]SelectionChange. Like, handleMethodSelectionChange or something.

darunrs (approved):
Mostly alright to me. Some comments on code readability. There's natrually going to be little testability to be done here, so mainly relying on the expectation that this has been run locally and works like expected.

frontend/widgets/src/QueryApi.Launchpad.jsx
Kevin101Zhang: Screenshot 2024-08-02 at 3 56 19 PM

Yeah that makes sense. The intent of the naming here is to mock what happens when a user selects a childcheckbox and parent checkbox.

handleParentChange - parent checkbox should select and unselect all child checkboxes
handleChildChange - child checkbox should just check the parent checkbox

The code is setup in a way where if we choose to display more then 1 level it should work with minimal modifications. 

low effort but perhaps? 
handleChildCheckboxChange 
handleParentCheckboxChange

PR #960, opened 2024-07-31
feat: added context.error to context methods

Kevin101Zhang:

PR #970, opened 2024-08-02
Prod Release 06/08/24

darunrs:
- fix: Require exact config response for lifecycle transitions (#956)
- feat: events workflow in wizard (#969)
- feat: added context.error to context methods (#960)
- chore: remove empty filter/empty indexer name JSX elements when creat… (#973)
- test: Test fetching indexer config from registry (#976)
- test: Coodinator handlers (#977)
- fix: Receiver blocks stream would not query today (#987)
- fix: display issue in explorer indexers (#981)
- feat: run formatter/prettier on launchpad's generated code in Editor (#975)
- fix: Decreasing Processed Block Error (#988)


darunrs:
Prod contract deployment was successful. Tested list_all and that worked. Will do release before testing register/delete.

PR #975, opened 2024-08-02
feat: run formatter/prettier on launchpad's generated code in Editor

### LINKED ISSUE #974: run formatter/prettier on launchpad's generated code in Editor

Kevin101Zhang:


frontend/src/components/Editor/EditorComponents/Editor.tsx
```diff
@@ -151,8 +151,9 @@ const Editor: React.FC = (): ReactElement => {
         }
 
         const codeResponse = await generateCode(wizardContractFilter, wizardMethods, wizardEvents);
-        setIndexingCode(codeResponse.jsCode);
-        setSchema(codeResponse.sqlCode);
+        const { validatedCode, validatedSchema } = reformatAll(codeResponse.jsCode, codeResponse.sqlCode);
```
Kevin101Zhang, line 154: Since we handle and error throwing and check for properties within generateCode, we can safely assume the response is correct at this point so we can call for formatting

darunrs: This use effect seems to be triggered by nothing? I'm confused when this runs. I'm mainly worried about the reformat changing the code suddenly for the user, which may be confusing.

darunrs (commented):
Question here as well. I noticed that if I go to an indexer, the diff view compares the default code (the one where we do context.set on the block height) with my current code. Not my code when I opened the page, and my current code. Could you investigate that? Seeing as this PR is non-essential and small, it might be better to do that here. 

Also, this PR is titled something related to formatting, but no files added to this PR have any formatting changes. Are you sure you added them? Otherwise, if the formatting was already correct, it should be re-titled to cover whatever DID change.

PR #977, opened 2024-08-05
test: Coodinator handlers

### LINKED ISSUE #993: Add unit tests for Coordinator

morgsmccauley:
Adds tests for: `DataLayerHandler`, `BlockStreamsHandler`, and `ExecutorsHandler`. No functional changes, but I did need to extract/create internal client wrappers so that the RPC requests can be mocked.

### DIFF
--- a/coordinator/src/handlers/block_streams.rs
+++ b/coordinator/src/handlers/block_streams.rs
@@ -186,7 +186,7 @@ impl BlockStreamsHandler {
         Ok(())
     }
 
-    async fn reconfigure_block_stream(&self, config: &IndexerConfig) -> anyhow::Result<()> {
+    async fn reconfigure(&self, config: &IndexerConfig) -> anyhow::Result<()> {
         if matches!(
             config.start_block,
             StartBlock::Latest | StartBlock::Height(..)
@@ -249,7 +249,7 @@ impl BlockStreamsHandler {
         Ok(height)
     }
 
-    async fn resume_block_stream(&self, config: &IndexerConfig) -> anyhow::Result<()> {
+    async fn resume(&self, config: &IndexerConfig) -> anyhow::Result<()> {
         let height = self.get_continuation_block_height(config).await?;
 
         tracing::info!(height, "Resuming block stream");
@@ -297,7 +297,7 @@ impl BlockStreamsHandler {
         Ok(())
     }
 
-    pub async fn synchronise_block_stream(
+    pub async fn synchronise(
         &self,
         config: &IndexerConfig,
         previous_sync_version: Option<u64>,
@@ -319,7 +319,7 @@ impl BlockStreamsHandler {
 
             self.stop(block_stream.stream_id.clone()).await?;
 
-            self.reconfigure_block_stream(config).await?;
+            self.reconfigure(config).await?;
 
             return Ok(());
         }
@@ -331,12 +331,12 @@ impl BlockStreamsHandler {
         }
 
         if previous_sync_version.unwrap() != config.get_registry_version() {
-            self.reconfigure_block_stream(config).await?;
+            self.reconfigure(config).await?;
 
             return Ok(());
         }
 
-        self.resume_block_stream(config).await?;
+        self.resume(config).await?;
 
         Ok(())
     }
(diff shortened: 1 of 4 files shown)


.github/workflows/coordinator-ci.yml
```diff
@@ -21,6 +21,13 @@ jobs:
       uses: arduino/setup-protoc@v2
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Install Rust
```
morgsmccauley, line 24: Will conflict with #976 - I'll remove from whichever is merged first

coordinator/src/handlers/block_streams.rs
```diff
@@ -335,3 +354,374 @@ impl BlockStreamsHandler {
...
+            .expect_get_last_published_block::<IndexerConfig>()
+            .returning(move |_| Ok(Some(last_published_block)));
+
+        let handler = BlockStreamsHandler {
+            client: mock_client,
+            redis_client: mock_redis,
+        };
+
+        handler
+            .synchronise(&config, Some(config.get_registry_version()))
```
darunrs, line 410: I think the utility of having a function which encapsulates "Make the state match the action" is neat, it feels like it abstracts a lot of behavior. I think I would prefer if we had functions which described what they did. For example, "handle_update" or "handle_restart". So, the handler contains direct and concrete actions you can take with the resource, and lifecycle is fully responsible for calling them when appropriate. This removes the sharing of logical responsibility between the handler and lifecycle, where lifecycle seems to defer to the handler for things. What do you think?

coordinator/src/handlers/block_streams.rs
```diff
@@ -335,3 +354,374 @@ impl BlockStreamsHandler {
...
+        };
+
+        handler
+            .synchronise(&config, Some(config.get_registry_version() - 1))
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    async fn ignores_healthy_streams() {
```
darunrs, line 632: Should there be similar tests for other non-unhealthy states such as WAITING?

coordinator/src/handlers/executors.rs
```diff
@@ -193,3 +211,199 @@ impl ExecutorsHandler {
...
+
+        let handler = ExecutorsHandler {
+            client: mock_client,
+        };
+
+        handler.synchronise(&config).await.unwrap()
+    }
+
+    #[tokio::test]
+    async fn restarts_unhealthy_executors() {
```
darunrs, line 319: So is it here that we would change the test, once we add in limited retries?

darunrs (approved):

coordinator/src/handlers/block_streams.rs
```diff
@@ -335,3 +354,374 @@ impl BlockStreamsHandler {
...
+        };
+
+        handler
+            .synchronise(&config, Some(config.get_registry_version() - 1))
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    async fn restarts_unhealthy_streams() {
```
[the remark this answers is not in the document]
morgsmccauley: `tokio::time::pause` handles that

coordinator/src/handlers/block_streams.rs
morgsmccauley: Yup, I can do that

coordinator/src/handlers/executors.rs
morgsmccauley: I think so, we'd expect it to stop after some amount of time.

coordinator/src/handlers/block_streams.rs
morgsmccauley: Yeah, I think I agree, but am also on the fence. On one hand, it's nice that the lifecycle is thin, and essentially manages state changes only. On the other, it encapsulates a lot like you've mentioned.

The encapsulation also allows us to refactor lifecycle quite easily, as the majority of the logic is delegated. This logic is pretty concrete and won't change. I've been burnt in the past having to refactor synchronisation many times, but with this encapsulation I can refactor around it.

I'd like to sit on this for a bit longer to really get a feel for what is/isn't working. I've also been burnt in the past pre-optimising. Are you ok with that?

darunrs: Yep! No issues with it. I'm happy to let it sit.

PR #981, opened 2024-08-05
fix: display issue in explorer indexers

### LINKED ISSUE #980: display issue in explorer indexers

Kevin101Zhang:
explore seems to be missing a few indexers and the reason was due to pagination starting from 0. so on load it would grab the intial 50 and then onclick for loadmore it would still grab the 0-50 already on display.

fix: changes page to intialize to 1. 

Initial Load (0-49) Indexers Loaded
First Click (page = 1): Loads indexers from 50 to 99.
Second Click (page = 2): Loads indexers from 100 to 149.
```
const handleLoadMore = () => {
  const start = page * PAGE_SIZE; // FIRST: 1*50 = 50 -> SECOND: 2*50 = 100
  const end = start + PAGE_SIZE; //  FIRST: 50+50 = 100 -> SECOND: 100+50 = 150
  const newIndexers = indexers.slice(start, end); // FIRST CLICK: 50-99 -> SECOND CLICK: 100-149
  setCurrentPageIndexer([...currentPageIndexer, ...newIndexers]); 
  // FIRST: 0-49 + 50-99, SECOND: 0-99 + 100 - 149
  setPage(page + 1); 
};
```

### DIFF
--- a/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
+++ b/frontend/widgets/src/QueryApi.IndexerExplorer.jsx
@@ -40,7 +40,7 @@ const [hasMetadataRendered, setHasMetadataRendered] = useState(false);
 const [indexers, setIndexers] = useState([]);
 const [total, setTotal] = useState(0);
 const [currentPageIndexer, setCurrentPageIndexer] = useState([]);
-const [page, setPage] = useState(1);
+const [page, setPage] = useState(0);
 
 const [myIndexers, setMyIndexers] = useState([]);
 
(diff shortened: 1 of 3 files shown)


frontend/widgets/src/QueryApi.IndexerExplorer.jsx
```diff
@@ -40,7 +40,7 @@ const [hasMetadataRendered, setHasMetadataRendered] = useState(false);
 const [indexers, setIndexers] = useState([]);
 const [total, setTotal] = useState(0);
 const [currentPageIndexer, setCurrentPageIndexer] = useState([]);
-const [page, setPage] = useState(0);
+const [page, setPage] = useState(1);
```
darunrs, line 43: I think setPage being initialized to 1 is confusing. Instead, could you modify the handleLoadMore to do something like:
```
const newPage = page + 1;
// Do load more logic using newPage
setPage(newPage);
```

I feel the above reads a little nicer? Ideally, you would do setPage first, but I wonder if there could be lag in setting the value, so I am fine making that the last item.

darunrs (approved):

frontend/widgets/src/QueryApi.IndexerExplorer.jsx
Kevin101Zhang: Hmmm, I was the one who did it in a more unconventional setting before.

It's more common to start the initial page state at 1. This is because pagination systems often use 1-based indexing, where the first page is labeled as page 1. Some systems and APIs use 0-based indexing, where the first page is labeled as page 0 but for consistency and user-friendliness, starting at 1 is generally more intuitive, as it aligns with typical expectations

PR #987, opened 2024-08-06
fix: Receiver blocks stream would not query today

darunrs:
The current date used to start the receiver blocks backfill would have hour, minute, second, and nanosecond data. This meant when it reached today's date, it may potentially be ahead. This would lead to today not being queried. This PR fixes this by ensuring the current date is at the very start of the day, which guarantees it passes the <= check.

### DIFF
--- a/block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
+++ b/block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
@@ -370,7 +370,7 @@ mod tests {
             .expect_get_text_file()
             .returning(move |_, _| {
                 Ok(crate::test_utils::generate_block_with_timestamp(
-                    &(Utc::now() - Duration::days(2) + Duration::hours(1))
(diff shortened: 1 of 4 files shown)


darunrs:
Would like to get this one out to prod, since it can help speed up a lot of indexers. Will merge now and verify it works like expected in dev. Please review after the fact if needed!

block-streamer/src/receiver_blocks/receiver_blocks_processor.rs
```diff
@@ -165,7 +165,15 @@ impl ReceiverBlocksProcessor {
         try_stream! {
             let start_date = self.get_nearest_block_date(start_block_height).await?;
             let contract_pattern_type = ContractPatternType::from(contract_pattern.as_str());
-            let mut current_date = start_date;
+            let mut current_date = start_date
+                .with_hour(0)
+                .unwrap()
```
morgsmccauley, line 170: Is it safe to do this? Can we handle `None` explicitly?

PR #988, opened 2024-08-06
fix: Decreasing Processed Block Error

darunrs:
Block streams are failing due to their start block being lower than the current. This PR fixes this issue by seeding the health task with the input start block. This should fix the issue in most cases. I've also ensured that the handoff between receiver_blocks and lake also do not repeat a block.

### DIFF
--- a/block-streamer/src/block_stream.rs
+++ b/block-streamer/src/block_stream.rs
@@ -187,7 +187,7 @@ impl BlockStream {
                             health_lock.processing_state = ProcessingState::Waiting;
                         }
                         Ordering::Equal => {
-                            tracing::error!(
+                            tracing::warn!(
                                 account_id = config.account_id.as_str(),
                                 function_name = config.function_name,
                                 "No block has been processed for {stalled_timeout_seconds} seconds"
@@ -352,7 +352,7 @@ pub(crate) async fn start_block_stream(
     .context("Failed while fetching and streaming bitmap indexer blocks")?;
 
     let last_indexed_near_lake_block = process_near_lake_blocks(
-        last_bitmap_indexer_block + 1,
+        last_bitmap_indexer_block,
         lake_s3_client,
         lake_prefetch_size,
         redis,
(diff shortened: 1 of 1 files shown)


block-streamer/src/block_stream.rs
```diff
@@ -342,7 +352,7 @@ pub(crate) async fn start_block_stream(
     .context("Failed while fetching and streaming bitmap indexer blocks")?;
 
     let last_indexed_near_lake_block = process_near_lake_blocks(
-        last_bitmap_indexer_block,
+        last_bitmap_indexer_block + 1,
```
darunrs, line 355: I did a +1 here too, as I suspect that the handoff results in a repeat processing of that particular block.

darunrs:
I couldn't figure out a way to unit test this due to the loop. We would need to add a breakpoint for testing. Not sure if that's the right move.

block-streamer/src/block_stream.rs
```diff
@@ -183,6 +187,11 @@ impl BlockStream {
                             health_lock.processing_state = ProcessingState::Waiting;
                         }
                         Ordering::Equal => {
+                            tracing::error!(
```
morgsmccauley, line 190: Let's use `warn`. `error` should be for unrecoverable situations.

block-streamer/src/block_stream.rs
morgsmccauley: Have you confirmed this? or just a suspicion?

morgsmccauley: I believe we use `start_after`, so the previous implementation was correct?

darunrs: Gotcha. I did see a repeat in queryapi_indexer, but I wasn't 100% sure. I'll keep it as is and separately investigate that.

block-streamer/src/block_stream.rs
darunrs: Gotcha

PR #989, opened 2024-08-07
chore: Add additional logs for state transitions

### LINKED ISSUE #942: Add logs for all state transitions

darunrs:
I added a little bit more logging to the lifecycle state changes, as well as renamed the `Stopping/Stopped` states to `Suspending/Suspended` states to more accurately capture the intention.

### DIFF
--- a/coordinator/src/indexer_state.rs
+++ b/coordinator/src/indexer_state.rs
@@ -80,17 +80,22 @@ impl IndexerStateManagerImpl {
             tracing::info!("Migrating {}", raw_state);
 
             let old_state: OldIndexerState = serde_json::from_str(&raw_state)?;
+            let migrated_lifecycle_state =
+                if old_state.lifecycle_state == OldLifecycleState::Stopping {
+                    LifecycleState::Suspending
+                } else if old_state.lifecycle_state == OldLifecycleState::Stopped {
+                    LifecycleState::Suspended
+                } else {
+                    tracing::warn!("Unknown lifecycle state: {:?}", old_state.lifecycle_state);
+                    continue;
+                };
 
             let state = IndexerState {
                 account_id: old_state.account_id,
                 function_name: old_state.function_name,
                 block_stream_synced_at: old_state.block_stream_synced_at,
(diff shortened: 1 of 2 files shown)


darunrs:
Thinking of adding unit tests to lifecycle, where I just test that the handle_[state] functions do what we expect. This requires importing a LOT of mocked things. Not totally sure what's the cleanest way of doing that.

coordinator/src/indexer_state.rs
```diff
@@ -81,9 +86,12 @@ impl IndexerStateManagerImpl {
                 function_name: old_state.function_name,
                 block_stream_synced_at: old_state.block_stream_synced_at,
                 enabled: old_state.enabled,
-                lifecycle_state: LifecycleState::Running,
+                lifecycle_state: if old_state.lifecycle_state == OldLifecycleState::Stopping {
+                    LifecycleState::Suspending
+                } else {
+                    LifecycleState::Suspended
```
morgsmccauley, line 92: This would set any state that isn't `Stopping` to `Suspended` right? We should only set this if it is `Stopped`. We probably need to map each state specifically since we have two types.

darunrs: So I tested this locally, and it actually doesn't reach this part of the function, unless the initial parse fails. So, all the other options ended up getting parsed correctly, and the only ones which went here, were the states which are no longer part of Lifecycle. 

I verified this in the below unit test. I have four states to parse: Initializing, Running, Stopping, and Stopped. The redis state set call was only called twice, and they were called with the correct Suspending and Suspended state respectively. 

I don't mind being explicit here though.

morgsmccauley: Ah you're right, we'd only get here if parsing fails, which would only be for `Stopping`/`Stopped`. I'm fine leaving it as is then :)

darunrs: Moved the if statement to a separate variable assignment. I'll be able to put some warn logs out if there's an unexpected input.

PR #999, opened 2024-08-08
feat: editor DiffView now recognizes LaunchPad generated code and other cases

### LINKED ISSUE #998: editor DiffView should recognize LaunchPad generated code

Kevin101Zhang:


darunrs (commented):
This is a problem for ALL indexers. Not just launchpad. Can you confirm if the fix applies in all cases?

Kevin101Zhang:
There are 3 cases Im aware of:

Lets define the diff view as originalcode and modified code.
**originalCode** is the comparison the modified code should be against.
**modifiedCode** is the active editor that the user can actively edit on

1. If there user comes from launchpad the originalCode is the generateCode. Any modified work will be any additional code the user will write.
2. If the user comes from 'create new indexer', the originalCode will be the defaultCode/defaultSchema, The modified will be any additional code the user will write.
3. If the user comes from a previously published indexer. (queryAPI_indexer, devhub indexer etc). The originalCode will be the last state of the published code. The modified will be any additional code the user will write.

This pr only handles the case of launchpad code I believe, Ill iterate on it to confirm the behavior a previously published Indexer. As long as we confirm the use cases above is what we want.

Kevin101Zhang:
Can confirm this works for all cases.

There is a subtle difference unfortunately when there is previously published code that was unformatted at publish 
(old indexers specifically) is navigated to on the editor and formatting takes place in the editor on load.

The original will be the unformatted code and the modified will automatically be formatted.

darunrs (approved):
There's actually a fourth case. If a user forks an indexer, the originalCode should be carried over from the original indexer. The page looks like a create new indexer page, but with prefilled information from the forked indexer. If there's some state variable called originalCode or something which is the left side, let's update that in this case too. Feel free to merge once that's done. 

Nice job covering the other cases!

frontend/src/components/Editor/EditorComponents/ResizableLayoutEditor.jsx
```diff
@@ -157,23 +161,24 @@ export default function ResizableLayoutEditor({
     sizeThresholdFirst: 60,
     sizeThresholdSecond: 20,
   });
-
+  const defaultCode = launchPadDefaultCode ? launchPadDefaultCode : contextCode ? contextCode : originalIndexingCode;
```
darunrs, line 164: Just curious. Would it be easier instead, to have a state variable for original code? And we just set it whenever we transition to the Editor page? I feel that would make this workflow friendly to ANY scenario where we change to the Editor component. Or making it a prop that must be passed in.

darunrs: So, we don't default to anything or have a chain of things to try. We instead expect the application to correctly populate it according to its situation before going into Editor.

Kevin101Zhang: This component needs some refactoring for sure when I get the chance. It hasnt been touched at all. As you can see its still a jsx page. It currently houses 2 actual components that make sense together but id prefer we separate them into different files. **ResizableEditor** and 
**ResizableLayoutEditor** are the two components it currently houses. Ideally after I refactor them in a separate PR, there would be a transition to TS, and we can figure out a component that follows a view/viewmodal controllor type design pattern. Frankly the simplest way to change it would be to have the current **ResizableLayoutEditor** be a view that just focuses on render and the other component be a Container that houses the logic.

But in reality there should be a large refactor if I wasnt heavily focused on features. 
The Editor page is currently handling too many cases. My last effort was focused on transitioning the entirety of the file to TS and abtracting smaller components such as publish modals, local storage class, the developer menu, and etc.

But currently some useEffects only apply to a case where they are on a published indexer and others revolve around a state when they are unpublished. A lot of logic such as storage and etc are only applicable with the published indexer state. With the introduction of the Launchpad this further complicated the Editor page as now we have more than just defaultCode/default Schema so some functionality was added. There are plenty of oddities such as local storage has no reason to even run on a forked/launchpad/newIndexer and adding functionality like that not would only complicate the page even further. In essence the page itself needs more refactoring.

Ideally, My intial throught was to breakdown the Editor into 2 specific states therefore simplifying the use/unuse of logic we have in the Editor page current.

Editor Page when forked/launchpad/createnewIndexer (Trial state). 
Editor Page when on published indexer (Published State). 

This way we are on the Editor page with what we need and we dont actually need to fetch for anything until the user makes an edit. When they Publish they will only have their an actual indexerName therefore a route is created. so we can redirect them. 

Its something I can focus on during/after our transition post BOS.

PR #1001, opened 2024-08-08
type generation for launchpad code

### LINKED ISSUE #1000: type generation for launchpad Code

Kevin101Zhang:
types are now generated for launchpad code


frontend/src/components/Editor/EditorComponents/Editor.tsx
```diff
@@ -45,6 +53,30 @@ const fetchWizardData = (req: string): Promise<WizardResponse> => {
   return request<WizardResponse>('launchpad-create-indexer', req);
 };
 
+const generateCode = async (contractFilter: string, selectedMethods: Method[], selectedEvents?: Event[]) => {
```
darunrs, line 56: The function name "generateCode" has been reused multiple times and each time it does something different. Can you rename all of these instances to be more specific? No two functions should ever be the same unless they're in entirely different projects or attached to some class.

frontend/src/components/Editor/EditorComponents/Editor.tsx
```diff
@@ -194,15 +204,14 @@ const Editor: React.FC = (): ReactElement => {
...
-
     if (schemaError || codeError) {
       if (schemaError) schemaErrorHandler(schemaError);
       if (codeError) indexerErrorHandler(codeError);
       return;
     }
 
     handleCodeGen();
-  }, [fileName]);
+  }, [fileName, launchPadDefaultCode, launchPadDefaultSchema, schema, indexingCode]);
```
darunrs, line 214: I checked the useEffects which involve calling handleCodeGen and I'm a bit confused. We had two. One which triggers on filename change, and another which triggers either on schemaType change or monaco mount. 

I don't think type gen executing on schemaType change makes sense as typegen is what makes the types chance anyway. 

The other thing is that I would imagine type gen would execute now every single time there is any change at all to either the schema or code. Typegen is not a quick and easy thing to run. This will slow down the browser and also overpopulate logs. 

Firstly, type gen on code changes is unnecessary. Only changes on schema would change the types generated. Secondly, if we want to do type gen on schema changes, I would say we should debounce it and run it if we ever switch off the schema tab to something else.

frontend/src/components/Editor/EditorComponents/Editor.tsx
Kevin101Zhang: They are all one function and reference. we have a api endpoint that includes generate code and a function that is called generate code that consumes the endpoint.

Screenshot 2024-08-09 at 4 24 43 PM

darunrs: There's two definitions isn't there? One [here](https://github.com/near/queryapi/blob/main/frontend/src/pages/api/WizardCodeGenerator.ts#L156) and another [here](https://github.com/near/queryapi/blob/main/frontend/src/components/Editor/EditorComponents/Editor.tsx#L126)? The former being a class function and another being an inlined arrow function. I would rather you name the arrow function something unique.

darunrs (approved):

PR #1002, opened 2024-08-08
feat: Surface provisioning failures to user

### LINKED ISSUE #901: Surface provisioning failures to user
With the move to user owned logs tables, we removed the provisioning failure logs - if we fail to provisioning, it may not be possible to log the failure. We need to surface this information somehow, otherwise developers won't know what's gone wrong.

We have two options:
1. Re-introduce a shared/permanent logs table, specifically for provisioning failures
2. Log failures to the user owned table

I'm leaning towards 2, but it is tricky given that provisioning of the logs table itself may fail. _But_, most failures happen within provisioning of user related resources, so we have some confidence that the logs table should always be provisioned unless something is seriously wrong. With the correct alerts and testing (which we have) in place, we should be able to catch _real_ provisioning failures through our own channels, and surface user related failures via there logs table.

morgsmccauley:
This PR updates provisioning such that user-related errors are written to the user-owned logs table, allowing them to debug issues with their schema.

Logging to the user table is tricky, since this table is created _during_ the provisioning step itself. Therefore, I have split provisioning in to two phases:
1. System Resources - Setups up system related entities: database, schema, logs table/jobs etc.
2. User Resources - Applies user schema, configures Hasura etc.

This separation allows us to isolate the tasks which are likely to fail due to user error, and therefore only surface errors which are relevant.

The creation of the logs table _should always succeed_, if it doesn't there is something wrong with the system, i.e. some form of bug has been introduced. Errors thrown during the System portion of provisioning will be error logged to the machine, and I will tune the existing alert so that we are notified of these errors.

Additionally, I have converted all non-critical error logs to warnings, so that we don't get alerted on non-issues.

closes: #901


runner/src/indexer-config/indexer-config.ts
```diff
@@ -17,7 +17,8 @@ export class ProvisioningConfig extends BaseConfig {
   constructor (
     public readonly accountId: string,
     public readonly functionName: string,
-    public readonly schema: string
+    public readonly schema: string,
+    public readonly logLevel: LogLevel = LogLevel.INFO
```
morgsmccauley, line 21: Needed to move this from `IndexerConfig` to `ProvisioningConfig` so that `IndexerMeta` could be constructed with it. We only have access to the latter in Provisioning.

runner/src/server/services/data-layer/data-layer-service.ts
```diff
@@ -111,16 +111,12 @@ export function createDataLayerService (
               .then(() => {
                 logger.info('Successfully provisioned Data Layer');
               })
-              .catch((err) => {
-                logger.error('Failed to provision Data Layer', err);
```
morgsmccauley, line 115: Now logged within `Provisioner`

runner/tests/integration.test.ts
```diff
@@ -291,7 +308,7 @@ async function prepareIndexer(indexerConfig: IndexerConfig, provisioner: Provisi
   );
 }
 
-async function indexerLogsQuery(indexerSchemaName: string, graphqlClient: GraphQLClient): Promise<any> {
+async function indexerLogsQuery (indexerSchemaName: string, graphqlClient: GraphQLClient): Promise<any> {
```
morgsmccauley, line 311: @darunrs I think our editors are fighting each other 😅

darunrs: LMAO. I actually like the space after the function name for TS. I need to check my linting config.

darunrs (approved):

PR #1004, opened 2024-08-10
feat: Idempotent Provisioning

### LINKED ISSUE #915: Refactor Provisioning into Multi-Step Process
Provisioning encapsulates multiple steps, each of which can potentially fail. However, provisioning is set up to run like a synchronous process, running one step after the next. This leads to problems as we've noticed that tracking foreign keys can fail repeatedly. Provisioning needs to be converted to a multi step process. We need to store the progress in Redis, and ensure Runner can start at a particular step, as well as retry X amount of times before accepting a failure.

darunrs:
Provisioning is occasionally finnicky, failing on different parts of the process. Having retryable provisioning allows us to heal indexer state through retries rather than manually actioning this. To this, the provisioning step needs to have access to accurate status of provisioning in order to make the correct decisions on which non-idempotent actions to take. 

This PR introduces a ProvisioningState class which represents the status of provisioning, and can be used to make decisions on what parts of provisioning need to be completed.


runner/src/provisioner/provisioning-state/provisioning-state.ts
```diff
@@ -0,0 +1,81 @@
...
+import type HasuraClient from '../hasura-client';
+
+export default class ProvisioningState {
+  constructor (
+    private readonly config: ProvisioningConfig,
+    private hasuraMetadata: HasuraMetadata,
+    private tablesInSource: string[],
+  ) {}
+
+  static async loadProvisioningState (hasuraClient: HasuraClient, provisioningConfig: ProvisioningConfig): Promise<ProvisioningState> {
```
darunrs, line 12: I decided not to have the hasuraClient as a direct dependency for this class as it usually does not need access to the client, once initialized.

runner/src/provisioner/provisioning-state/provisioning-state.ts
```diff
@@ -0,0 +1,81 @@
...
+      const error = err as Error;
+      if (error.message.includes('source with name') && error.message.includes('not-exists')) {
+        return [];
+      }
+      throw error;
+    });
+    return new ProvisioningState(provisioningConfig, hasuraMetadata, tablesInSource);
+  }
+
+  async reload (hasuraClient: HasuraClient): Promise<void> {
```
darunrs, line 24: Technically, you can just recreate the state class to reload it, but I thought adding a specific function felt more semantically correct.

darunrs:
Tested locally with the following cases:
1. Completely new source, schema, and such
2. Re-provision the above with no changes
3. Remove the provisioning user resources part, such that only system resources are provisioned. Then retry with all provisioning enabled.
4. Remove adding permissions to system tables. Then retry with it enabled again. 

In all above cases, provisioning properly handled or healed the state.

runner/src/provisioner/provisioning-state/provisioning-state.ts
```diff
@@ -0,0 +1,81 @@
+import { type ProvisioningConfig } from '../../indexer-config';
+import { type HasuraTableMetadata, type HasuraMetadata, type HasuraSource, HASURA_PERMISSION_TYPES } from '../hasura-client';
+import type HasuraClient from '../hasura-client';
+
+export default class ProvisioningState {
```
darunrs, line 5: The intention with this class is to be a lightweight interface which exposes the state of resources currently, and can be used to bring the current state closer to the intended one.

runner/src/provisioner/provisioner.test.ts
```diff
@@ -13,15 +14,24 @@ describe('Provisioner', () => {
   let indexerConfig: IndexerConfig;
 
   const tableNames = ['blocks'];
+  const systemTables = [METADATA_TABLE_NAME, LOGS_TABLE_NAME];
+  const tableNamesWithSystemTables = ['blocks', ...systemTables];
   const accountId = 'morgs.near';
   const functionName = 'test-function';
   const databaseSchema = 'CREATE TABLE blocks (height numeric)';
   indexerConfig = new IndexerConfig('', accountId, functionName, 0, '', databaseSchema, LogLevel.INFO);
+  const emptyHasuraMetadata = generateDefaultHasuraMetadata();
```
darunrs, line 23: I chose to make the state class a variable rather than a class variable because the provisioner class is reused for all provisioning tasks and I did not want to manage cleaning the state between calls. If I re-created the state, it would make DI much harder. So, I decided to keep things simple and have it used in line. This does make mocking a little bit tougher as I need to mock metadata instead, but it's done in multiple places, so I felt this was fine.

runner/src/provisioner/provisioner.ts
```diff
@@ -364,47 +353,90 @@ export default class Provisioner {
...
+    }
   }
 
-  async provisionUserResources (indexerConfig: ProvisioningConfig): Promise<void> {
+  async provisionUserResources (indexerConfig: ProvisioningConfig, provisioningState: ProvisioningState): Promise<void> {
     const databaseName = indexerConfig.databaseName();
     const schemaName = indexerConfig.schemaName();
 
-    await this.runIndexerSql(databaseName, schemaName, indexerConfig.schema);
+    const onlySystemTablesCreated = provisioningState.getCreatedTables().every((table) => this.SYSTEM_TABLES.includes(table));
```
darunrs, line 411: There's only two states in which we reach this point in the code. Either the system provisioning was successful in this current pass, or we are revisiting. If we are in the initial pass, the state will actually say we have created and tracked nothing since we have not reloaded. That's fine for this line through, as we are really just interested if any non system tables have already been created or not. It's not important if the state is up to date following system provisioning, as the outcome would be the same.

darunrs:
Pointed my local Provisioner to prod and ran provisioning over the broken strachu.near/easy-poll-v0.3_testing indexer. It completely fixed it and the indexer now runs.

PR #1005, opened 2024-08-12
refactor: Move synchronisation logic to `LifecycleManger`

### LINKED ISSUE #1006: Suspend stalled indexers after X restarts

morgsmccauley:
This PR moves the synchronisation logic within `ExecutorsHandler`/`BlockStreamsHandler` in to `LifecycleManager`. The motive behind this is to provide greater awareness/control over the individual sync tasks (managing unhealthy, reconfiguring, etc.) to `LifecycleManger`.

I plan to build on top of this in my next PR, and have `LifecycleManager` count how many times an unhealthy stream/executor was restarted, and move the Indexer to `Suspended` if it goes above a configured threshold. This refactor makes "counting" actually possible.

Wanted to land the refactor PR first to make review easier.

### DIFF
--- a/coordinator/src/handlers/block_streams.rs
+++ b/coordinator/src/handlers/block_streams.rs
@@ -18,8 +18,6 @@ use tonic::transport::channel::Channel;
 use crate::indexer_config::IndexerConfig;
 use crate::redis::{KeyProvider, RedisClient};
 
-const RESTART_TIMEOUT_SECONDS: u64 = 600;
-
 #[derive(Debug, PartialEq)]
 pub enum BlockStreamStatus {
     /// Block Stream is running as expected
--- a/coordinator/src/handlers/block_streams.rs
+++ b/coordinator/src/handlers/block_streams.rs
@@ -29,7 +29,7 @@ pub enum BlockStreamStatus {
     /// Existing Block Stream is not running
     Inactive,
     /// Block Stream is not synchronized with the latest config
-    Unsynced,
+    Outdated,
     /// Block Stream has not been encountered before
     NotStarted,
 }
@@ -324,7 +324,7 @@ impl BlockStreamsHandlerImpl {
             .await?
         {
             if block_stream.version != config.get_registry_version() {
-                return Ok(BlockStreamStatus::Unsynced);
+                return Ok(BlockStreamStatus::Outdated);
             }
 
             if !self.is_healthy(&block_stream) {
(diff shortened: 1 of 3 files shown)


coordinator/src/handlers/block_streams.rs
```diff
@@ -20,6 +20,20 @@ use crate::redis::{KeyProvider, RedisClient};
 
 const RESTART_TIMEOUT_SECONDS: u64 = 600;
```
darunrs, line 21: Do we still need this timeout anymore if we no longer use it here?

coordinator/src/handlers/executors.rs
```diff
@@ -16,6 +16,18 @@ use crate::redis::KeyProvider;
 
 const RESTART_TIMEOUT_SECONDS: u64 = 600;
```
darunrs, line 17: Same for this one. Do we still need it?

coordinator/src/handlers/executors.rs
```diff
@@ -16,6 +16,18 @@ use crate::redis::KeyProvider;
...
 
+#[derive(Debug, PartialEq)]
+pub enum ExecutorStatus {
+    /// Executor is running as expected
+    Active,
+    /// Executor is in an unhealthy state
+    Unhealthy,
+    /// Executor
```
darunrs, line 25: I don't think you finished this comment

darunrs (commented):
Can we retain the restart timeout until the retry configuration is in place? The timeout of retries originally helped in two cases: Block streams which would restart over incorrect stale checks, and executors which constantly crash. The former is partially remediated due to the stale check being longer. But the latter is still an issue. Basically, I believe these indexers make some async call which fails, and it escapes our try catches for some reason. As a result, it crashes. This dumps logs into Runner and is also unnecessarily taxing on the system as restarts involve block prefetches. It also messes with the metrics. I have not gotten around to fixing any of those issues yet. Keeping in the timeout on retries at least mitigates the issue until we either have retry policies in place.

morgsmccauley:
Ah sorry, I meant to add the timeout back in but forgot - will do that now!

coordinator/src/handlers/executors.rs
morgsmccauley: Move to lifecycle

coordinator/src/handlers/block_streams.rs
morgsmccauley: Moved to lifecycle

coordinator/src/handlers/executors.rs
morgsmccauley: Woops, finished!

PR #1018, opened 2024-08-22
feat: Prevent registration of new accounts/indexers

### LINKED ISSUE #1015: Prevent creation of new indexers

morgsmccauley:
This PR updates the registry contract to prevent registration of new Accounts and also new Indexer Functions under existing accounts. Most of the changes are making the tests work due to registration now being prevented.

### DIFF
--- a/registry/contract/src/lib.rs
+++ b/registry/contract/src/lib.rs
@@ -270,12 +270,12 @@ impl Contract {
             _ => {}
         }
 
-        let account_indexers =
-            self.registry
-                .entry(account_id.clone())
-                .or_insert(IndexerConfigByFunctionName::new(StorageKeys::AccountV5(
-                    env::sha256_array(account_id.as_bytes()),
-                )));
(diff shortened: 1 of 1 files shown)


darunrs (approved):
Since we're preventing registration of new accounts, should we also disable the add_user function too or just ignore it?

gabehamilton:
Maybe we should check in the previous version of this file as the contract template? In case anyone wants to spin QueryApi up.

morgsmccauley:
> Since we're preventing registration of new accounts, should we also disable the add_user function too or just ignore it?

@darunrs Users are kind of meaningless at this point - I'll ignore for now.

> Maybe we should check in the previous version of this file as the contract template? In case anyone wants to spin QueryApi up.

@gabehamilton good point - I'll deploy this change but not merge.

morgsmccauley:
Deployed to:
✅ `dev-queryapi.dataplatform.near`
✅ `queryapi.dataplatform.near`

